PR #9, opened 2015-05-19
Monkey-patch on('exit') so that we always get it

isaacs:
Make sure that outputCoverage is always run at the *end* of the exit
handlers, even if that means it gets run multiple times.

Otherwise, a user-added on('exit') handler won't get coverage.

Note: this brings test coverage of node-tap's `lib/root.js` up to 94.44 where it belongs ;)

### DIFF
--- a/index.js
+++ b/index.js
@@ -76,9 +76,23 @@ NYC.prototype._wrapExit = function () {
       )
     }
 
-  onExit(function () {
-    outputCoverage()
-  })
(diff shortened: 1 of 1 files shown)


isaacs:
This needs a test, clearly.

bcoe:
@isaacs sorry about that! addressed in https://github.com/bcoe/nyc/pull/10

PR #16, opened 2015-05-23
load in instrumenter configuration from .istanbul.yml

bcoe:
Thanks for the awesome work @shackpank. I added a few tests, but otherwise things seemed to work great. a couple notes:

* Since `nyc` relies on the global coverage variable being `__coverage__` I've made this value hardcoded (in subprocesses I've seen some issues surrounding accessing `process.globals`, so I think it's good to push people towards `__coverage__`).
* I use the `this.cwd` directory variable, so that we'll keep loading the appropriate `.istanbul.yml` as `cwd` is changed in tests.
* I've added you as an owner to the GitHub project, I'm excited for any future contributions you might be able to make.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,17 @@ adding the following configuration:
 }}
 ```
 
+## Configuring Istanbul
+
+Behind the scenes nyc uses [istanbul](https://www.npmjs.com/package/istanbul). You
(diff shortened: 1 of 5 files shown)


shackpank:
:+1:

Many thanks for following this up, and fixing the things I missed.

PR #23, opened 2015-05-26
Remove unused jsonstream dep

zertosh:
I was going to rename it to JSONStream but it turns out that it's not being used.

Because of https://github.com/dominictarr/JSONStream/issues/68, it's causing lots of `npm ERR! missing: JSONStream@^1.0.3` and `npm ERR! extraneous: jsonstream@1.0.3 ` when your dependencies depend on both in case-insensitive OSes (e.g. browserify and tap).

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
   "dependencies": {
     "foreground-child": "^1.2.0",
     "istanbul": "^0.3.14",
-    "jsonstream": "^1.0.3",
     "lodash": "^3.8.0",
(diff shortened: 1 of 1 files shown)


bcoe:
@zertosh thank you, I don't know how that snuck in there -- probably unrelated debugging I was doing with the jsonstream/JSONStream bug you're mentioning.

bcoe:
@zertosh fixed and published in nyc@2.2.1, thank you.

Issue #29, opened 2015-06-14
Allow specifying reporters with `nyc` directly


Raynos:
To get a report I currently run

```sh
nyc node test/index.js && nyc report --reporter text --reporter html
```

It would be great to just run

```sh
nyc node test/inedx.js --reporter text --reporter html
```

Also note I'm relying on the "--reporter" can appear multiple times feature which may be accidental :)

bcoe:
@Raynos I love this idea. I will start looking into adding the option this week; it's not trivial because I need to be able to handle commands like:

`nyc ./node_modules/.bin/mocha --check-globals`, so I'll need to add some logic to intelligently munge `process.argv`.

Raynos:
@bcoe 

The way istanbul does this is `nyc --reporter text --reporter html -- node test/index.js` 

It can also be handled like node where flags after `nyc` and before the binary belong to `nyc` and all other flags belong to the binary.

bcoe:
@Raynos, I've got a testing branch that adds this support. I use `yargs` to parse `process.argv` and then I use the information I extract to slice the original arguments:

```js
NYC.prototype.mungeArgs = function (yargv) {
  return process.argv.slice(
    process.argv.indexOf(yargv._[0])
  )
}
```

Works great, but I'm waiting to land this pull, or something similar on one of the dependent libraries:

https://github.com/isaacs/foreground-child/pull/3

CC: @isaacs this was the feature request I was talking about, I think it's pretty cool.

Raynos:
Nice work :) I'll try it out once it's ready. Thanks @bcoe appreciate it.

PR #30, opened 2015-06-20
add shortcut for istanbul's check-coverage command

bcoe:
Added shortcut for istanbul's check-coverage command, see #29

@Raynos, would you mind giving this branch a spin, and letting me know if this was what you're looking for?

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 var foreground = require('foreground-child'),
+  path = require('path'),
   sw = require('spawn-wrap')
 
 if (process.env.NYC_CWD) {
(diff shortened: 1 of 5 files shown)


Raynos:
I will do; Working on a node service scaffold this weekend; I'll add `nyc` to it.

bcoe:
:+1:

Raynos:
Verified locally; works for me :) thanks

bin/nyc.js
```diff
@@ -13,36 +14,70 @@ if (process.env.NYC_CWD) {
...
+  if (~argv._.indexOf('report')) {
     // run a report.
     process.env.NYC_CWD = process.cwd()
 
     ;(new NYC({
       reporter: argv.reporter
     })).report()
+  } else if (~argv._.indexOf('check-coverage')) {
+    foreground(
+      path.resolve(__dirname, '../node_modules/.bin/istanbul'),
```
rmg: This will break if the `istanbul` dependency was fulfilled by a version installed in a parent node_modules. Or if your dependency tree has been flattened by npm@3, if I'm not mistaken...

PR #33, opened 2015-07-10
spawn istanbul in such a way that works for npm@3.x

bcoe:
@rmg pointed out in this pull request #30 that istanbul was being invoked in a manner that would work poorly if istanbul got hoisted up into a higher folder.

I've switched to using an approach which @othiym23 and @iarna suggested today:

1. I aded istanbul to npm run scripts.
2. I muck with the `cwd`, so that I can invoke the npm run script from the context of nyc.
3. I invoke the npm run script.

@jasisk I also added the `statements` option in this pull we didn't actually collide on the `s` key, because `s` was used on the outer `yargs` chain, not the inner `yargs` chain.

Let me know what ya'll think.

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -81,24 +81,15 @@ if (process.env.NYC_CWD) {
 
     report(argv)
   } else if (~argv._.indexOf('check-coverage')) {
-    var nycOutputDir = path.resolve(process.cwd(), './.nyc_output/*.json')
-
-    // switch the working directory to nyc,
-    // so that we have the standard bin.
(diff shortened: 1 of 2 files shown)


rmg:
I was thinking more like just replacing the resolve with something like this:
```js
require.resolve('istanbul/lib/cli');
```

bcoe:
@rmg that's clever, I like that it wouldn't require calling `chdir`.

jasisk:
:+1:. Sorry for the oversight on my part.

PR #41, opened 2015-08-27
Break out of loop once condition has been found

ronkorving:
This should improve performance somewhat


bcoe:
@ronkorving great little fix :+1:

PR #45, opened 2015-09-09
pulled in patched versions of foreground-child and spawn-wrap which work on windows

bcoe:


novemberborn:
Am curious if these patches are available upstream now? I can see a commit from you in `foreground-child` which has since been released: <https://github.com/isaacs/foreground-child/pull/4>.

bcoe:
@novemberborn oh, awesome, it looks like we could probably pull in `foreground-child` with some testing, but we'll have to keep the forked version of spawn-wrap:

https://github.com/isaacs/spawn-wrap/pull/5

I'll see if I can prod @isaacs and get this fixed up at some point -- the Windows shims are important for `yargs`, and `node-redis`.

PR #59, opened 2015-11-28
Exclusion improvements

Lalem001:
An attempt at improving file exclusion.

Since there are breaking changes here I decided to work off of the babel branch which already has a major semver change associated with it. Documentation may need to be adjusted.

Changes:
- Always exclude `node_modules`
- Exclude using glob/minimatch patterns instead of RegExp (breaking change)
- addAllFiles now excludes during glob.sync, significantly improving performance

I also have ideas for an include key, backed by glob, based on this PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 var _ = require('lodash')
 var fs = require('fs')
 var glob = require('glob')
+var minimatch = require('minimatch')
 var mkdirp = require('mkdirp')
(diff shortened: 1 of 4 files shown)


Lalem001:
Will make new PR resolving conflicts, and targeting master.

bcoe:
@Lalem001 :+1:

Issue #56, opened 2015-11-24
Normalize paths to posix before matching with `exclude` option


sindresorhus:
It's pretty awkward having to write a regex that supports both forward and backwards slashes.

https://github.com/jamestalmage/ava/blob/172f6d8f298351dc78ca9b06ffce26ba6a80d468/package.json#L81-L84
https://github.com/bcoe/nyc/blob/4e9545a0c039032585149bf503c797f975e1e982/index.js#L28

-

Ideally, `exclude` would just support globbing.

bcoe:
:+1: I will try to get a release of nyc out within the next few days, I'm currently working on adding support for babel's require hook which has taken me down a bit of a refactoring rabbit hole.

bcoe:
@sindresorhus well that took more days than I thought it would, sorry about that. when you have the time, could you give `nyc@5.0.0` a shot and let me know what you think of the new globbing functionality?

sindresorhus:
Tried it out. Nice work!

In AVA, for our own tests, we use `nyc` indirectly through `tap`. Would it be possible to bump `nyc` in `tap` and do a new release?

bcoe:
@sindresorhus I'll chat with @isaacs and see if we can work on upgrading, I'd love to move off of the fork of `spawn-wrap` potentially too.

Issue #50, opened 2015-10-27
coverage using exclude option


davidchase:
I am trying to use `nyc` with `ava` and running into a small issue for good measure here's a link to the conversation at [ava repo](https://github.com/sindresorhus/ava/issues/94).

So essentially i have a few directories in `node_modules` directory that i would like to cover since i do unit tests for them.

Unfortunately a setup like so:

```json
"config": {
    "nyc": {
      "exclude": [
        "node_modules/"
      ]
    }
  }
```

results in no coverage at all.

setting up more specific like `node_modules/utils` which are some of the files i need to cover results in a coverage containing too many files/packages such as `babel`, etc which i am not testing.

any suggestions on how to allow certain files from within `node_modules` directory?

thanks in advanced

davidchase:
any interest in adding an `include` key

bcoe:
@davidchase sorry for the slow reply, was on vacation and am just getting back in the swing of things.

I like the idea of an include key! do you want to take a stab at this? I don't think it would be too hard, just toggle the logic slightly when instrumenting files depending on whether an `exclude` or `include` option has been set when creating the nyc instance.

davidchase:
@bcoe cool no worries added a PR #54

bcoe:
@davidchase could you try `nyc@5.0.0`, we now have an `include` and `exclude` key, both of which use globs.

Issue #37, opened 2015-07-21
I see coverage reports for my node_modules


sam-github:
The README suggests that this shouldn't happen.

I notice I see it for the only one of my deps that are npm linked into my project.

Its not a big deal, I can visually skip over it the parts I don't care of, but is this intended behaviour?

bcoe:
@sam-github I've noticed this bug if I symlink files as well, I haven't had a chance to dig into the cause yet -- but seems like a legitimate bug.

bcoe:
@sam-github could you try `nyc@5.0.0` and let me know if it solves this problem for you, we've rewritten the logic for excluding files -- it's now glob based.

PR #60, opened 2015-11-29
Exclusion improvements

Lalem001:
An attempt at improving file exclusion.

The biggest improvement appears when using the `--all` flag.

Changes:
- Always exclude `node_modules`
- Exclude using glob/minimatch patterns instead of RegExp (breaking change)
- addAllFiles now excludes during glob.sync, significantly improving performance

Resolves:
- #56: Exclude now uses glob patterns via minimatch

Complicates:
- #50: node_modules are now always excluded

Feel free to question my changes/decisions. I am open to compromise if necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,19 +104,31 @@ nyc report --reporter=lcov
 
 ## Excluding Files
 
-By default nyc does not instrument files in `node_modules`, or `test`
-for coverage. You can override this setting in your package.json, by
-adding the following configuration:
+You can tell nyc to ignore specific files and directories by adding
+an `config.nyc.exclude` array to your `package.json`. Each element of
+the array is a glob pattern indicating which paths should be omitted.
 
-```js
+Globs are matched using [minimatch](https://github.com/isaacs/minimatch)
+
+In addition to patterns specified in the package, nyc will always ignore
+files in `node_modules`.
+
+For example, the following config will ignore all of `node_modules`,
+any files with the extension `.spec.js`, and anything in the `build/`
+directory:
+
+```json
 {"config": {
   "nyc": {
     "exclude": [
-      "node_modules/"
+      "**/*.spec.js",
+      "build/**"
     ]
   }
 }}
 ```
+
+> Note: exclude defaults to [`test`, `test.js`].
 
 ## Include Reports For Files That Are Not Required
 
--- a/README.md
+++ b/README.md
@@ -104,17 +104,17 @@ nyc report --reporter=lcov
 
 ## Excluding Files
 
-You can tell nyc to ignore specific files and directories by adding
+You can tell nyc to exclude specific files and directories by adding
 an `config.nyc.exclude` array to your `package.json`. Each element of
 the array is a glob pattern indicating which paths should be omitted.
 
 Globs are matched using [minimatch](https://github.com/isaacs/minimatch)
 
-In addition to patterns specified in the package, nyc will always ignore
+In addition to patterns specified in the package, nyc will always exclude
 files in `node_modules`.
 
-For example, the following config will ignore all of `node_modules`,
-any files with the extension `.spec.js`, and anything in the `build/`
+For example, the following config will exclude all `node_modules`,
+any files with the extension `.spec.js`, and anything in the `build`
 directory:
 
 ```json
@@ -122,13 +122,14 @@ directory:
   "nyc": {
     "exclude": [
       "**/*.spec.js",
-      "build/**"
+      "build"
     ]
   }
 }}
 ```
 
-> Note: exclude defaults to [`test`, `test.js`].
+> Note: exclude defaults to `['test', 'test{,-*}.js']`, which would exclude
+the `test` directory as well as `test.js` and `test-*.js` files
 
 ## Include Reports For Files That Are Not Required
 
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ function NYC (opts) {
   config = config.nyc || {}
 
   // load exclude stanza from config.
-  this.exclude = ['node_modules/**'].concat(config.exclude || ['test/**', 'test.js'])
+  this.exclude = ['**/node_modules/**'].concat(config.exclude || ['test/**', 'test{,-*}.js'])
   if (!Array.isArray(this.exclude)) this.exclude = [this.exclude]
   this.exclude = this._prepExcludePatterns(this.exclude)
 
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 var _ = require('lodash')
 var fs = require('fs')
 var glob = require('glob')
+var minimatch = require('minimatch')
 var mkdirp = require('mkdirp')
 var path = require('path')
 var rimraf = require('rimraf')
@@ -30,11 +31,9 @@ function NYC (opts) {
   config = config.nyc || {}
 
   // load exclude stanza from config.
-  this.exclude = config.exclude || ['node_modules[\/\\\\]', 'test[\/\\\\]', 'test\\.js']
+  this.exclude = ['node_modules'].concat(config.exclude || ['test', 'test.js'])
   if (!Array.isArray(this.exclude)) this.exclude = [this.exclude]
-  this.exclude = _.map(this.exclude, function (p) {
-    return new RegExp(p)
-  })
+  this.exclude = this._excludeDirs(this.exclude)
 
   // require extensions can be provided as config in package.json.
   this.require = config.require ? config.require : this.require
@@ -63,6 +62,13 @@ NYC.prototype._createInstrumenter = function () {
     noCompact: !instrumenterConfig.compact,
     preserveComments: instrumenterConfig['preserve-comments']
   })
+}
+
+NYC.prototype._excludeDirs = function (excludes) {
+  var appendix = _.map(excludes, function (exclude) {
+    return exclude + '/**'
+  })
+  return _.union(excludes, appendix)
 }
 
 NYC.prototype.addFile = function (filename, returnImmediately) {
@@ -101,7 +107,7 @@ NYC.prototype.addContent = function (filename, content) {
 NYC.prototype.shouldInstrumentFile = function (relFile, returnImmediately) {
   // only instrument a file if it's not on the exclude list.
   for (var i = 0, exclude; (exclude = this.exclude[i]) !== undefined; i++) {
-    if (exclude.test(relFile)) {
+    if (minimatch(relFile, exclude)) {
       return false
     }
   }
@@ -113,7 +119,7 @@ NYC.prototype.addAllFiles = function () {
 
   this._createOutputDirectory()
 
-  glob.sync('**/*.js', {nodir: true}).forEach(function (filename) {
+  glob.sync('**/*.js', {nodir: true, ignore: this.exclude}).forEach(function (filename) {
     var obj = _this.addFile(filename, true)
     if (obj.instrument) {
       module._compile(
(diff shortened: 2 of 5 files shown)


bcoe:
@Lalem001 I dig these changes :+1: a couple asks:

* would it be possible to add one unit test that demonstrates the glob-based exclusions in action., perhaps mock out the compile step and ensure that the appropriate set of files gets flagged.
* could we update the README with an example of the glob based approach, perhaps a common config somewhat might use?

Once we get this merged, and add the includes functionality let's publish this as 5.x.

Lalem001:
You ok with more commits? Or prefer squashed single commit?

bcoe:
@Lalem001 extra commits are fine, I tend to only squash in extreme cases, e.g., I've made 99 silly mistakes before opening a pull, in the general case I prefer to have a record of what actually happened.

Lalem001:
Regarding the communications toward issue #50, I was thinking maybe we can implement something similar to ESLint with exclusion negations? Thoughts?

bcoe:
@sindresorhus does this branch address the issues you raised in:

https://github.com/bcoe/nyc/issues/56

and

https://github.com/bcoe/nyc/issues/57

Mind giving it a spin?

bcoe:
@Lalem001 hey, just remembered we already use `glob` in the codebase, I don't think we need `glob` and `minimatch` ... perhaps you could replace our usage of `glob` with `minimatch` instead?

Lalem001:
glob uses minimatch internally. I can look to see if I can access minimatch via glob in the morning.

bcoe:
@Lalem001 gotcha, sorry for being confusing, I was thinking that `minimatch` was a stripped down version of `glob` (like optimist vs. minimist); I'd rather pull in both dependencies than jump through hoops to make `glob` work as our string matcher.

bcoe:
@Lalem001 let's give @sindresorhus a chance to chime in on this, since it solves a couple issues he opened, and then we'll get this merged :+1:

README.md
```diff
@@ -104,20 +104,32 @@ nyc report --reporter=lcov
...
+any files with the extension `.spec.js`, and anything in the `build/`
+directory:
+
+```json
 {"config": {
   "nyc": {
     "exclude": [
-      "node_modules/"
+      "**/*.spec.js",
+      "build/**"
```
sindresorhus: According to line 75 the `/**` is not needed

index.js
```diff
@@ -65,6 +64,24 @@ NYC.prototype._createInstrumenter = function () {
...
+NYC.prototype._prepExcludePatterns = function (excludes) {
+  var directories = []
+  excludes = _.map(excludes, function (exclude) {
+    // Remove leading "current folder" prefix
+    if (_.startsWith(exclude, './')) {
+      exclude = exclude.slice(2)
+    }
+
+    // Allow gitignore style of directory exclusion
+    if (!_.endsWith(exclude, '/**')) {
```
sindresorhus: What if it ends with `/` like the `node_modules/` in the readme example? Then it will become `node_modules//**`.

index.js
```diff
@@ -30,11 +31,9 @@ function NYC (opts) {
   config = config.nyc || {}
 
   // load exclude stanza from config.
-  this.exclude = config.exclude || ['node_modules[\/\\\\]', 'test[\/\\\\]', 'test\\.js']
+  this.exclude = ['node_modules/**'].concat(config.exclude || ['test/**', 'test.js'])
```
sindresorhus: I would ignore `test-*.js` by default too. It's a very common pattern. https://github.com/sindresorhus/ava/blob/eaf3e0f8a66980e87496a30b9f4a8c45329f0ae9/api.js#L198

index.js
```diff
@@ -65,6 +64,24 @@ NYC.prototype._createInstrumenter = function () {
   })
 }
 
+NYC.prototype._prepExcludePatterns = function (excludes) {
+  var directories = []
+  excludes = _.map(excludes, function (exclude) {
+    // Remove leading "current folder" prefix
```
sindresorhus: Why is this needed? Pretty sure `glob` handles this fine.

index.js
sindresorhus: What if you also have `node_modules` directories in sub-directories? Shouldn't the pattern be `**/node_modules/**`?

https://github.com/sindresorhus/ava/blob/eaf3e0f8a66980e87496a30b9f4a8c45329f0ae9/api.js#L203

README.md
```diff
@@ -104,20 +104,32 @@ nyc report --reporter=lcov
...
+the array is a glob pattern indicating which paths should be omitted.
 
-```js
+Globs are matched using [minimatch](https://github.com/isaacs/minimatch)
+
+In addition to patterns specified in the package, nyc will always ignore
+files in `node_modules`.
+
+For example, the following config will ignore all of `node_modules`,
+any files with the extension `.spec.js`, and anything in the `build/`
```
sindresorhus: `build/` => `build`, for consistency

index.js
bcoe: Agreed, it seems reasonable to assume we'd never want to instrument in `node_modules` regardless of the nesting.

bcoe:
@Lalem001 probably worth adding a few tests for the scenarios @sindresorhus brought up:

`node_modules/`,  `/foo/node_modules/foo`, and `./foo`.

In the test for `./foo`, you can check whether or not removing the leading `./` is required :+1:

bcoe:
@Lalem001 I should add, sorry for all the back and forth! your work is looking great.

index.js
Lalem001: This was done to mimic behavior seen in ESLint's ignore methods. This may actually be an issue with ESLint as well.

Would you guys prefer I went with strict glob? No pattern adjustments at all?

index.js
Lalem001: Again, something I saw in ESLint's source.

es128:
Just to chime in with a quick suggestion, the breaking change with dropping support RegExp could be avoided by using an interface that supports both globs and RegExp such as [anymatch](https://github.com/es128/anymatch).

Using a [bound matcher function](https://github.com/es128/anymatch#anymatch-matchers) in place of `shouldInstrumentFile` will also improve performance, preventing the need for minimatch to re-parse each glob on each file.

Lalem001:
@es128 `anymatch` would work as a `minimatch` replacement, but `glob` still runs on `minimatch`.

The primary improvement coming from this PR comes from the way I pass the patterns to `glob` when it is getting all files in the directory. As far as I know glob will not accept REGEX patterns.

index.js
Lalem001: Just did some quick testing with minimatch itself

```javascript
// minimatch(path, pattern)

minimatch('./node_modules/foo.js', 'node_modules/**') // false
minimatch('node_modules/foo.js', './node_modules/**') // false
minimatch('node_modules/foo.js', 'node_modules/**')   // true
```

Perhaps I should do as ESLint and also remove `./` from the path that is passed in?

es128:
Good point. The maintainers here don't seem concerned about the breaking change, so it's probably best left with the approach you have at this point.

But if there is any motivation to preserve support for regex, one more suggestion would be to also use [readdirp](https://github.com/thlorenz/readdirp) instead of node-glob and leverage its ability to handle functions for filtering. There may be a bit of perf boost to that change as well.

2¢

bcoe:
@es128, @Lalem001, I feel that globs are probably the more expected behavior for supporting exclusions -- I used regexes as a naive first pass at the problem.

I think we should land this update on `nyc@5.x` and if allow future pull requests to ask for regex support :+1: -- as is, switching to glob-based matching solves a few open issues.

@Lalem001 if you don't mind addressing Sindre's comments (adding a couple extra tests, I'm in a agreement that we should also ignore test-*.js), I'll get this merged tonight.

index.js
sindresorhus: I didn't comment on the merit of it. Just that I would handle a trailing `/` better by first removing it.

Lalem001: Ok, I will remove any trailing `/` before appending `/**`

Lalem001:
I would like to point out a circular dependency between `nyc` and `tap`.

Tap depends on `nyc@3.1.0`. Thus the root package still should to specify `node_modules` until tap is updated.

Lalem001:
This is ready for another review

bcoe:
@Lalem001 :+1: I will make an effort to get things merged tonight.

bcoe:
@Lalem001 I haven't had too much of a chance to debug yet, but I'm noticing that running coverage takes  a very long time on some libraries now, I tested on the following:

https://github.com/sindresorhus/pageres
https://github.com/kriskowal/q
https://github.com/bcoe/yarsay

The problem seems to relate to the default exclusion rules not properly disregarding `node_modules`:

screen shot 2015-12-02 at 8 07 18 pm

Lalem001:
I am looking into it now. Thanks for the heads up.

Lalem001:
Did some preliminary debugging.

I can reproduce the issue you observed when I `npm link nyc` into `yarsay`. In this case, the `node_modules` getting coverage are the `node_modules` of nyc.

When I installed nyc via `npm install <githubname>/<githubrepo>[#<commit-ish>]`, everything appeared to work as expected.

Did you install via another method?

Should we attempt to handle this case?

Lalem001:
Perhaps this is #37 reoccurring?

Lalem001:
Ultimately the issue is the fact that the offending relative paths lead with `../` and that `**/node_modules/**` will not match such paths.

One solution would be to remove leading `../` from all relative paths.
`Path.normalize` might handle this for us.

bcoe:
@Lalem001 I think this would be worthwhile. Having the `npm link` flow work is great for testing, so I think we should make sure to support this.

Lalem001:
@bcoe I apologize in advance for the wall of text.

I have experimented with the proposed solutions. I have had some success with sending both the relative and full path to `shouldInstrumentFile`. If either path is excluded, then I exclude the file.

There is a caveat however. If, by chance a js file in the root of the symlinked module was required, then that file would not be excluded by the `**/node_modules/**` rule.

For example, if a project under test, symlinked glob, the paths could be something like:

- `../glob/glob.js` -> included
- `/Users/user/glob/glob.js` -> included

As you can see, `node_modules` does not appear in these paths.

In the mean time, glob's references to modules would be excluded.

- `../glob/node_modules/minimatch/minimatch.js` -> included
- `/Users/user/glob/node_modules/minimatch/minimatch.js` -> **excluded**

If you are ok with this caveat, I am ready to commit/push.

---

If not, I have an idea that would transform symlink target paths to paths that are more "logical", but it would likely require a small performance hit.

Basically I would glob the full project searching for all symlinks. Then compare each file's path to a list of known symlink targets. If there is a match, replace the matching portion of the file's path with the symlink's path.

example:
Given a symlink: `project/node_modules/nyc -> /Users/user/nyc`
```javascript
var path = require.resolve('nyc')
// > '/Users/user/nyc/index.js'
path = path.replace('/Users/user/nyc', 'project/node_modules/nyc')
// > 'project/node_modules/nyc/index.js'
```

The full project glob `**/*/`, can likely be reduced to `node_module/*/` glob, significantly reducing the performance hit, but may let edge cases through.

Lalem001:
A somewhat related, open, issue can be found in https://github.com/nodejs/node/issues/3402

bcoe:
@Lalem001 thanks for all the hard work. This caveat sounds reasonable, and should considerably increase the speed of running `nyc` when link has been used. If you push up code today, I'm hoping to do some OSS work today.

Lalem001:
I am not sure why these tests are failing. Even my local copy of latest master is failing. Ideas?

bcoe:
@Lalem001 merging two branches broke each other in conjunction, I'll have a patch in shortly.

Lalem001:
@bcoe Travis tests are passing now

bcoe:
Closing in favor of https://github.com/bcoe/nyc/pull/84, which has a few minor tweaks.

PR #65, opened 2015-11-30
Improve support for custom require hooks

novemberborn:
First of, thanks for adding source map support! I've been trying to use `nyc@4` with the [project](https://github.com/novemberborn/buoyant) that lead to my [code coverage article](https://medium.com/@novemberborn/code-coverage-with-babel-istanbul-nyc-83b8c2f1093#.29ern9dza) from last week. I got it working but had to make some changes. This is the first PR which covers the custom require hook implementation.

---

The custom hook should decide which modules it wants to ignore. For example Babel lets you customize this beyond just `node_modules` using the [`only` and `ignore` options](https://babeljs.io/docs/usage/options/).

Custom hooks often store the hook they're replacing, and may call this hook when loading a module (e.g. if they decide to ignore that particular module). Hooks may also be removed, restoring the original hook. Libraries like [proxyquire](https://github.com/thlorenz/proxyquire) always fall back to the original hook but modify the module object.

The implementation was changed to:

* always pass the original module object to the custom hook
* support restoring older hooks
* correctly compile the code by using `Module.prototype._compile`

It's getting late in the day in my timezone so I haven't had a chance to update the tests to cover this new behavior. The test suite in my own project passes though when run under `nyc` so that's encouraging.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ var _ = require('lodash')
 var fs = require('fs')
 var glob = require('glob')
 var mkdirp = require('mkdirp')
+var Module = require('module')
 var path = require('path')
 var rimraf = require('rimraf')
 var onExit = require('signal-exit')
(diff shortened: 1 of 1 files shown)


bcoe:
@novemberborn this is looking great, thanks for the contribution! I'll have some comments after the work day today.

index.js
```diff
@@ -129,40 +130,54 @@ NYC.prototype.addAllFiles = function () {
...
-    }
+    // always use node's original _compile method to compile the instrumented
+    // code. if a custom hook invoked the default hook the code should not be
+    // compiled using the custom hook.
+    Module.prototype._compile.call(module, obj.content, filename)
+  }
 
-    module._compile(obj.content, filename)
+  var wrapCustomHook = function (hook) {
+    return function (module, filename) {
```
bcoe: I love this approach, much cleaner than what I was doing before -- also I think that it fixes the issue I was originally having that required me to ignore the `node_modules` folder -- great job.

bcoe:
@novemberborn this is great work, and I think we're sufficiently covered by the existing test cases -- I also tested it on several libraries I've instrumented with nyc (yargs, async, pageres, yarsay).

I say :shipit:

PR #69, opened 2015-12-02
Update require hook documentation

novemberborn:
It struck me that the documentation is still heavily biased towards the initial approach in <https://github.com/bcoe/nyc/pull/58>. I've tried to rewrite it to talk about custom require hooks in general, illustrated by using Babel.

I'm not quite sure what the screen shot was meant to illustrate either so I figured I'd remove it for simplicity. Happy to put it back or make other changes though!


novemberborn:
@bcoe while we're on the subject I'm of a mind to change the `babel` tests to not use Babel either. This should let us test falling back to the default handler as well as multiple require hooks. Will need to deal with the test issues in #66 first though. Does that sound like a good idea?

bcoe:
@novemberborn this sounds smart. I've been concerned for a while with how complex the tests are, turns out I didn't have a clue how to exercise test coverage for a test coverage library -- any ideas you have that help simplify and generalize the tests are greatly appreciated.

novemberborn:
> turns out I didn't have a clue how to exercise test coverage for a test coverage library

Wrapping in-process and then requiring seems smart enough! The fixtures are confusing though :) I'll see if that can be simplified. I'd also prefer simple require hooks rather than pulling in Babel. Easier to see what's really going on that way.

---

Any objections to the documentation change in this PR?

README.md
```diff
@@ -36,28 +36,17 @@ If you're so inclined, you can simply add nyc to the test stanza in your package
...
-3. install nyc, and run it with the appropriate `--require` flags:
-
-```sh
-nyc --require babel-core/register mocha
-```
-
-nyc uses source-maps to map coverage information back to the appropriate lines of the pre-transpiled code:
-
-<img width="350" src="screen.png">
+Source maps are used to map coverage information back to the appropriate lines
```
bcoe: :+1:

bcoe:
:shipit:

PR #71, opened 2015-12-03
stop bundling foreground-child

bcoe:
@novemberborn pointed out that we no longer need to bundle foreground-child.

It sure would be great if we could land https://github.com/isaacs/spawn-wrap/pull/5 too, perhaps I'll take another look at this on the weekend.


novemberborn:
Nice!

PR #73, opened 2015-12-03
Rewrite custom require hook & source map tests

novemberborn:
This is a rewrite of the source map tests so I can add the test coverage required in #66. This necessitated replacing the Babel test with a better one for generic custom require hooks.

This uses a new npm package named [source-map-fixtures](https://github.com/novemberborn/source-map-fixtures). It provides a few source files with different source map variants, making it easier to generate fixtures. I'm planning on updating the source map tests in AVA as well.

I've solved the problem of generating coverage reports for testing by adding a `_generateCoverage` script in `test/fixtures`. It generates a report for the source map fixtures.

And finally I've excluded the two tests from the coverage report. Seemed silly to get missing test coverage for the tests…

PR #76, opened 2015-12-06
the combination of #73 and #66 broke tests on master

bcoe:
combining #73 and #66 broke tests on master:

* since #66 rewrites paths based on the source-map, the paths being used to lookup coverage info in tests were broken.
* removing babel from `nyc` in #73 identified an issue with `require.main.paths()` I've instead moved to an approach that: 1. attempts requiring from the current working directory 2. falls back to requiring the full path of the additional module.

CC: @Lalem001, @novemberborn

### DIFF
--- a/test/source-map-cache.js
+++ b/test/source-map-cache.js
@@ -7,6 +7,8 @@ var fixture = require('source-map-fixtures').inline('branching')
 // Coverage for the fixture is stored relative to the root directory. Here
 // compute the path to the fixture file relative to the root directory.
 var relpath = './' + path.relative(path.join(__dirname, '..'), fixture.file)
+// the sourcemap itself remaps the path.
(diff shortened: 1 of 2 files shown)


test/source-map-cache.js
```diff
@@ -7,6 +7,8 @@ var fixture = require('source-map-fixtures').inline('branching')
 // Coverage for the fixture is stored relative to the root directory. Here
 // compute the path to the fixture file relative to the root directory.
 var relpath = './' + path.relative(path.join(__dirname, '..'), fixture.file)
+// the sourcemap itself remaps the path.
+var mappedPath = './' + path.join(path.dirname(relpath), '../src/branching.js')
```
novemberborn: I'll see if I can add this to to the fixture object instead.

novemberborn:
@bcoe well I hadn't updated #66 yet to add tests ;-)

PR #77, opened 2015-12-07
use capture-require

jamestalmage:
Well, this doesn't work, but I am a bit stumped as to why.

It no longer distinguishes between `default` and `wrapped` (it just uses the wrapping function to wrap the current top of `require.extensions['.js']`). This also means it never calls `addFile`, but addFile isn't really necessary. The default `require.extensions['.js']` should be forwarding you the content anyway.

@novemberborn - can you take a look and give me a guess as to why those tests are failing (like what information `nyc` isn't collecting correctly). I've spent quite a bit of time groking the `wrapRequire` section of the `nyc` codebase, but haven't really dived deep anywhere else.

I will keep trying, just hoping for another set of eyeballs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@ var _ = require('lodash')
 var fs = require('fs')
 var glob = require('glob')
 var mkdirp = require('mkdirp')
-var Module = require('module')
+var captureRequire = require('capture-require')
 var path = require('path')
 var rimraf = require('rimraf')
 var onExit = require('signal-exit')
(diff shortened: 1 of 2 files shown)


bcoe:
@jamestalmage we'll get it to work, I'm really happy with how much logic this will trim out of the core library -- `nyc`'s strength is in it's subprocess handling, I love the idea of splitting more and more logic into other modules.

Complete aside _but_, I wonder if it's worth pulling the source-map-remapping logic into its own module.... it is pretty domain specific, since it assumes an Istanbul coverage output.

novemberborn:
@jamestalmage I'm not quite sure. It may be to do with <https://github.com/bcoe/nyc/issues/79>, or that the tests tend to reload `index.js` after clearing it from the require cache.

jamestalmage:
> Complete aside but, I wonder if it's worth pulling the source-map-remapping logic into its own module.... it is pretty domain specific, since it assumes an Istanbul coverage output.

Modularize all the things. :+1:

jamestalmage:
I think part of the problem might be that the old `.wrap()` just blew away whatever was existing on `require.extensions['.js']`. That meant repeated calls to `.wrap()` did not have side effects. In an attempt to be more generalized `capture-require` actually looks at `Object.getOwnPropertyDescriptor` and will forward through future added extensions to the previous `getter`/`setter` pair. Since `nyc` installs its hook multiple times in the test, this ends up meaning the coverage transform gets applied multiple times (once for the hook installed in this test, and once for every previous test).

All speculation on my part at this point. But if it turns out to be true, then we actually only have a problem in the tests, since `nyc` won't invoke `capture-require` multiple times IRL.

Not sure on the best strategy for solving that.

novemberborn:
> Not sure on the best strategy for solving that.

Let's see if we can tackle #80.

bcoe:
@jamestalmage any headway on this bug?

jamestalmage:
closed via #90

PR #78, opened 2015-12-07
Add source map tests

novemberborn:
Tests `_rewritePath()` as requested by @bcoe in #66.

I'd like to add a test for reading a source map file but I'm not quite sure where to put it. I mean I can test the `add()` method I suppose but there's no real test even verifying it's called. @bcoe any ideas? Happy with the coverage we have now?

---

This bumps `source-map-fixtures`. I've added a `sourceFile` path which can be used to create the `mappedPath` from #76.

I realized the generated coverage report already applied source maps. It now writes the raw coverage file instead.

I've added a few more fixtures (with coverage) to test `_rewritePath()`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
   "devDependencies": {
     "chai": "^3.0.0",
     "sinon": "^1.15.3",
-    "source-map-fixtures": "^0.1.0",
+    "source-map-fixtures": "^0.2.0",
     "standard": "^5.2.1",
     "tap": "^1.3.4"
   },
--- a/test/fixtures/_generateCoverage.js
+++ b/test/fixtures/_generateCoverage.js
@@ -14,14 +14,9 @@ var NYC = require('../../')
 // Load the 'branching' source map fixture.
 var fixture = require('source-map-fixtures').inline('branching')
 
-// Prevent pollution from earlier nyc runs.
-var tempDirectory = path.join(__dirname, '.nyc_output')
-rimraf.sync(tempDirectory)
-
 // Inject nyc into this process.
 var nyc = (new NYC({
-  cwd: path.join(__dirname, '..', '..'),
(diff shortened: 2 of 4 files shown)


bcoe:
@novemberborn one other annoying thing worth keeping in mind, I started doing a bunch of Windows testing on the weekend and, with the source-map fixtures, we run into issues with `path.sep` -- whether `/` or `\` is used is all over the place:

* istanbul seems to sometimes set `./`.
* we sometimes set `./`.

Now, our core functionality does almost work on Windows without `path.sep` problems -- I hunted down the problem with spawning -- but the test suite is a ways of working; it might be smart to abstract the fixture lookup to not care about `path.sep` ... or something like that.

novemberborn:
> it might be smart to abstract the fixture lookup to not care about path.sep ... or something like that.

`source-map-fixtures` uses `path.join()` so I think it's just the test itself which prefixes with `./`.

bcoe:
@novemberborn the generated `coverage` file varies in `path.sep` usage too I think.

bcoe:
If you ever get adventurous, this is what I use for Windows Node testing, it's free and works pretty well on OSX:

https://github.com/xdissent/ievms

novemberborn:
> @novemberborn the generated coverage file varies in path.sep usage too I think.

Ah yes it would. Maybe you could track progress in an issue and we can hash stuff out there? Can we use Travis to test Node on Windows?

novemberborn:
This is passing CI now.

bcoe:
@novemberborn AppVeyor seems to be the best available:

https://ci.appveyor.com/project/bcoe/yargs

Let's move Windows chatting to a separate ticket, there's enough problems with Windows right now that it shouldn't block this branch :+1:

bcoe:
:shipit:

Issue #80, opened 2015-12-07
nyc-test wraps itself again and again


novemberborn:
Multiple tests in `test/nyc-test` instantiated `NYC` and call `wrap()`, wrapping the current process. This leads to poor test isolation. Each test should probably be run in its own process.

jamestalmage:
I think this is (at least partly) the source of my trouble with `capture-require`

PR #82, opened 2015-12-08
proposal: split into one test per file

jamestalmage:
This solves #80, and fixes the problems I was having integrating `capture-require` (at least for the one test I've actually separated out so far).

I really do not like this, but I am not sure of an easier way to pull it off.


jamestalmage:
closing in favor of #89

Issue #79, opened 2015-12-07
nyc-test needs tap --coverage


novemberborn:
`$(npm bin)/tap test/nyc-test.js` fails, where `$(npm bin)/tap --coverage test/nyc-test.js` succeeds. This seems wrong, `tap` comes bundled with `nyc` for its own coverage reports but that shouldn't impact the tests.

jamestalmage:
closed via #89

PR #83, opened 2015-12-08
fork for every test in `nyc-test`

jamestalmage:
Another solution for #80. Basically an automated version of #82.

This is sort of a crazy idea. I hacked it together quick, so forgive me for the fugly parts.

The meat of what happens is in `_build-tests.js`, it implements it's own `mocha` interface, and instead of running tests it just builds a bunch of individual test files using `function.toString()`.

This method makes helper methods a bit weird. I copied them all to `_header.js`, and prepended that to each generated file. We could maybe do something slicker, but it would require doing AST transforms. We probably could write a Babel plugin, or use `recast` to do that, but I think it would be a lot of work.

One tests in particular did not work for me. I did not understand the errors I was getting.

One thing I was surprised to find out is that your tests are pretty brittle when executed out of order. I needed to prefix the file names with an index, that fixed like 6 or 7 tests. Ideally, you would stop testing using `index.js` as your fixture entirely, and blow away `test/fixtures/.nyc_ouput` between each test. I actually think that would be pretty easy to do.

Also, rather than the hacky way I have moved helper functions into `_header.js`, it might make more sense to just create `test-utils.js` and require it at the top of each generated file. That would stop the linter from complaining (you will notice I disabled it). It would be some pretty minor refactoring from there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 coverage
 node_modules
 !node_modules/spawn-wrap
+built-*
--- a/package.json
+++ b/package.json
@@ -4,8 +4,9 @@
   "description": "a code coverage tool that works well with subprocesses.",
   "main": "index.js",
   "scripts": {
-    "pretest": "standard",
-    "test": "tap --coverage ./test/*.js"
+    "test": "npm run clean && npm run build && tap --coverage ./test/built-* ./test/source-map-cache.js -b",
+    "clean": "rm -rf ./.nyc_output && rm -rf ./test/fixtures/.nyc_output",
+    "build": "rm -rf ./test/built-* && node ./test/_build-tests.js"
   },
   "bin": {
     "nyc": "./bin/nyc.js"
@@ -18,6 +19,9 @@
         "coverage",
         "test/nyc-test.js",
         "test/source-map-cache.js",
+        "test/_build-tests.js",
+        "test/_header.js",
+        "test/built-.*\\.js",
         "test/fixtures/_generateCoverage.js"
       ]
     }
@@ -61,6 +65,7 @@
   },
   "devDependencies": {
     "chai": "^3.0.0",
+    "parse-function": "^2.0.1",
     "sinon": "^1.15.3",
     "source-map-fixtures": "^0.2.0",
     "standard": "^5.2.1",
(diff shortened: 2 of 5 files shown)


novemberborn:
My even crazier idea was to change the runner a little bit. Still one test file, but it starts each test in a child process and communicates the results. That means any shared setup code is still executed for each test without having to be duplicated.

> One thing I was surprised to find out is that your tests are pretty brittle when executed out of order. I needed to prefix the file names with an index, that fixed like 6 or 7 tests. Ideally, you would stop testing using index.js as your fixture entirely, and blow away test/fixtures/.nyc_ouput between each test. I actually think that would be pretty easy to d

Agreed.

jamestalmage:
> Still one test file, but it starts each test in a child process and communicates the results

Yeah, it would be nice not to have to do that, it ends up so verbose. This saves you quite a bit of verbosity (though it is kinda "magic").

I am also wondering if it wouldn't just be easier to just use a task runner (`gulp` probably), and apply coverage manually. If the tests required the covered file directly, there would be no need to use `nyc` to provide coverage for `nyc`.

novemberborn:
> apply coverage manually

How do you mean?

> If the tests required the covered file directly, there would be no need to use nyc to provide coverage for nyc.

Except for #79.

jamestalmage:
>> apply coverage manually
> How do you mean?

`$ istanbul instrument index.js > index.covered.js`

In test files:

```js
require('../index.covered.js');
```

It would be a bit more involved than that (we would need to use [`transform-deps`](https://www.npmjs.com/package/transform-deps) to rewrite `require(./lib/...)`, but that would be pretty easy. I think gulp would be a good fit.

> Except for #79.

Yeah - but we should fix that.

bcoe:
@jamestalmage, @novemberborn, I'll give some feedback on these two test housekeeping pulls tonight -- several great ideas being thrown out there which I need to grok :+1:

novemberborn:
@bcoe 

> I'll give some feedback on these two test housekeeping pulls tonight -- several great ideas being thrown out there which I need to grok

Here's another: #85.

jamestalmage:
Closing in favor of #89

PR #84, opened 2015-12-08
glob-based include/exclude functionality

bcoe:
A few slight tweaks to Lalem001's awesome work on glob-based excludes:

* I've switched to (at this time an older) version of `micromatch`. This works with leading `./` so that using `npm link` does not break exclusions.
* I've added support for `include` along with `exclude`.

Reviewers: @Lalem001, @davidchase, @novemberborn

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ You can tell nyc to exclude specific files and directories by adding
 an `config.nyc.exclude` array to your `package.json`. Each element of
 the array is a glob pattern indicating which paths should be omitted.
 
-Globs are matched using [minimatch](https://github.com/isaacs/minimatch)
+Globs are matched using [micromatch](https://www.npmjs.com/package/micromatch)
 
 In addition to patterns specified in the package, nyc will always exclude
 files in `node_modules`.
@@ -119,6 +119,19 @@ directory:
 
 > Note: exclude defaults to `['test', 'test{,-*}.js']`, which would exclude
 the `test` directory as well as `test.js` and `test-*.js` files
+
+## Including Files
+
+As an alternative to providing a list of files to `exclude`, you can provide
+an `include` key to specify specific files that should be covered:
+
+```json
+{"config": {
+  "nyc": {
+    "include": ["**/build/umd/moment.js"]
+  }
+}}
+```
 
 ## Include Reports For Files That Are Not Required
 
--- a/README.md
+++ b/README.md
@@ -93,17 +93,17 @@ nyc report --reporter=lcov
 
 ## Excluding Files
 
-You can tell nyc to ignore specific files and directories by adding
+You can tell nyc to exclude specific files and directories by adding
 an `config.nyc.exclude` array to your `package.json`. Each element of
 the array is a glob pattern indicating which paths should be omitted.
 
 Globs are matched using [minimatch](https://github.com/isaacs/minimatch)
 
-In addition to patterns specified in the package, nyc will always ignore
+In addition to patterns specified in the package, nyc will always exclude
 files in `node_modules`.
 
-For example, the following config will ignore all of `node_modules`,
-any files with the extension `.spec.js`, and anything in the `build/`
+For example, the following config will exclude all `node_modules`,
+any files with the extension `.spec.js`, and anything in the `build`
 directory:
 
 ```json
@@ -111,13 +111,14 @@ directory:
   "nyc": {
     "exclude": [
       "**/*.spec.js",
-      "build/**"
+      "build"
     ]
   }
 }}
 ```
 
-> Note: exclude defaults to [`test`, `test.js`].
+> Note: exclude defaults to `['test', 'test{,-*}.js']`, which would exclude
+the `test` directory as well as `test.js` and `test-*.js` files
 
 ## Include Reports For Files That Are Not Required
 
--- a/README.md
+++ b/README.md
@@ -93,19 +93,31 @@ nyc report --reporter=lcov
 
 ## Excluding Files
 
-By default nyc does not instrument files in `node_modules`, or `test`
-for coverage. You can override this setting in your package.json, by
-adding the following configuration:
+You can tell nyc to ignore specific files and directories by adding
+an `config.nyc.exclude` array to your `package.json`. Each element of
+the array is a glob pattern indicating which paths should be omitted.
 
-```js
+Globs are matched using [minimatch](https://github.com/isaacs/minimatch)
+
+In addition to patterns specified in the package, nyc will always ignore
+files in `node_modules`.
+
+For example, the following config will ignore all of `node_modules`,
+any files with the extension `.spec.js`, and anything in the `build/`
+directory:
+
+```json
 {"config": {
   "nyc": {
     "exclude": [
-      "node_modules/"
+      "**/*.spec.js",
+      "build/**"
     ]
   }
 }}
 ```
+
+> Note: exclude defaults to [`test`, `test.js`].
 
 ## Include Reports For Files That Are Not Required
 
--- a/README.md
+++ b/README.md
@@ -133,6 +133,8 @@ an `include` key to specify specific files that should be covered:
 }}
 ```
 
+> Note: include defaults to `['**']`
+
 ## Include Reports For Files That Are Not Required
 
 By default nyc does not collect coverage for files that have not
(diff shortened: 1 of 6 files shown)


Lalem001:
Looks good to me. :thumbsup:

Lalem001:
I see a potential problem with the include implementation.

If both `include` and `exclude` are defined, `shouldInstrumentFile` will ignore the `this.exclude` completely.

Also, if someone wished to include the entire contents of their source directory without using `--all`, this implementation of include will not do that. It will only specifically include whatever was `required`.

jamestalmage:
Personally, I have come to prefer [`multimatch`](https://www.npmjs.com/package/multimatch), you write specific includes, and can exclude with negated (`!`) patterns at any time. 

If you are going to provide both `include` and `exclude`, my opinion is that exclude should take priority (i.e. if it matches exclude then it is excluded no matter what include says) [`include-exclude`](https://www.npmjs.com/package/include-exclude) works that way.

README.md
```diff
@@ -93,20 +93,46 @@ nyc report --reporter=lcov
...
     "exclude": [
-      "node_modules/"
+      "**/*.spec.js",
+      "build"
     ]
   }
 }}
 ```
 
+> Note: exclude defaults to `['test', 'test{,-*}.js']`, which would exclude
```
novemberborn: There's a note paragraph at the end of the readme which is in italics, yet this is a quote. Would be nice to be consistent.

index.js
```diff
@@ -107,14 +127,14 @@ NYC.prototype.addContent = function (filename, content) {
...
-  // only instrument a file if it's not on the exclude list.
-  for (var i = 0, exclude; (exclude = this.exclude[i]) !== undefined; i++) {
-    if (exclude.test(relFile)) {
-      return false
-    }
+NYC.prototype.shouldInstrumentFile = function (filename, relFile) {
+  relFile = relFile.replace(/^\.\//, '') // remove leading './'.
+
+  if (this.include) {
+    return micromatch.any(filename, this.include) || micromatch.any(relFile, this.include)
```
novemberborn: Why check both `filename` (absolute?) and `relFile`?

index.js
```diff
@@ -31,11 +32,10 @@ function NYC (opts) {
   config = config.nyc || {}
 
   // load exclude stanza from config.
-  this.exclude = config.exclude || ['node_modules[\/\\\\]', 'test[\/\\\\]', 'test\\.js']
+  this.include = this._prepGlobPatterns(config.include)
+  this.exclude = ['**/node_modules/**'].concat(config.exclude || ['test/**', 'test{,-*}.js'])
```
novemberborn: The next line is no longer necessary, right? Which means you could wrap it with `this._prepGlobPatterns()` and only assign `this.exclude` once.

index.js
Lalem001: Had to do with how `minimatch` would choke on leading dots in a path. Not sure if `micromatch` has the same issue.

novemberborn:
> Also, if someone wished to include the entire contents of their source directory without using --all, this implementation of include will not do that. It will only specifically include whatever was required.

I guess the README could make that more clear (though the section after `config.include` covers it).

novemberborn:
> If you are going to provide both include and exclude, my opinion is that exclude should take priority (i.e. if it matches exclude then it is excluded no matter what include says)

There are default excludes though. Would be good to allow a particular module to be covered, even if generally all modules are excluded.

Lalem001:
> There are default excludes though. Would be good to allow a particular module to be covered, even if generally all modules are excluded.

@jamestalmage hit on this a bit. It would be possible to have exclusion negations with a leading`!`. ESLint does this. For example, since `node_modules` is always excluded, a negation to allow `glob` to be covered could be something like:

```javascript
{
  exclude: ['!node_modules/glob']
}
```

This is effectively the same as the current *idea* of include, where a file is included despite having been excluded in other patterns.

index.js
bcoe: Because `filename` is the complete path, we want to also check the relative-file or globs don't behave as one would expect. Take excluding the test folder:

I would expect to be able to write `test/**`, this however fails to exclude: 
`/Users/benjamincoe/bcoe/yargs/test/validation.js`.

testing against both the absolute and rebased path seems like an elegant solution.

jamestalmage:
Yep, I think that's why it may make sense to have separate include/exclude globs, with the following defaults:

```json
{
  "include": "**/*",
  "exclude": "!**/node_modules/"
}
```

Lalem001:
@jamestalmage That example is effectively what `--all` will accomplish with this PR.

What would be nice, and I was hoping to accomplish in a future PR, is to have something like:

```javascript
{
  include: ['src/**/*.js', 'bin/**/*.js']
}
```

This is not *precisely* possible with the current include implementation.

bcoe:
the two use-cases that I'd like to solve are:

1. a project has many files, but only cares about covering one -- for instance a codebase that transpiles quite a few files into a single asset (`moment` does this) -- I see `include` solving this problem.
2. a project that has many files, and only wants to exclude a few -- this is the more common case, and I think is solved by `exclude`.

I wonder what the most elegant approach to solve this is.

index.js
jamestalmage: Most API's require you to take absolute file names into account when specifying a glob pattern `**/test/**`, I am not sure if there is a good reason behind this or not. 

Might it be better to make people write their excludes with a leading `./`, and then replace that with the dirname of `package.json` when we load the config? So `./test/**` becomes `/Users/benjamincoe/bcoe/yargs/test/**`. It's certainly more explicit. 

I don't have a good reason not to do it this way, other than "it's not how I've seen it done before". Certainly not a good enough reason not to do it, but I am curious why others don't.

jamestalmage:
Right, I think what I suggested before:

```js
var defaults = {
  include: everything,
  exclude: node_modules_and_common_test_paths
};
```

Our algorithm for matching is:
```js
var fileMatches = includes.matches(file) && !excludes.matches(file)
```

1. If they provide their own includes we **replace** the default array, this makes a whitelist approach easy (but still provides automatic exclusion of `node_modules`, just in case they carelessly do something like `**/lib/*.js`).

2. If they provide their own excludes we append to the default array. This makes excluding one/two other files in addition to the sensible defaults easy.

3. If they really want to include something in `node_modules`, they have to provide a negation for it (since we aren't allowing them to replace the default excludes). It's a bit weird, but it is also a super rare need. Even then, it's probably the simplest (fewest characters) for the user once they understand. We obviously would need to document this with an example.

@bcoe, I think this method gives you the best answer to your request:

> 1. a project has many files, but only cares about covering one -- for instance a codebase that transpiles quite a few files into a single asset (moment does this) -- I see include solving this problem.

Just add a glob for that file to `include`.

> 2. a project that has many files, and only wants to exclude a few -- this is the more common case, and I think is solved by exclude.

Just add that file to `exclude`, since we are appending it to the default exclusion list (instead of replacing the default), you won't suddenly be covering `node_modules`, etc when you provide an exclusion pattern.They can use negation patterns for `exclude` in the corner cases where they actually want to exclude `node_modules`

bcoe:
@jamestalmage I am sold \o/

index.js
bcoe: @jamestalmage conceptually, I can see a developer assuming we match the path in the form:

`./foo.js` or `foo.js`, seems like a reasonable usability tweak to accept both -- let's try the approach and see if it bites us.

bcoe:
@jamestalmage I think this is ready to land, I've tested thoroughly at this point.

Lalem001:
Sorry for the last minute comment. Please bear with me, but I would like to air out my thoughts.

I see files being loaded in two distinct modes right now:

- Files required after `wrap` has been invoked (default)
- Files picked up by glob when the `--all` flag is present

In the default mode, include and exclude are only working against a very limited set of files.
Include is not actually adding any files that were not required.

In the second mode, all files that are not explicitly excluded are included.
Include, I think, would have little effect here.

I was thinking that include can instead replace the `**/*.js` that is currently in `addAllFiles`. Basically have a new function `addFiles` and `addAllFiles` simply calls the prior with `**/*.js`. Then include can be sent to `addFiles`.

This way include can specifically add files that are not required.

bcoe:
@Lalem001 let's open a discussion about `addAllFiles` in a separate thread, I would like to land a first pass of these glob changes, and get a release out of `nyc` out the door -- we have several important patches ready to go live.

Lalem001:
@bcoe understood. Go for it

jamestalmage:
:shipit:

PR #86, opened 2015-12-09
the JSON output dir test does not test what we think it does

jamestalmage:
Just documenting this. 

I suspect this is due to the fact we are running in a container instrumented by NYC

### DIFF
--- a/test/nyc-test.js
+++ b/test/nyc-test.js
@@ -138,37 +138,36 @@ describe('nyc', function () {
       var nyc = new NYC({
         cwd: process.cwd()
       })
-      var proc = spawn(process.execPath, ['./test/fixtures/sigint.js'], {
-        cwd: process.cwd(),
-        env: process.env,
-        stdio: 'inherit'
-      })
-      var start = fs.readdirSync(nyc.tmpDirectory()).length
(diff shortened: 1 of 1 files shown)


test/nyc-test.js
```diff
@@ -138,37 +138,36 @@ describe('nyc', function () {
...
-              // the subprocess we ran should output reports
-              // for files in the fixtures directory.
-              Object.keys(report).should.match(/.\/test\/fixtures\//)
-            }
+      //  var proc = spawn(process.execPath, ['./test/fixtures/sigint.js'], {
+      //    cwd: process.cwd(),
+      //    env: process.env,
+      //    stdio: 'inherit'
+      //  })
+      //  var start = fs.readdirSync(nyc.tmpDirectory()).length
```
jamestalmage: This test works even though we never spawn another process.

novemberborn:
#85 has a fix for this, spawning `bin/nyc.js` to run the `sigint.js` file.

jamestalmage:
> #85 has a fix for this, spawning bin/nyc.js to run the sigint.js file.

I took the same approach in #89 :smile:

novemberborn:
#104 fixes this.

PR #88, opened 2015-12-09
use a tilde range for micromatch

jamestalmage:
The intent was to use an older version of micromatch, but the current
semver specification allows up to latest.

The carrot range will allow any version up to, but not including 3.0.0.

Switched to a tilde range which allows up to, but not including 2.2.0.


novemberborn:
:+1:

bcoe:
I've talked with the micromatch folks, and they should have a release out this morning that lets us move to latest.

PR #90, opened 2015-12-11
Recast forking with capture require

jamestalmage:
Well this was easy!

This is #89, with with one commit from cherry picked from #77 (unmodified, not even a conflict!).

If nothing else, it's confirmation our testing strategy is moving in the right direction.

Now that I have `capture-require` actually working with something beyond it's own test suite, I guess it is time to publish it for real.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,7 @@ var fs = require('fs')
 var glob = require('glob')
 var micromatch = require('micromatch')
 var mkdirp = require('mkdirp')
-var Module = require('module')
+var captureRequire = require('capture-require')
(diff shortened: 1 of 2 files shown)


bcoe:
@jamestalmage this answers my question on #77, great work -- excited to get some of this work merged, and to then concentrate on better Windows support.

ariporad:
Thanks so much @jamestalmage!

@bcoe: Any idea when we might expect to see this merged?

jamestalmage:
#89 needs to get merged first.

PR #92, opened 2015-12-14
adapt remap-istanbul's mapping logic for nyc

### LINKED ISSUE #67: Ignoring code
Hi! Are there any alternatives to `/* istanbul ignore next */` and [similar](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md)? Doesn't seem like the Istanbul variant is taken into consideration at least.

novemberborn:
remap-istanbul's mapping function seems to cover more edge cases, as well as preserving "skip" annotations. This commit rewrites nyc's mapping logic to use a helper function adapted from remap-istanbul. See <https://github.com/SitePen/remap-istanbul/blob/30b67ad3f24ba7e0da8b8888d5a7c3c8480d48b2/lib/remap.js#L55:L108>
for the original.

Fixes #67.

@bcoe `remap-istanbul` is [BSD licensed](https://github.com/SitePen/remap-istanbul/blob/30b67ad3f24ba7e0da8b8888d5a7c3c8480d48b2/LICENSE). Not quite sure what attribution is required in the source code for copying their function.

It's possible that including the ["skip" annotation](https://github.com/SitePen/remap-istanbul/blob/30b67ad3f24ba7e0da8b8888d5a7c3c8480d48b2/lib/remap.js#L105) in nyc's existing logic is sufficient. Here's where I have to confess I have no clue how the source map format actually works :wink:

Regardless it seems like `remap-istanbul`'s logic is stronger (though in other places overly so). Unfortunately the package as a whole does too much to be hooked into nyc, and the remapping logic is tightly coupled with retrieving source maps so it can't be used directly.

### DIFF
--- a/lib/source-map-cache.js
+++ b/lib/source-map-cache.js
@@ -35,6 +35,51 @@ SourceMapCache.prototype.applySourceMaps = function (coverage) {
   return mappedCoverage
 }
 
+// Maps the coverage location based on the source map. Adapted from getMapping()
+// in remap-istanbul:
+// <https://github.com/SitePen/remap-istanbul/blob/30b67ad3f24ba7e0da8b8888d5a7c3c8480d48b2/lib/remap.js#L55:L108>.
+function mapLocation (sourceMap, location) {
+  var start = sourceMap.originalPositionFor(location.start)
+  var end = sourceMap.originalPositionFor(location.end)
+
+  /* istanbul ignore if: edge case too hard to test for */
+  if (!start || !end) {
+    return null
+  }
+  if (!start.source || !end.source || start.source !== end.source) {
+    return null
+  }
+  /* istanbul ignore if: edge case too hard to test for */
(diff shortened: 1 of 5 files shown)


bcoe:
@novemberborn quite happy to pull this in, would it be possible to add one or two tests to outline the difference in our `SourceMapCache` test suite?

bcoe:
@novemberborn saw your comment on #67 great work on this :+1: one test for the ignoring code functionality would be awesome, then let's get this out the door.

novemberborn:
@bcoe I've amended the commit with a test. Basically looking for the `skip` property on the appropriate statement.

My `source-map-fixtures` package is coming in handy for this!

jamestalmage:
@novemberborn I am done merging all my changes.
This LGTM, I vote for a merge as soon as you rebase.

novemberborn:
@jamestalmage cheers, done.

PR #93, opened 2015-12-14
`a` must sometimes be able to be a #

bcoe:


novemberborn:
I think your commit messages means "a must sometimes be a number"? Maybe? Could test for a string type instead of the `replace` method I suppose.

bcoe:
@novemberborn I like that approach better, covers our butt if argv ever has `null` or `undefined` in it -- I don't think this is possible, but better safe than sorry.

PR #98, opened 2015-12-15
Fix ability to self-test using `npm link`.

jamestalmage:
We had two problems. 

1. Our attempt to resolve the module passed to the `--require` 
flag was grabbing from `nyc/node_modules. A switch to `require-from` solved that.

2. We were collecting source-map information from every file required. This
caused bad things to happen when convert-source-map saw it's source-map-comment
regular expression and thought it was an actual source map comment. Only
collecting source map information when needed made things run faster and 
fixed the problem

### DIFF
--- a/index.js
+++ b/index.js
@@ -116,21 +116,6 @@ NYC.prototype.addFile = function (filename, returnImmediately) {
   }
 }
 
-NYC.prototype.addContent = function (filename, content) {
-  var relFile = path.relative(this.cwd, filename)
-  var instrument = this.shouldInstrumentFile(filename, relFile)
-
-  if (instrument) {
-    content = this.instrumenter.instrumentSync(content, './' + relFile)
-  }
-
-  return {
-    instrument: instrument,
-    content: content,
-    relFile: relFile
-  }
-}
-
 NYC.prototype.shouldInstrumentFile = function (filename, relFile) {
   relFile = relFile.replace(/^\.\//, '') // remove leading './'.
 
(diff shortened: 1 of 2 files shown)


jamestalmage:
@novemberborn @bcoe - This should finally fix up the remaining issues from bringing in `append-transform`. My original change made us too aggressive in collection source-map information. There is no need to do so for files that are not instrumented.

index.js
```diff
@@ -36,6 +37,8 @@ function NYC (opts) {
   var config = require(path.resolve(this.cwd, './package.json')).config || {}
   config = config.nyc || {}
 
+  this.requireTransform = this._requireTransform.bind(this)
```
novemberborn: Why bind here? `this.requireTransform` seems to be referenced only once?

index.js
```diff
@@ -155,15 +144,21 @@ NYC.prototype.addAllFiles = function () {
 }
 
 NYC.prototype._wrapRequire = function () {
-  var _this = this
+  appendTransform(this.requireTransform)
+}
 
-  appendTransform(function (compiledSrc, filename) {
-    _this.sourceMapCache.add(filename, compiledSrc)
+NYC.prototype._requireTransform = function (code, filename) {
```
novemberborn: Why not do this inline in `_wrapRequire()`?

novemberborn:
> My original change made us too aggressive in collection source-map information. There is no need to do so for files that are not instrumented.

Nice!

jamestalmage:
@novemberborn - updated with your feedback

novemberborn:
@jamestalmage LGTM!

bcoe:
Amazing debugging work :) I'll test against all my testing projects tonight, but sounds like you nailed it -- I'm excited that nyc has been trimmed down so much, to mostly its core spawn capturing funtionality.

bcoe:
@jamestalmage @novemberborn everyone feel free to ignore the AppVeyor tests for the time being, until we put the foot-work in to making them pass.

PR #101, opened 2015-12-16
Caching

jamestalmage:
So...
Caching makes a pretty big difference.

The AVA test suite went from `42`s to `32`s on my machine, even with this relatively trivial implementation. It generates a new directory: `cwd/.nyc_cache` and writes cached files there based on the `md5` hash of their pre-instrumented content.

It is probably worth checking out `cacha` for this.

@floatdrop, my impression of https://github.com/sindresorhus/ava/pull/189 was that it was still a WIP. Where is `cacha` currently? Is it ready for us to push into production? If not, I am happy to help any way I can.


floatdrop:
@jamestalmage `cacha` is ready, but I could not catch up with rebasing PR.

.gitignore
```diff
@@ -1,4 +1,5 @@
 .nyc_output
+.nyc_cache
```
sindresorhus: Could this be put in the home directory instead? It's annoying having to .gitignore this in every repo using nyc. ESLint does this by default: https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--cache-location

jamestalmage: > ESLint does this by default

It looks to me like ESLint's default behavior is similar to this one.

> Path to the cache location. Can be a file or a directory. If none specified .eslintcache will be used. The file will be created in the directory where the eslint command is executed.

I agree, it is annoying to have to ignore it every time. I am not sure defaulting to the home directory is the best choice though. One advantage of the annoying folder in your project's base directory is that you notice it, and it becomes obvious what it is for. Hiding it away in your home directory puts it out of mind when you are trying to troubleshoot problems. Maybe we could put it in `./node_modules/nyc/.nyc_cache`, that way they end up clearing the cache if they rimraf `node_modules` while troubleshooting.

jamestalmage: Or better yet, maybe we should put it in:

`./node_modules/.cache/nyc`, and encourage other modules to use `./node_modules/.cache/<module-name>` as well. Then `trash node_modules/.cache` becomes the de facto way to clear caches when troubleshooting.

jamestalmage:
@bcoe @novemberborn The self-coverage stuff is actually a pretty interesting profiling tool when used in conjunction with `npm link`. Since we automatically use `index.covered.js` if it exists, we can get a better idea how our code behaves IRL. All these screenshots were generated running the AVA test suite: 

### Comparing runs with an empty cache, vs ones with a complete cache

*First run, no cache hits:*

 We can see that out of 50 forked processes, it only becomes necessary to create an instrumenter in 3 of those processes (meaning that 47 forked processes simply pulled from the cache - even on the first run). This makes sense with the speedups I am seeing. You still get the majority of the benefit from caching even on your first run. The second run may be faster, but imperceptibly so.

screenshot 2015-12-16 23 11 05

Also notice that the instrumenter is only accessed 14 times. There are 14 files instrumented in AVA. This means we never had a race condition for the cache (two threads seeing a cache miss, instrumenting and then racing to write the result back). Since `tap` executes tests synchronously and serially, this makes sense. However, it means we aren't exercising a potential failure mode here. We will definitely need to find a way to simulate race conditions as part of our test suite.

*Second run, no cache misses:*

We can see that the instrumenter is never created. Just a validation that two subsequent runs with identical content never has a cache miss.

screenshot 2015-12-16 23 15 32

### Optimization

Also, using self coverage on a real world test suite helps create a heat map and identify the best places to focus optimization efforts.
 
screenshot 2015-12-16 23 53 00


screenshot 2015-12-16 23 47 32

test/src/nyc-test.js
```diff
@@ -161,7 +161,7 @@ describe('nyc', function () {
     })
 
     describe('custom require hooks are installed', function () {
-      it('wraps modules with coverage counters when the custom require hook compiles them', function () {
+      /* it('wraps modules with coverage counters when the custom require hook compiles them', function () {
```
jamestalmage: @novemberborn - we still have a few tests that test `index.js`, including this one. I believe you had a PR that fixed some of these.

index.js
```diff
@@ -153,8 +168,16 @@ NYC.prototype._wrapRequire = function () {
...
-    // now instrument the compiled code.
-    return _this.instrumenter.instrumentSync(code, './' + relFile)
+    var hash = md5(code)
+    var cacheFilePath = path.join(_this._cacheDirectory(), hash + '.js')
+
+    try {
+      return fs.readFileSync(cacheFilePath, 'utf8')
+    } catch (e) {
+      var instrumented = _this.instrumenter().instrumentSync(code, './' + relFile)
+      fs.writeFile(cacheFilePath, instrumented)
```
novemberborn: `writeFileSync()`?

novemberborn: I suppose it's a cache so you don't need to block the process nor do you care if it fails… maybe add a comment to that effect?

index.js
```diff
@@ -164,6 +187,7 @@ NYC.prototype.cleanup = function () {
 
 NYC.prototype._createOutputDirectory = function () {
   mkdirp.sync(this.tmpDirectory())
+  mkdirp.sync(this._cacheDirectory())
```
novemberborn: Not sure `_cacheDirectory` needs to be more private than `tmpDirectory`.

test/src/nyc-test.js
novemberborn: Yea I need to go back and see what's still needed…

index.js
```diff
@@ -11,6 +11,14 @@ var onExit = require('signal-exit')
 var stripBom = require('strip-bom')
 var SourceMapCache = require('./lib/source-map-cache')
 var resolveFrom = require('resolve-from')
+var crypto = require('crypto')
+
+function md5 (str) {
+  return crypto
+    .createHash('md5')
+    .update(str, 'utf8')
+    .digest('hex')
```
novemberborn: Personally I'd be happy with this on a single line.

Is md5 still acceptable for these kind of hashes?

index.js
jamestalmage: yes, but we pass allow `cacheDirectory` and `tempDirectory` options in the constructor. The collision is avoided with `temp` vs `tmp`, not sure how to pull that off elegantly here. Really I think we should rework option handling so that the option ends up being the same name as the accessor method.

index.js
jamestalmage: https://blog.risingstack.com/automatic-cache-busting-for-your-css/

> For this, since we care mostly about the hash value changing if images we provide have changed, let’s use MD5, a cryptographic hash function. While MD5 is not appropriate for verification of untrusted data, it does provide uniform distribution when truncated and if we use the first 32 bits, there will be a 1 in 3,506,097 chance of a collision if we have 50 revisions of the same file

via 

https://github.com/sindresorhus/rev-hash

index.js
jamestalmage: > I suppose it's a cache so you don't need to block the process

That was my thought, but honestly [I wasn't sure](https://github.com/jamestalmage/nyc/commit/ed1a9eeb235fefc1cd21c919af3b73e29f2414ac).

> maybe add a comment to that effect?

Sounds like a plan. Is there any danger to writing async though? Should I change it back? I do not know enough about how file system failures propogate. Also, as stated [here](https://gist.github.com/jamestalmage/758b09c42069ad96c792), it doesn't seem cache collisions are happening at all in the tests I've run, so I'm not even sure what behavior it will have. Definitely needs some more testing.

index.js
sindresorhus: MD5 is totally fine for non-cryptographically stuff.

jamestalmage: I couldn't find `rev-hash` when I wrote this, we could switch to that. I wonder, Is there any advantage to truncating to 10 characters in a temp folder?

// @sindresorhus

.gitignore
sindresorhus: That could work too.

index.js
sindresorhus: No.

https://github.com/sindresorhus/md5-hex

index.js
novemberborn: Yea, fair enough.

index.js
novemberborn: Nice write-up, thanks.

index.js
novemberborn: I'd assume the writes are consecutive, not parallel, so even if there is a collision it just means the cache file is written a couple times. It'll still be a valid cache file in the end. Errors won't be reported though, e.g. if we somehow screw up writing the cache dir we'd never notice. OTOH if we fail to write the cache file once in a while that's not the end of the world (and a test run shouldn't fail because of it).

I suppose this could use async with a `console.warn() if there are errors?

jamestalmage: > I suppose this could use async with a `console.warn()` if there are errors?

I think we are going to run into a lot more write collisions when they start using `nyc` with `AVA`. My tests to this point have been the `AVA` test suite (which is run with TAP). I think we only need to warn on `read` errors.

Perhaps we could log `write` errors with [`debug`](https://github.com/visionmedia/debug).

jamestalmage: Turns out `writeFileSync()` is required. It is why I have a failing test.

jamestalmage:
Just pushed another commit that squeezes out an additional 2 or 3 seconds from the AVA test suite.

It works by caching source-map information as well. Saved only once when it sees a cache miss for the source. If there is a cache hit, it does not even extract the source map (source map extraction can be expensive for large files). It does not apply source map information when exiting the process, but instead adds a `contentHash` key to each files coverage report. That value is used to load the source-map from disk whenever `report` is run.

The only downside to this is that you can no longer use just the contents of `.nyc_output/*.json` to generate coverage (you need `node_modules/.cache/*.map`).

jamestalmage:
Some stats from running the AVA test suite: 

- | run time | coverage penalty | reduction in penalty
----|----|---|---
no coverage                 |  25 sec  |    -        | -
nyc master                      |  42 sec  | 17 sec | -
cache instrumentation (1d625e8a7ff8)  |  32 sec  |  7 sec  | 59%
cache source maps (9b014bbc22)     | 30 sec  | 5 sec  | 29% (70% overall)

jamestalmage:
The latest commit drops `lodash` from `index.js`. I did not remove it from `source-map-cache`, but that is only loaded once per execution for running reports.

That dropped the coverage penalty from `5s` to `4s` on the AVA test suite.

`lodash` is a 12,000 line monster that is slow to eval. The only function we were using in `index.js` that is not available natively was `string.endsWith`

jamestalmage:
See: https://github.com/jamestalmage/nyc/commit/f14f98fe07dc1301f6aa5b88475c3ece9062b86b

It lazy-loads any dependency that we are not guaranteed to need in a forked process, that includes:

* `convert-source-map` - not needed if everything is cached
* `rimraf` && `mkdirp` - only needed in the top level process (required moving `mkdirp` execution to `cleanup` method, which is not the right spot for it)
* `glob` - only needed for `--all` option
* `stripBom` - only needed for `addFile()`, which appears only to be needed for the `--all` option

That appears to shave of *some* time. The AVA test runs go from a shaky 29s/30s to a solid 29s on every run. It breaks a bunch of tests so I am going to hold off on it until this is merged and we develop some more accurate benchmarking.

I think I am reaching the limits of using AVA's test suite and `npm run test` as a viable benchmarking tool. It only provides a resolution of 1 second, and at most there are only 4.5 more seconds of optimizations we can squeeze from that test suite. Also, with it's sizable dependency graph, previous benchmarks become useless with each `npm update`. Anybody got any clever ideas for continuing to optimize this?

jamestalmage:
Shaved off another few seconds.
Instead of defaulting to a glob pattern of `**`, I just shortcut that manually with 

```js
if (!this.include || micromatch.any(filename, this.include)...
```

index.js
```diff
@@ -18,37 +20,31 @@ if (/index\.covered\.js$/.test(__filename)) {
...
-  this.include = config.include || ['**']
-  this.include = this._prepGlobPatterns(this.include)
+  this.include = false
+  if (config.include) {
+    this.include = this._prepGlobPatterns(arrify(config.include))
+  }
 
-  this.exclude = ['**/node_modules/**'].concat(config.exclude || ['test/**', 'test{,-*}.js'])
-  if (!Array.isArray(this.exclude)) this.exclude = [this.exclude]
+  this.exclude = ['**/node_modules/**'].concat(arrify(config.exclude || ['test/**', 'test{,-*}.js']))
```
novemberborn: Could call `this._prepGlobPatterns()` here rather than reassigning `this.exclude`?

index.js
```diff
@@ -67,14 +63,18 @@ NYC.prototype._loadAdditionalModules = function () {
...
 NYC.prototype._createInstrumenter = function () {
   var configFile = path.resolve(this.cwd, './.istanbul.yml')
 
   if (!fs.existsSync(configFile)) configFile = undefined
 
-  var instrumenterConfig = this.istanbul.config.loadFile(configFile).instrumentation.config
+  var instrumenterConfig = this.istanbul().config.loadFile(configFile).instrumentation.config
 
-  return new this.istanbul.Instrumenter({
+  return new (this.istanbul()).Instrumenter({
```
novemberborn: Assign `this.istanbul()` within this function?

index.js
```diff
@@ -85,41 +85,41 @@ NYC.prototype._createInstrumenter = function () {
...
+    if (result.indexOf(pattern) === -1) {
+      result.push(pattern)
+    }
+  }
+
+  patterns.forEach(function (pattern) {
     // Allow gitignore style of directory exclusion
-    if (!_.endsWith(pattern, '/**')) {
-      directories.push(pattern.replace(/\/$/, '').concat('/**'))
+    if (!endsWith(pattern, '/**')) {
```
novemberborn: Why not `!/\/\*\*$/.test(pattern)` and do away with the `ends-with` dependency?

index.js
```diff
@@ -85,41 +85,41 @@ NYC.prototype._createInstrumenter = function () {
...
+      add(pattern.replace(/\/$/, '').concat('/**'))
```
novemberborn: Apparently [`String#concat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#Performance) is slow.

Also, apparently it exists! :wink:

novemberborn: (And yes you didn't write `concat()`. But since we're speading things up :stuck_out_tongue_winking_eye: )

index.js
```diff
@@ -85,41 +85,41 @@ NYC.prototype._createInstrumenter = function () {
...
-    return {}
-  }
-
+  var source = stripBom(fs.readFileSync(filename, 'utf8'))
+  var content = this._addSource(source, filename, relFile)
   return {
-    instrument: instrument,
-    content: content,
-    relFile: relFile
+    instrument: !!content,
```
novemberborn: This is pretty subtle (as is `content || source` below).

index.js
```diff
@@ -141,20 +141,38 @@ NYC.prototype.addAllFiles = function () {
   this.writeCoverageFile()
 }
 
+var hashCache = {}
```
novemberborn: I suppose it doesn't matter in practice but may be better to have this on the instance?

index.js
```diff
@@ -212,24 +241,58 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
   reporter.write(collector, true, cb)
 }
 
+var loadedMaps = {}
```
novemberborn: Same concern here as for `hashCache`.

index.js
```diff
@@ -141,20 +141,38 @@ NYC.prototype.addAllFiles = function () {
   this.writeCoverageFile()
 }
 
+var hashCache = {}
+
+NYC.prototype._addSource = function (code, filename, relFile) {
```
novemberborn: If you rename this to `_maybeInstrumentSource()` it's easier to see why it may return `null`. Then you could assign to a `instrumentedSource` variable and the logic in `addFile()` becomes more clear.

index.js
```diff
@@ -212,24 +241,58 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
...
 }
 
+var loadedMaps = {}
+
 NYC.prototype._loadReports = function () {
   var _this = this
   var files = fs.readdirSync(this.tmpDirectory())
 
-  return _.map(files, function (f) {
+  var SourceMapCache = require('./lib/source-map-cache')
```
novemberborn: Did you move the require statement down here because `source-map-cache` pulls in `lodash`? Because we could probably remove it there as well.

index.js
```diff
@@ -212,24 +241,58 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
...
-      return JSON.parse(fs.readFileSync(
+      report = JSON.parse(fs.readFileSync(
         path.resolve(_this.tmpDirectory(), './', f),
         'utf-8'
       ))
     } catch (e) { // handle corrupt JSON output.
       return {}
     }
+
+    if (report) {
```
novemberborn: This condition seems unnecessary. If the file no longer exists, or did not contain valid JSON, we'll already have returned. The other option is that it contains valid JSON but it's not an object. There's plenty other ways the report could be invalid for the code that follows.

index.js
```diff
@@ -212,24 +241,58 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
...
+      Object.keys(report).forEach(function (relFile) {
+        var fileReport = report[relFile]
+        if (fileReport && fileReport.contentHash) {
+          var hash = fileReport.contentHash
+          if (!(loadedMaps[hash] || (loadedMaps[hash] === false))) {
```
novemberborn: Why the second `=== false` comparison?

novemberborn: Oh wait that prevents multiple attempts to read the source map. Maybe rewrite to `if (loadedMaps[hash] !== false && !loadedMaps[hash])` so that's a bit more obvious?

novemberborn: Or just `!(hash in loadedMaps)`.

index.js
```diff
@@ -212,24 +241,58 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
...
+            try {
+              var mapPath = path.join(cacheDir, hash + '.map')
+              loadedMaps[hash] = fs.readFileSync(mapPath, 'utf8')
+            } catch (e) {
+              loadedMaps[hash] = false
```
novemberborn: Could add a comment "set to false to avoid repeatedly trying to load the map"

index.js
```diff
@@ -212,24 +241,58 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
...
+      })
+      report = sourceMapCache.applySourceMaps(report)
+    }
+    return report
   })
 }
 
-NYC.prototype.tmpDirectory = function () {
-  return path.resolve(this.cwd, './', this.tempDirectory)
+NYC.prototype.tmpDirectory = NYC.prototype.tempDirectory = function () {
```
novemberborn: `tempDirectory()` is now preferred, right? Worth deprecating `tmpDirectory()?` Regardless would be good to add a comment as to why there's two options. Digging through the commit history to find out later will be painful.

lib/source-map-cache.js
```diff
@@ -15,6 +15,10 @@ SourceMapCache.prototype.add = function (filename, source) {
   if (sourceMap) this.cache['./' + path.relative(this.cwd, filename)] = new SourceMapConsumer(sourceMap.sourcemap)
 }
 
+SourceMapCache.prototype.addMap = function (relFile, mapJson) {
```
novemberborn: No longer need `SourceMapCache#add()` right?

novemberborn:
@jamestalmage 

> Just pushed another commit that squeezes out an additional 2 or 3 seconds from the AVA test suite.
> 
> It works by caching source-map information as well.

Very cool!

> The only downside to this is that you can no longer use just the contents of .nyc_output/*.json to generate coverage, you also need node_modules/.cache/*.map.

That's OK I think. If you have source maps then you've done transpiling so you're used to doing a bit more effort.

> It lazy-loads any dependency that we are not guaranteed to need in a forked process, that includes:
> […]
> That appears to shave of some time. The AVA test runs go from a shaky 29s/30s to a solid 29s on every run.

Is that really worthwhile? As far as understanding code goes I'd prefer dependencies listed at the top, rather than sprinkled throughout.

lib/source-map-cache.js
jamestalmage: Except in your tests I think.

index.js
jamestalmage: `tmpDirectory` is there for backwards compatibility. I don't think anybody is using that outside us, and it certainly is not documented. I think it's safe just to drop the other one.

index.js
jamestalmage: yep

jamestalmage:
> Is that really worthwhile?

Not unless I can prove it's actually optimizing something.  That's why I pushed it to a different branch and abandoned it for the time being.

jamestalmage:
Another attempted optimization that I couldn't prove worked: https://github.com/jamestalmage/nyc/commit/7f80c30399d2ef2a98a3f8b7261f191fdf342ccb

Tried to use `minimatch.matcher()` to precompile patterns, did not notice an improvement. That one might still be worth pursuing. It falls inside our require hook (that is executed many thousands of times), and the minimatch documentation says it is better to precompile them that way. The documented optimization is for a single pattern, maybe I optimized the array of patterns incorrectly.

jamestalmage:
@novemberborn @bcoe  @sindresorhus  - it is probably worth you guys putting a second/third/fourth set of eyeballs on the `append-transform` code to see if I am missing any optimizations. It is called 10,000+ times in the AVA tests suite alone.

jamestalmage:
@novemberborn - incorporated your feedback. I am using `lazy-req` now. Provides the same speed benefits as lazily requiring inline, but pushes your list of dependencies to the top.

Not sure why the coverage dropped for that last commit (it was higher previously).

index.js
```diff
@@ -214,22 +250,54 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
...
+      }
+    })
+    report = sourceMapCache.applySourceMaps(report)
+    return report
   })
 }
 
-NYC.prototype.tmpDirectory = function () {
-  return path.resolve(this.cwd, './', this.tempDirectory)
+NYC.prototype.tempDirectory = function () {
```
novemberborn: :+1: dropping `tmpDirectory()`.

index.js
```diff
@@ -141,29 +149,46 @@ NYC.prototype.addAllFiles = function () {
...
+    return _this._maybeInstrumentSource(code, filename, relFile) || code
   })
 }
 
 NYC.prototype.cleanup = function () {
-  if (!process.env.NYC_CWD) rimraf.sync(this.tmpDirectory())
+  if (!process.env.NYC_CWD) rimraf.sync(this.tempDirectory())
 }
 
 NYC.prototype._createOutputDirectory = function () {
```
novemberborn: This is a bit of a misnomer now.

index.js
```diff
@@ -214,22 +250,54 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
...
+            loadedMaps[hash] = false
+          }
+        }
+        if (loadedMaps[hash]) {
+          sourceMapCache.addMap(relFile, loadedMaps[hash])
+        }
+      }
+    })
+    report = sourceMapCache.applySourceMaps(report)
+    return report
```
novemberborn: Could just return the result of `applySourceMaps(report)`

novemberborn:
@jamestalmage 

> Not unless I can prove it's actually optimizing something. That's why I pushed it to a different branch and abandoned it for the time being.

Cool.

> Another attempted optimization that I couldn't prove worthwhile: jamestalmage@7f80c30
> 
> Tried to use minimatch.matcher() to precompile patterns, did not notice an improvement. That one might still be worth pursuing. It falls inside our require hook (that is executed many thousands of times), and the minimatch documentation says it is better to precompile them that way. The documented optimization is for a single pattern, maybe I optimized the array of patterns incorrectly.

I'm just guessing here but that code path looks like it might be optimized quite well by V8. IIRC it also caches regex compilation.

> incorporated your feedback. I am using lazy-req now. Provides the same speed benefits as lazily requiring inline, but pushes your list of dependencies to the top.

That's nicer though if the performance penalty isn't too high I'd still rather require them normally. There's now extra weirdness in using `SourceMapCache`.

> it is probably worth you guys putting a second/third/fourth set of eyeballs on the append-transform code to see if I am missing any optimizations

Nothing jumps out at me other than <https://github.com/jamestalmage/append-transform/commit/73f7abaee94c134575cf08be29943da834ccafc5#commitcomment-15067878>.

---

I'll try and push up some commits removing `lodash` usage from `source-map-cache` and switch the tests to use `addMap()`.

novemberborn:
> I'll try and push up some commits removing lodash usage from source-map-cache and switch the tests to use addMap().

@jamestalmage see https://github.com/jamestalmage/nyc/pull/1 :)

jamestalmage:
OK, I merged @novemberborn's commits simplifying `source-map-cache` and reverted the `lazy-req` stuff since lodash is only a dev-dependency now.

I think this PR is in a good place and ready for a merge.

sindresorhus:
:+1: I'm in no way familiar enough with the code base to be able to review it, but from a quick shim of the changes it looks pretty good to me. Excited about the perf improvements! Super nice work @jamestalmage :)

jamestalmage:
Just added a test to create cache collisions. It spawns 4 processes (all wrapped by `nyc`), and coordinates an attempt to `require` the same file at exactly the same time using `process.hrtime`.

I added some logging statements to our cache hook (not included in the commit), and I can verify that all 4 processes see a cache miss (the coordination worked). All 4 processes also successfully write back to the cache, no error is thrown even though they are all trying to write to the same file. It does not appear we have to worry about write collisions, and that competing `fs.writeFileSync(..)` calls will just overwrite each-other, and the last one "wins". Since we know each competing process is trying to write identical content, I don't think we need to coordinate any further.

sindresorhus:
> It does not appear we have to worry about write collisions, and that competing fs.writeFileSync(..) calls will just overwrite each-other

That's not [my experience](https://github.com/yeoman/configstore/issues/20). I don't think it locks the file unless you do `fs.open`, though that might have changed in more recent Node.js versions.

jamestalmage:
I have tried a few more iterations of my test with various changes trying to cause a problem, but I can't. Even in Node 0.10. I don't know how `configstore` works, but [this comment](https://github.com/yeoman/configstore/issues/20#issuecomment-125224909) makes me think the file interactions are more complex than what we are doing here:

> In casual testing, though, the race condition I hit more often was **caused by the read-mutate-write sequences** in set and del – competing processes end up clobbering one another's changes when those steps interleave, so you get data loss. Not as dire as straight-up corrupting the file, but has a way bigger window it can happen in, so ¯\_(ツ)_/¯

We don't care about the previous contents of the file, we just clobber with data we know to be good. There is no *read-mutate-write* sequence

> I don't think it locks the file unless you do `fs.open`

It seems crazy to me that `fs.writeFileSync` would not lock the file and try to make writes as atomic as possible. Either way, it is frustrating that I can't find much information on how `fs` operations work at a lower level. Somebody needs to do a blog post... I guess digging into the Node source is always an option.

sindresorhus:
> We don't care about the previous contents of the file, we just clobber with data we know to be good. 

Good point. Guess it's a non-issue in this case then.

> It seems crazy to me that fs.writeFileSync would not lock the file and try to make writes as atomic as possible.

That's just what many people have told me. No idea if it's actually accurate.

> It seems crazy to me that fs.writeFileSync would not lock the file and try to make writes as atomic as possible. Either way, it is frustrating that I can't find much information on how fs operations work at a lower level. Somebody needs to do a blog post... I guess digging into the Node source is always an option.

Can you open an issue on Node.js about better docs regarding that? I would definitely like to see that too.

bcoe:
@jamestalmage I like the idea of caching for large projects that benefit from it.

It does, however, introduce quite a few more moving parts. I've learned making nyc run in as many setups and environments as possible (*cough* Windows) that every moving part is going to be more edge-cases for us to debug on various outlier platforms....

So, my ask for caching is that we make it a configuration option, and my temptation in `5.x.x` is to default with it to `false`.

Would love to hear other people's thoughts.

jamestalmage:
> large projects that benefit from it.

The difference is noticeable even with the 50 line module I am working on now.

> So, my ask for caching is that we make it a configuration option, and my temptation in 5.x.x is to default with it to false.

I think your concerns are reasonable, especially before we get AppVeyor up and running. I think we should eventually turn it on by default once we see green from AppVeyor, and we have dogfooded it for a while. A feature toggle is certainly a good idea as it gives us a way to eliminate the cache implementation as the source of reported bugs.

bcoe:
@jamestalmage cool, I think we're on the same page :+1: turning this on by default just feels a fairly big change for a `minor` release, I'd rather let it bake as an optional setting for a bit, then bump to `6.x.x` with it enabled by default.

novemberborn:
> So, my ask for caching is that we make it a configuration option, and my temptation in 5.x.x is to default with it to false.

:+1:

PR #102, opened 2015-12-17
add profiling instructions

jamestalmage:
This adds a `maintaining.md` file where we can store information helpful to maintainers.

The first entry is a set of instructions for self-profiling, based on my comments [here](https://github.com/bcoe/nyc/pull/101#issuecomment-165337057).

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -1,0 +1,80 @@
+
+## Profiling:
+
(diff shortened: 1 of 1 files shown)


bcoe:
@jamestalmage this is great :+1:

My only thought, perhaps we can move it into a `/docs` directory, I bet we'll have more documents like this we add over time.

bcoe:
@jamestalmage if you don't mind moving this into a `/docs` folder, and rebasing, let's land this.

PR #104, opened 2015-12-18
Do fix remaining tests

### LINKED ISSUE #103: Fix remaining tests.
There are a few tests I think I did a poor adapting to fit our new testing strategy. @novemberborn had an alternate test strategy PR that we didn't go with, but I think it did adapt a few of the tests better.

We need to make sure we identify those and merge them in.

See: 
  - https://github.com/bcoe/nyc/pull/85
  - https://github.com/bcoe/nyc/pull/101#discussion_r47897033

novemberborn:
See commits. Fixes #103.

@jamestalmage

### DIFF
--- a/test/fixtures/check-instrumented.js
+++ b/test/fixtures/check-instrumented.js
@@ -1,0 +1,7 @@
+function probe () {}
+
+// When instrumented there will be references to variables like
+// __cov_pwkoI2PYHp3LJXkn_erl1Q in the probe() source.
+module.exports = function () {
+  return /\b__cov_\B/.test(probe + '')
+}
--- a/test/fixtures/spawn.js
+++ b/test/fixtures/spawn.js
@@ -1,0 +1,5 @@
+#!/usr/bin/env node
+
+var spawn = require('child_process').spawn
+spawn(process.execPath, ['sigint.js'], { cwd: __dirname })
+spawn(process.execPath, ['sigterm.js'], { cwd: __dirname })
--- a/test/src/nyc-test.js
+++ b/test/src/nyc-test.js
@@ -235,7 +235,7 @@ describe('nyc', function () {
         cwd: fixtures
       })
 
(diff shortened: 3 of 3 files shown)


jamestalmage:
:+1: 

I think there are still a couple tests that use the require hook, delete from the require cache, and then require `index.js`. That seems like a bad idea. Can those be replaced with fixture tests as well?

bcoe:
:shipit:

novemberborn:
> I think there are still a couple tests that use the require hook, delete from the require cache, and then require index.js. That seems like a bad idea. Can those be replaced with fixture tests as well?

@jamestalmage @bcoe pushed a commit which does that now.

bcoe:
Great work. :shipit:, I'll start looking more into Windows tests and help as best I can this weekend.

novemberborn:
`master` is failing now, though it was OK in this PR. @bcoe @jamestalmage any idea? See <https://travis-ci.org/bcoe/nyc/builds/97862828>.

bcoe:
@novemberborn it looks like `_compile` expects the `filename` as the second argument to compile, but it's not being passed in -- I'm not quite sure why this was working before.

`append-transform` overrides _compile, perhaps the behavior of the shim isn't quite identical? any ways, quick fix: pass the filename. but, I'd love to know why this did work.

jamestalmage:
[This change in `append-transform`](https://github.com/jamestalmage/append-transform/commit/ad3d22b1a60ded33b666aad4a20a23866c36dd42#diff-168726dbe96b3ce427e7fedce31bb0bcR50) is why that broke. It was previously masking the bad behavior of require extensions that did not pass the filename as the second arg (by ignoring it and using the filename it had seen previously).

I originally just masked that bad behavior because it seemed convenient (if you were writing a transform you knew would always be on top of `append-transform`, it seemed unnecessary to have to pass that around), but then I realized that would mean masking errors for people developing their own transforms and testing with `nyc`.

PR #106, opened 2015-12-19
module._compile() correctly expects a second filename argument

bcoe:
see #104

CC: @novemberborn

### DIFF
--- a/test/src/nyc-test.js
+++ b/test/src/nyc-test.js
@@ -157,7 +157,7 @@ describe('nyc', function () {
     describe('custom require hooks are installed', function () {
       it('wraps modules with coverage counters when the custom require hook compiles them', function () {
         var hook = sinon.spy(function (module, filename) {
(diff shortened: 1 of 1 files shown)


jamestalmage:
technically, it's not `append-transform`. That require extension was always wrong. The signature for `module._compile` is `module._compile(code, filename)` and both arguments are required. `append-transform` used to silently fix it for you by ignoring whatever `filename` argument you passed in and using the correct value. Without `append-transform` installed that require extension would always have failed. I think the solution previous to `append-transform` behaved the same way.

bcoe:
@jamestalmage updated description :+1: 

I've tested this branch using npm link with: `nyc`, `yarsay`, `yargs`, `mkdirp`, `uuid`, `pageres` everything is working great with `npm link nyc` -- amazing work \o/

I'm going to officially start playing more with Windows. I think the test isolation will help a ton.

PR #107, opened 2015-12-20
get tests passing on Windows

### LINKED ISSUE #81: Windows support
Ticket to track progress. @bcoe should probably flesh this out more :smiley_cat:

bcoe:
This wasn't as much work as I'd feared, fixes #81 \o/

screen shot 2015-12-19 at 10 14 35 pm

Here's a break down of the changes I've made:

* I've added the `appveyor.yml` that @sindresorhus recommended.
* I switched to using `rimraf` for deleting left over test data, since it's cross-platform.
* windows doesn't (or at least `signal-exit` on Windows) doesn't support `process.kill()`, I've disabled the two tests that require `process.kill` using the `is-windows` module.
  * I've also made the spawn tests not rely on scripts that call `process.kill`.
* I was running into issues with `path.sep` being `\` on Windows and `/` on *nix. 
  * `nyc` and `Istanbul` have paths that are all over the place on Windows (sometimes using one separator, sometimes another).
  * rather than try to unravel this mess (since it doesn't break anything but tests) I've written the helper module `any-path`: https://www.npmjs.com/package/any-path

This work would not have been possible if it wasn't for the hard work of @jamestalmage @novemberborn , refactoring and fixing up tests.

Thank you everyone for all the hard work on this module recently \o/

Reviewers: @jamestalmage @novemberborn

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 [![Build Status](https://travis-ci.org/bcoe/nyc.png)](https://travis-ci.org/bcoe/nyc)
 [![Coverage Status](https://coveralls.io/repos/bcoe/nyc/badge.svg?branch=)](https://coveralls.io/r/bcoe/nyc?branch=)
 [![NPM version](https://img.shields.io/npm/v/nyc.svg)](https://www.npmjs.com/package/nyc)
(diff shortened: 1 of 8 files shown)


novemberborn:
Nice! :shipit:

sindresorhus:
:shipit:

PR #108, opened 2015-12-20
Caching 2.0

jamestalmage:
see #101 for historical conversation.

-------------

## First, An Apology:

This is PR #101 rebased onto master. The rebase was not entirely smooth, and getting to a perfect git history has proven fairly difficult. My apologies to reviewers, but I think this is about as good as it is going to get.
 
I have squashed the initial ~27 commits down to 14. The commit messages are now very detailed, so I encourage you to step through them one by one. 

Most of the commits are logical units of work, but there are two unfortunate "fixup" commits, that I would have preferred squashing with earlier commits, but reordering the git history to accommodate proved too difficult to be worth it.

A number of the commit messages contain a "Future Changes" section, where I explain what eventually is changed by the "fixup" commits. Not a perfect solution, but hopefully helpful to you as you review.

The meat of the work can be found in the 5 commits linked in the benchmarking section below. Much of those are refactored in future commits, based on feedback received here and in the #101 PR.

---

## Overview:

This PR greatly improves the speed of `nyc` via a number of strategies, the primary one being caching.

- It caches the results of `istanbul` transforms.
- It caches the inline source-maps it extracts before instrumentation. Instead of remapping coverage reports as they are saved to disk, it instead uses the cached source-map information to remap coverage when it is loaded from disk when a report is requested.
- It drops our runtime dependency on `lodash`, which was expensive to `require`.
- It defers `require`ing instanbul until it is actually needed (i.e. there is a cache miss). `istanbul` is also expensive to require.
- A minor optimization of how we handle globs.

---

## Benchmarks

I decided to use the AVA test suite as my primary benchmark. The AVA test suite forks 50 some processes, nearly all of which collect coverage information. This makes it an ideal candidate for benchmarking `nyc`s performance. I have tested on much simpler projects, and the improvements are noticeable even in projects that only fork one additional process (though I suspect that has to do more with simplifying our dependency graph than caching).

I use the term "coverage penalty" to mean how much additional time the test suite takes to execute  when `nyc` is in use. This provides a better indication of the significance of each optimization.

- | run time | coverage penalty | reduction in penalty
----|----|---|---
no coverage                 |  25 sec  |    -        | -
nyc master                      |  42 sec  | 17 sec | -
cache instrumentation (3f58e88)  |  32 sec  |  7 sec  | 59%
cache source maps (2125be0)     | 30 sec  | 5 sec  | 29% (70% overall)
drop lodash (0a04c3b) | ~29 sec | 4 sec | 20% (76% overall)
shortcut include glob (7452b40) | 28 sec | 3 sec | 25% (82% overall)


bcoe:
@jamestalmage give this a shot:

https://www.npmjs.com/package/win-fork

jamestalmage:
I found at least part of the problem. See #109

test/src/nyc-test.js
```diff
@@ -430,4 +434,25 @@ describe('nyc', function () {
       return done()
     })
   })
+
+  describe('cache', function () {
+    it('handles collisions', function (done) {
+      var nyc = new NYC({cwd: fixtures})
+      nyc.clearCache()
+
+      var args = isWindows ? [bin] : [bin, process.execPath]
```
jamestalmage: This makes me thinks we are making a mistake in `mungeArgs` when it coms to windows.

jamestalmage:
@novemberborn - This test suite now makes it all the way through to source-maps. Some of the tests you merged into my branch are broken on Windows. It looks like it may just be a matter of different path separators being used.

bcoe:
@jamestalmage make sure you wrap all the appropriate objects with this helper:

https://github.com/bcoe/any-path

YOLO.

jamestalmage:
> make sure you wrap all the appropriate objects with this helper

I am OK with that as a temporary solution. I think it would be better if the [istanbul spec](https://github.com/gotwarlost/istanbul/blob/master/coverage.json.md) were changed to insist on posix paths.

test/src/source-map-cache.js
```diff
@@ -35,10 +35,16 @@ try {
...
+  var source = fixture.contentSync()
+  var sourceMap = convertSourceMap.fromSource(source) || convertSourceMap.fromMapFileSource(source, fixture.relpath)
+  if (sourceMap) {
+    sourceMapCache.addMap(fixture.relpath, sourceMap.sourcemap)
+  }
 })
 
-var coverage = ap(require('../fixtures/coverage'))
+var getReport = function () {
+  return _.cloneDeep(require('../fixtures/report'))
```
bcoe: I think if you switch this line to:

```js
return ap(_.cloneDeep(require('../fixtures/report')))
```

I wouldn't complain if istanbul only stored posix-style paths, it's a bit of an edge-case though -- both `\`, and `/` work on Windows, it's just biting us in the lookups we're doing for tests.

jamestalmage: Looks like another place I screwed up the rebase.

> posix-style paths

yeah, might not even make sense to request that change. Since the spec demands absolute paths anyways, that would make windows paths look really weird: `C:/foo/bar`

bcoe:
@jamestalmage I can figure out these annoying path issues in my Windows VM, if you want to work on the logic for making caching toggled via a CLI flag.

jamestalmage:
I would be incredibly happy to punt that one to you!!

bcoe:
@jamestalmage if you chery-pick this change:

https://github.com/bcoe/nyc/commit/25eb6a78748374e6f5489b856c48a544dd2c9238

I think it will get your tests passing.

jamestalmage:
woot! green tests

jamestalmage:
OK, just pushed an `enableCache` property on the NYC instance. With it hard coded to false, the tests pass locally.

How do we want to enable/disable the cache?

bcoe:
@jamestalmage my preference would be a flag like this:

https://github.com/bcoe/nyc/blob/master/bin/nyc.js#L72

index.js
```diff
@@ -214,22 +272,56 @@ NYC.prototype.report = function (cb, _collector, _reporter) {
...
       ))
     } catch (e) { // handle corrupt JSON output.
       return {}
     }
+
+    if (!_this.enableCache) {
+      return report
+    }
+
+    Object.keys(report).forEach(function (relFile) {
```
bcoe: This logic is fairly complicated, I'd be tempted to pull it into a logically named "private" method (prefix with `_`).

bcoe:
@jamestalmage since caching is defaulting to false now, I think we suddenly stopped covering quite a bit of the caching logic -- hence coveralls yelling.

It would probably be good to add a couple explicit tests here:

https://github.com/bcoe/nyc/pull/108/files#diff-7332f11ea438d83b6d316f2064202ebbR442

Creating the `nyc` instance like so:

```js
new NYC({enableCache: true})
```

Another nit, there's quite a bit of refactoring work on this branch, that I think isn't directly related to the cache logic?

https://github.com/bcoe/nyc/pull/108/files#diff-168726dbe96b3ce427e7fedce31bb0bcR107
https://github.com/bcoe/nyc/pull/108/files#diff-168726dbe96b3ce427e7fedce31bb0bcL21

This makes it hard for me to differentiate the logic related to the new cache, vs. other optimizations -- making my code-review less directed, and making me more gun-shy about merging.

In a perfect world, we'd land these unrelated changes in separate patches -- this also helps communicate in the git-history, the reasoning behind each change (they all seem logical from what I'm reading).

index.js
bcoe: an alternative thought, perhaps we try to pull the caching logic into `/lib/cache.js`, like we did for `source-map-cache.js` .

Keep `index.js` as thin as possible.

jamestalmage:
> Another nit, there's quite a bit of refactoring work on this branch that I think isn't directly related to the cache logic

Yeah, I let it get away from me. I could try and squash into a more logical series of commits, then cherrypick into multiple PR's. It might be a pain though. The Windows changes broke a number of things on this branch, and my rebase was less than "smooth"

bcoe:
@jamestalmage perhaps, since there are so many commits from our Windows changes, etc, just amend the git history at some point on this branch, and squash it into a few logical steps?

bcoe:
I'd say, let's shoot for this:

1. get everything working.
2. do an interactive squash on this branch and break the changes into logical steps.
3. update the issue description to provide a description of each refactor, and a thorough description of the caching logic.
4. loop in a couple more code reviewers, once everything is looking ready.
5. :shipit:

jamestalmage:
> It would probably be good to add a couple explicit tests here:

I opted to just run all the tests twice, both ways. There are probably a number of tests where it doesn't make sense to do the duplication (i.e. `_prepGlobPatterns`, etc). But I think we should explicitly move those to another test file that does not get run twice. Otherwise I think we end up in situations where we are running lots of duplicate tests, or are not catching corner cases that are different with/without caching.

bcoe:
@jamestalmage running twice seems reasonable to me. Good way to thoroughly test the cache logic :+1: 

Looks like we just need to figure out a good way to toggle the value -- could you simply generate two copies of each test during the build phase? Maybe I'm blanking on an easier solution for Windows, since I'm tired :)

jamestalmage:
> could you simply generate two copies of each test during the build phase

working on that now.

test/src/nyc-test.js
```diff
@@ -405,7 +421,7 @@ describe('nyc', function () {
       })
       var report = reports[0]['./test/fixtures/not-loaded.js']
 
-      reports.length.should.equal(1)
+      reports.length.should.equal(enableCache ? 2 : 1)
```
jamestalmage: I had to add this because the tests don't clean up after themselves between runs.
They really should.
After we merge this, we should just add `rimraf .... all the output dirs...` to the top of the test file. That acts like a `beforeEach` will get us a clean state for every test.

If that breaks a test because it needed a specific output from a previous test, it should instead explicitly spawn the CLI. That makes our tests much more explicit.

jamestalmage:
@bcoe - OK, I think I've squashed it down as much as I can without things getting painful. It's still not the cleanest history, but reducing further would require a lot of work. I should still go back and add some more detailed commit messages (it is currently just the default squash result - concatenation of all the squashed messages).

@novemberborn - I just squashed your contributions into a single commit.

jamestalmage:
@bcoe - I guess I am a little unclear on how I should use spawn wrap to propagate the cache setting to child processes. Set a `process.env` variable?

bcoe:
@jamestalmage sorry for the slow reply, Mondays are pretty busy here. 

If you look at `./bin/nyc.js` you'll see a line near the top that looks something like this:

`process.env.NYC_REQUIRE`

All of the configuration options that are set via a flag are passed through spawn-wrap as `NYC_` prefixed environment variables.

jamestalmage:
Yep - already got it: https://github.com/jamestalmage/nyc/commit/c604d94191353e0b4f8c13556d9c90e0d40b67ca#diff-806ac12b28e9cd571152f80071cf50b3R16

jamestalmage:
OK @bcoe @novemberborn - See updated description at the top, and updated commit messages. I think I'm as ready as I can get!

bcoe:
@jamestalmage thanks for the thorough updated description :smile: I'll get you some feedback soon, and will test it on my collection of pet projects.

index.js
```diff
@@ -141,29 +156,57 @@ NYC.prototype.addAllFiles = function () {
...
 }
 
-NYC.prototype._createOutputDirectory = function () {
-  mkdirp.sync(this.tmpDirectory())
+NYC.prototype.clearCache = function () {
+  rimraf.sync(this.cacheDirectory())
+}
+
+NYC.prototype._createDatastoreDirectories = function () {
+  mkdirp.sync(this.tempDirectory())
```
novemberborn: This now only creates the temp directory, since `cache-transform` takes care of creating the cache directory.

index.js
novemberborn: :+1:

novemberborn:
This seems to break `node test/fixtures/_generateReport.js`:

```
./nyc/index.js:92
  var instrumenterConfig = istanbul.config.loadFile(configFile).instrumentation.config
                                          ^

TypeError: Cannot read property 'loadFile' of undefined
    at NYC._createInstrumenter (./nyc/index.js:92:43)
    at NYC.instrumenter (./nyc/index.js:82:59)
    at NYC._transformFactory (./nyc/index.js:171:27)
    at transform (./nyc/node_modules/caching-transform/index.js:40:19)
    at NYC.transform (./nyc/node_modules/caching-transform/index.js:52:11)
    at NYC._maybeInstrumentSource (./nyc/index.js:166:15)
    at ./nyc/index.js:196:18
    at Module.replacementCompile (./nyc/node_modules/append-transform/index.js:52:13)
    at Module._extensions..js (module.js:406:10)
    at Object.<anonymous> (./nyc/node_modules/append-transform/index.js:57:4)
```

jamestalmage:
> This seems to break node test/fixtures/_generateReport.js:

I don't know what that file is supposed to do. Why not just use the API to generate a report?

bcoe:
@jamestalmage I've now tested with `xml2js`, `mkdirp`, `pageres`, `yarsay`, `yargs`, and `redis` -- everything seemed to work quite well \o/ and `--cache` definitely has quite a noticeable performance boost.

I managed to get an error once with `yarsay`, which might be worth digging into:

```bash
/Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/report/html.js:241
            text = structuredText[startLine].text;
                                            ^

TypeError: Cannot read property 'text' of undefined
    at /Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/report/html.js:241:45
    at Array.forEach (native)
    at annotateFunctions (/Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/report/html.js:224:26)
    at HtmlReport.Report.mix.writeDetailPage (/Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/report/html.js:427:9)
    at /Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/report/html.js:489:26
    at SyncFileWriter.extend.writeFile (/Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/util/file-writer.js:57:9)
    at FileWriter.extend.writeFile (/Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/util/file-writer.js:147:23)
    at /Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/report/html.js:488:24
    at Array.forEach (native)
    at HtmlReport.Report.mix.writeFiles (/Users/benjamincoe/bcoe/nyc/node_modules/istanbul/lib/report/html.js:482:23)
```

I couldn't, however, reproduce -- I wonder if it might be a race condition of some sort?

All and all things seem quite stable, and I think I'm ready to sign off as soon as you sort out the issue being discussed with @novemberborn, regarding `_generateReport.js` :+1:

jamestalmage:
> I wonder if it might be a race condition of some sort

That is entirely possible. After poking around the Node source, and the response I got [here](https://github.com/nodejs/node/issues/4352#issuecomment-166774031), I am concerned there is potential for reading bad data from a file that has not yet been fully written (I still am optimistic that we are not vulnerable to corrupt data, and that simply rerunning the test suite will fix the problem).

Did you try all those in your Windows VM? That would be where I am most concerned.

I am hoping to use `caching-transform` as the mechanism for caching Babel transforms in AVA. The extra exposure of the shared module can only help improve it.

novemberborn:
>> This seems to break node test/fixtures/_generateReport.js:

> I don't know what that file is supposed to do. Why not just use the API to generate a report?

It writes the `report.js` fixture, which includes coverage reports for [source map fixtures](https://www.npmjs.com/package/source-map-fixtures), *prior* to applying the respective source maps.

I just figured out what the problem is. `source-map-fixtures` is inside `node_modules` so it's always excluded. The script sets `nyc.exclude = []` to override. That's all fine (for its particular purposes) except that `istanbul` is now required lazily. Since it's required *after* `nyc.exclude` has been reset, the `istanbul` module itself needs to be instrumented, which creates a circular dependency. An empty object is returned instead.

Requiring `istanbul` before wrapping fixes the issue.

```diff
diff --git a/test/fixtures/_generateReport.js b/test/fixtures/_generateReport.js
index 58e55cc..945135d 100644
--- a/test/fixtures/_generateReport.js
+++ b/test/fixtures/_generateReport.js
@@ -20,6 +20,9 @@ var fixtures = {
   none: sourceMapFixtures.none('branching')
 }

+// Ensure istanbul is loaded before nyc wraps the require method.
+require('istanbul')
+
 // Inject nyc into this process.
 var nyc = (new NYC({
   cwd: path.join(__dirname, '..', '..')
```

bcoe:
@novemberborn @jamestalmage, I was talking to @isaacs about the race condition today -- `readFileSync` and `writeFileSync` are not atomic, so will cause problems when run in multiple processes. Having said this, npm uses this module:

https://www.npmjs.com/package/write-file-atomic

You should be able to use this to eliminate weirdness in caching-transform.

jamestalmage:
> readFileSync and writeFileSync are not atomic

That was what I suspected after reading the source. I have hammered caching-transform with a dozen processes at once, and not seen a failure, but good to get a definitive answer.

> write-file-atomic

Yep, we have discussed that before. I was waiting on a definitive answer before proceeding with it.

jamestalmage:
OK,
I have bumped `caching-transform` to a new release that uses `write-file-atomic`.

I also made the recommended change to `_generateReport`, this is what I see now:

```
$  node test/fixtures/_generateReport.js 
Written coverage report.
```

It exits with a `0` exit code, so I'm guessing that is correct. (I still have no idea what that file does).

jamestalmage:
Assuming CI passes, I think we are good to go.

novemberborn:
> I also made the recommended change to _generateReport

Thanks.

> It exits with a 0 exit code, so I'm guessing that is correct. (I still have no idea what that file does).

It writes the `report.js` fixture, which includes coverage reports for [source map fixtures](https://www.npmjs.com/package/source-map-fixtures), *prior* to applying the respective source maps. The `source-map-cache` tests use this report to test whether source maps are applied correctly.

jamestalmage:
Gotcha. I figured out that it wrote `report.js` bug couldn't figure out what it did.

isaacs:
@jamestalmage 

> That was what I suspected after reading the source. I have hammered caching-transform with a dozen processes at once, and not seen a failure, but good to get a definitive answer.

Chalk this one up to "aren't file systems fun?"

If you have a dozen *processes* but it's always reading/writing to the same file, then the OS's heuristics will see that the same file is being read and written repeatedly, and keep it in the file system's cache.  (Ie, it won't even get written to disk.)  This means that the operations take a lot less time, and race conditions magically disappear!  Lovely!

Except, when you have dozens or hundreds of *files*, especially if they're pretty big, and even a small number of processes, what happens is that the caching mechanisms can't hold all the data, and the race conditions get more relevant again.

So, "works on my machine", "tests all pass", but repeatedly breaks in production.  File systems are liars.  Well-intended liars that usually make programs faster and more reliable, but... still liars.

Not to mention, often stuff will work fine on a Linux machine, but not on Windows.  Or work on a Linux VM on a windows machine, but not on an actual linux machine.  Or it'll work on OS X, but not if you use a specific FS option when you set up your machine, or with SSDs but not with spinning platter drives.

And here I thought server side JS would mean getting away from all that browser incompatibility bs. ;P

bcoe:
Great work on this @jamestalmage; I'll do a final pass of tests on the usual suspects, and some testing on Windows -- I'm pretty excited about this branch.

bcoe:
@jamestalmage https://twitter.com/jfuerth/status/587249053630648320

bcoe:
@jamestalmage did another smoke test with the current branch, I tested the following libraries with `--cache` enabled and disabled:

`yargs`, `yarsay`, `mkdirp`, `uuid`, `moment`, `coffeescript`, I found one weird edge-case in the process:

1. link this branch of `nyc`.
2. run the `yarsay` test-suite with `--cache` added.
3. run `nyc report --reporter=lcov`.

I get the following error:

screen shot 2015-12-24 at 10 45 07 pm

Everything else worked like a charm.

jamestalmage:
@bcoe fixed in 4d4093b

bcoe:
@jamestalmage unfortunately, I'm still getting this behavior:

screen shot 2015-12-25 at 10 23 46 am

I've changed the testing line to the following:

```sh
nyc --reporter=lcov --cache --require babel-core/register mocha
```

So it doesn't appear to be an issue with running `report` independently of running the actual testing suite.

Again everything works if I don't pass cache, seems like it must be something small.

jamestalmage:
Try deleting the cache. This is yargs right? It all worked for me.

bcoe:
@jamestalmage can't seem to get it to run, I'm testing using [yarsay](https://github.com/bcoe/yarsay) (which is my one library with a babel require hook, so I find it's good for smoke tests).

It would probably be smart for us to create a script that tests against several libraries, rather than my painful manual step.

jamestalmage:
I used yarsay to debug the problem, and made the same script modifications you did. I'm away from my computer most of the day - I can verify later.

jamestalmage:
@bcoe 
I found my problem. I was testing almost identically to you, but I had added `rm -rf node_modules/.cache` to yarsay's `pretest` script. This mean I always had a cache miss. So I never saw the problem I just fixed in c3b47fc.

bcoe:
@jamestalmage awesome, I will re-test today.

bcoe:
@jamestalmage I've confirmed that everything is working with my set of smoke-testing libraries :+1: rebase and merge at will?

jamestalmage:
I am on mobile ATM, but it appears mergeable now.

bcoe:
@novemberborn @jamestalmage give the `next` tag a spin, it has the cache functionality (also we're no longer bundling spawn-wrap).

```sh
npm i nyc@next
```

bcoe:
@novemberborn @jamestalmage @sindresorhus `nyc@5.1.1` is out with caching.

ronkorving:
Awesome job guys!

Just some numbers from [a very small project](https://www.npmjs.com/package/express-rested) to confirm the improvements:

(I took the most average result of 4 runs)

Test only:

```
$ time ./node_modules/.bin/tape 'test/*.js' > /dev/null 2>&1

real	0m0.590s
user	0m0.519s
sys	0m0.067s
```

nyc 4.0.0:

```
$ time ./node_modules/.bin/nyc --silent --all ./node_modules/.bin/tape 'test/*.js' > /dev/null 2>&1

real	0m3.838s
user	0m3.102s
sys	0m0.852s
```

nyc 5.3.0:

```
$ time ./node_modules/.bin/nyc --silent --all ./node_modules/.bin/tape 'test/*.js' > /dev/null 2>&1

real	0m1.821s
user	0m1.754s
sys	0m0.183s
```

PR #118, opened 2015-12-27
switch to canonical version of spawn-wrap

bcoe:
thanks to @isaacs' hard work, and lots of debugging on various platforms, we've now not a version of spawn-wrap published that works on OSX, Windows, and Linux \o/


bcoe:
screen shot 2015-12-27 at 12 12 33 pm

@rmg @isaacs finally \o/

PR #121, opened 2015-12-29
Add reset, clean up before every test

### LINKED ISSUE #112: Clean state before every test.
There is at least one test that still requires existing state from a previous test.

I propose we wipe out `.nyc_output` and `test/fixtures/.nyc_output` between tests, as well as the cache directories.

Any tests that fail due to these changes should be updated to explicitly generate their prerequisite state.

### LINKED ISSUE #120: `--all` appears to be broken in 5.1.0
In `nyc@5.1.0`, I am getting the following error whenever `--all` is used:

```bash
user$nyc --all ava 'test/**/*.js'
fs.js:549
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open '/Users/user/project/.nyc_output/7910.json'
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.writeFileSync (fs.js:1156:15)
    at NYC.writeCoverageFile (/Users/user/project/node_modules/nyc/index.js:259:6)
    at NYC.addAllFiles (/Users/user/project/node_modules/nyc/index.js:171:8)
    at Object.<anonymous> (/Users/user/project/node_modules/nyc/bin/nyc.js:122:23)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
```

Not sure where to start looking on this issue.

jamestalmage:
This adds a `reset` method which deletes the temp directory then adds it back. `createTempDirectory` is no longer called for every forked process.

Instead of writing a test for  #120, I just went ahead and fixed #112. I have verified it would have caught the error.

Fixes #120 
Fixes #112

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -117,7 +117,7 @@ if (process.env.NYC_CWD) {
   } else if (argv._.length) {
     // wrap subprocesses and execute argv[1]
     var nyc = (new NYC())
-    nyc.cleanup()
+    nyc.reset()
 
     if (argv.all) nyc.addAllFiles()
     if (!Array.isArray(argv.require)) argv.require = [argv.require]
--- a/index.js
+++ b/index.js
@@ -49,8 +49,6 @@ function NYC (opts) {
   // require extensions can be provided as config in package.json.
   this.require = arrify(config.require || opts.require)
 
-  this._createDatastoreDirectories()
-
   this.transform = this._createTransform()
 
   this.sourceMapCache = new SourceMapCache()
@@ -220,8 +218,13 @@ NYC.prototype.clearCache = function () {
   rimraf.sync(this.cacheDirectory())
 }
 
-NYC.prototype._createDatastoreDirectories = function () {
+NYC.prototype.createTempDirectory = function () {
   mkdirp.sync(this.tempDirectory())
+}
+
+NYC.prototype.reset = function () {
+  this.cleanup()
+  this.createTempDirectory()
 }
 
 NYC.prototype._wrapExit = function () {
--- a/test/src/nyc-test.js
+++ b/test/src/nyc-test.js
@@ -159,6 +159,7 @@ describe('nyc', function () {
       var nyc = new NYC({
         cwd: process.cwd()
       })
+      nyc.reset()
       nyc.wrap()
 
       var check = require('../fixtures/check-instrumented')
--- a/test/src/nyc-test.js
+++ b/test/src/nyc-test.js
@@ -27,23 +27,29 @@ var sinon = require('sinon')
 var isWindows = require('is-windows')()
 var spawn = require('win-spawn')
 var fixtures = path.resolve(__dirname, '../fixtures')
+var projectDir = path.resolve(__dirname, '../..')
+var projectTempDir = path.join(projectDir, '.nyc_output')
+var projectCacheDir = path.join(projectDir, 'node_modules', '.cache', 'nyc')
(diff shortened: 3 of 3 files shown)


bin/nyc.js
```diff
@@ -117,7 +117,7 @@ if (process.env.NYC_CWD) {
   } else if (argv._.length) {
     // wrap subprocesses and execute argv[1]
     var nyc = (new NYC())
-    nyc.cleanup()
+    nyc.reset()
```
novemberborn: I wonder if the `if (!process.env.NYC_CWD)` clause in `cleanup()` is causing some confusion here. I think it ensures child processes don't perform any cleanup, but maybe that guard should be in this file and not in the implementation.

TBH I'm not entirely sure what the expected behavior is and why it impacts `--all`.

jamestalmage: The issue with `all` was that it was calling `cleanup()`, which deleted directories, but did not recreate them.

> but maybe that guard should be in this file and not in the implementation.

I think I agree with you. I just wanted to avoid an additional change creeping into this PR.

bcoe:
@jamestalmage how do you feel about moving the cache folder into `.nyc_output`? I think it would be nice if we only created one top-level folder with nyc, rather than a top-level folder and a folder in node_modules -- I probably should have brought this up in the earlier code review :)

jamestalmage:
Separate folders makes it easier to clean between tests (we don't want to clean the cache each time).

Also, AVA has already followed suit and caches Babel transforms in `node_modules/.cache/ava`.  

If anything, I would like to move `.nyc_output ` into `node_modules/.temp/nyc` or similar. Then people would not need to modify their `.gitignore` for `nyc`.

bcoe:
@jamestalmage this seems like a reasonable argument, it's good that AVA and nyc are standardizing on an approach.

Let's hold off on moving `.nyc_output` at this point, I'd like to get a stable `5.1.1` out the door and tested for a bit, before we make many more changes.

bcoe:
I'm happy with this branch, once you and @novemberborn both sign off.

bin/nyc.js
novemberborn: > The issue with all was that it was calling cleanup(), which deleted directories, but did not recreate them.

Where did it call `cleanup()`? That's the part I'm not understanding, I suppose.



> I just wanted to avoid an additional change creeping into this PR.

Sure.

novemberborn:
I don't quite understand why `--all` is broken and how this fixes it, but no objections otherwise.

bcoe:
@Lalem001, can you confirm that this branch solves your `--all` problem?

bin/nyc.js
jamestalmage: > Where did it call cleanup()? 

Right there in that diff. Now instead it calls `reset`, which deletes and recreates the directory.

The issue is that `all` immediately and synchronously writes a "preample" for every included file to disk, before continuing with the test, so the folder needs to exist, even though we haven't yet forked a process.

novemberborn: Ah! OK, resetting before running anything makes sense.

Lalem001:
The issue is resolved. :+1:

bcoe:
:shipit:

bcoe:
@Lalem001 `nyc@5.1.1` is now out with caching, and fixes for `--all`.

PR #122, opened 2016-01-01
Use find-cache-dir to determine cache directory.

jamestalmage:
Uses `find-cache-dir` for consistent cache locating with other modules (currently that list only includes AVA, but one can hope). It takes the extra step of ensuring a `package.json` file exists (and hunting up the directory tree for it), otherwise it returns null. If it can't find a suitable cache directory, caching is disabled.

I also dropped the `cacheDirectory()` method in favor of just setting a property.

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,7 @@ var arrify = require('arrify')
 var SourceMapCache = require('./lib/source-map-cache')
 var convertSourceMap = require('convert-source-map')
 var md5hex = require('md5-hex')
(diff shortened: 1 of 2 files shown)


novemberborn:
Nice!

bcoe:
@jamestalmage this seems like a smart change, land when you're ready -- perhaps we can get a new nyc out the door today once I test @isaacs's branch a bit more on Windows.

PR #127, opened 2016-01-03
fix caching ... again.

jamestalmage:


index.js
```diff
@@ -44,12 +44,9 @@ function NYC (opts) {
...
   )
 
-  this.enableCache = opts.enableCache === true || process.env.NYC_CACHE === 'enable'
+  this.cacheDirectory = findCacheDir({name: 'nyc', cwd: this.cwd})
 
-  if (this.enableCache) {
-    this.cacheDirectory = findCacheDir({name: 'nyc', cwd: this.cwd})
-    this.enableCache = Boolean(this.cacheDirectory)
-  }
+  this.enableCache = Boolean(this.cacheDirectory && (opts.enableCache === true || process.env.NYC_CACHE === 'enable'))
```
bcoe: Since the only time we set `NYC_CACHE` is to pass state to `spawn-wrap`, I don't think we need to have this check for `process.env.NYC_CACHE` here -- I'd rather keep `index.js` agnostic of environment variables if possible.

jamestalmage: Sure. Just following suite with `NYC_CWD`.

bcoe:
@jamestalmage fix confirmed, once tests pass let's :shipit:

jamestalmage:
Landed

PR #132, opened 2016-01-04
Use `nyc` instead of `config.nyc` in package.json

### LINKED ISSUE #131: `package.json`: un-nest `config.nyc` to just `nyc`
According to @sindresorhus, that is [the more common convention](https://github.com/sindresorhus/ava/issues/405#issuecomment-168569929).

Obviously we would need to support the current usage as well for the foreseeable future.

jamestalmage:
Fixes #131

This also uses `pkg-up` to locate the nearest `package.json` from `cwd`, and will move `cwd` to there. This allows you to invoke `nyc` from a subdirectory of your project and still have it behave as expected.  This is something our caching mechanism already does. If it can't find a `package.json`, it just uses the original `cwd` option.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,6 +15,8 @@ var SourceMapCache = require('./lib/source-map-cache')
 var convertSourceMap = require('convert-source-map')
 var md5hex = require('md5-hex')
 var findCacheDir = require('find-cache-dir')
+var pkgUp = require('pkg-up')
+var readPkg = require('read-pkg')
 
 /* istanbul ignore next */
 if (/index\.covered\.js$/.test(__filename)) {
@@ -27,12 +29,11 @@ function NYC (opts) {
   this._istanbul = opts.istanbul
   this.subprocessBin = opts.subprocessBin || path.resolve(__dirname, './bin/nyc.js')
   this._tempDirectory = opts.tempDirectory || './.nyc_output'
-  this.cwd = opts.cwd || process.env.NYC_CWD || process.cwd()
-  this.reporter = arrify(opts.reporter || 'text')
 
-  // you can specify config in the nyc stanza of package.json.
-  var config = require(path.resolve(this.cwd, './package.json')).config || {}
-  config = config.nyc || {}
+  var config = this._loadConfig(opts)
+  this.cwd = config.cwd
+
+  this.reporter = arrify(opts.reporter || 'text')
 
   // load exclude stanza from config.
   this.include = false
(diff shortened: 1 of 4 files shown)


novemberborn:
LGTM

index.js
```diff
@@ -27,12 +29,22 @@ function NYC (opts) {
...
-  this.reporter = arrify(opts.reporter || 'text')
+
+  var cwd = opts.cwd || process.env.NYC_CWD || process.cwd()
+  var pkgPath = pkgUp.sync(cwd)
 
   // you can specify config in the nyc stanza of package.json.
-  var config = require(path.resolve(this.cwd, './package.json')).config || {}
-  config = config.nyc || {}
+  var config
+  if (pkgPath) {
```
bcoe: the constructor is getting pretty verbose at this point, mind moving the config logic into a private method, e.g., `_loadConfig`.

index.js
```diff
@@ -27,12 +29,22 @@ function NYC (opts) {
...
+    cwd = path.dirname(pkgPath)
+    var pkg = readPkg.sync(pkgPath, {normalize: false})
+    config = pkg.nyc || (pkg.config && pkg.config.nyc)
+  }
+  config = config || {}
+
+  this.cwd = cwd
```
bcoe: It seems reasonable to walk the directory structure to grab the nearest config, but I don't think the config location should have an effect on `cwd`.

The idea of passing around `process.NYC_CWD` is that regardless of the behavior of subprocesses, the `cwd` stays the same as when the `cwd` command is initially executed.

bcoe:
@jamestalmage looks good, just a couple comments.

index.js
jamestalmage: Each process will do exactly the same thing, so if `cwd` does get shifted, it gets shifted the same across processes (unless a new `package.json` magically appears in the directory tree during the test - which seems unlikely).

With this, I could do:

```sh
$ cd test/foo/bar
# then the following repeatedly until the test works
$ nyc tap ./baz.js
```

instead of 

```js
$ nyc tap test/foo/bar/baz.js
```

every time.

Also, if I did the first thing (cd into a subdir), I would rather `.nyc_output` continue to be mounted in the base of my project, not whatever directory I happen to be in.

If there is no `package.json` in the hierarchy, it falls back to the old method (it puts `.nyc_output` wherever you happen to be at the time).

bcoe: what if we have a library under coverage that has a few inner `package.json` files? perhaps in a fixtures directory, or with some projects like [babel](https://github.com/babel/babel/tree/master/packages) I've seen a mono-repo approach that has many package.json files, is there a danger that we'll accidentally source the inner package.json in a subprocess rather than the root directory's?

jamestalmage: Only if the NYC_CWD environment variable got manipulated (or deleted and process.cwd() was manipulated). But that would break the old logic as well.

bcoe: cool, fears alleviated :+1:

bcoe: @jamestalmage once you've refactored the config loading logic into a helper, I say :shipit:

index.js
```diff
@@ -59,6 +60,24 @@ function NYC (opts) {
   this.loadedMaps = null
 }
 
+NYC.prototype._loadConfig = function (opts) {
+  var cwd = opts.cwd || process.env.NYC_CWD || process.cwd()
```
bcoe: rock on \o/

PR #139, opened 2016-01-15
updated documentation based on the discussion in #134

### LINKED ISSUE #134: Add better examples to the documentation
Update from moderator:

The biggest issue is that our documentation shows global use of `nyc` without ever suggesting that `nyc` be installed globally.

Original Message:
---
I am not quite sure how the instructions in the README ever worked. `npm i nyc` installs the module just locally and nothing gets added to the path. Calling

    nyc tap ./test/*.js

gives a "command not found" (to no surprise). One would think that

    ./node_modules/nyc/bin/nyc.js tap ./test/*.js

should work. But instead that gives

        $ ./node_modules/nyc/bin/nyc.js tap ./test/*.js
        events.js:141
              throw er; // Unhandled 'error' event
              ^

        Error: spawn tap ENOENT
            at exports._errnoException (util.js:874:11)
            at Process.ChildProcess._handle.onexit (internal/child_process.js:178:32)
            at onErrorNT (internal/child_process.js:344:16)
            at doNTCallback2 (node.js:439:9)
            at process._tickCallback (node.js:353:17)
            at Function.Module.runMain (module.js:469:11)
            at startup (node.js:134:18)
            at node.js:961:3

bcoe:
I've updated documentation based on the discussion in #134. @jamestalmage, @tcurdt, could you give this a read?

fixes #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,30 +5,17 @@
 [![NPM version](https://img.shields.io/npm/v/nyc.svg)](https://www.npmjs.com/package/nyc)
 [![Windows Tests](https://img.shields.io/appveyor/ci/bcoe/nyc/master.svg?label=Windows%20Tests)](https://ci.appveyor.com/project/bcoe/nyc)
 
-
-```shell
-nyc npm test
-```
-
 a code coverage tool built on [istanbul](https://www.npmjs.com/package/istanbul)
 that works for applications that spawn subprocesses.
 
 ## Instrumenting Your Code
 
-Simply run your tests with `nyc`, and it will collect coverage information for
-each process and store it in `.nyc_output`.
+You can install nyc as a development dependency and add it to the test stanza
+in your package.json.
 
 ```shell
-nyc npm test
+npm i nyc --save-dev
 ```
-
-you can pass a list of Istanbul reporters that you'd like to run:
-
-```shell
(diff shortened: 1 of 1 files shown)


tcurdt:
Does help - but still a bit vague on whether referring to the binary needs a `./node_modules/.bin` prefix or not.

bcoe:
@tcurdt in the `test` stanza of your `package.json`, you should never need to provide the absolute path `./node_modules/.bin` -- so as long as you follow the documentation in the README.md, life should be good.

tcurdt:
@bcoe odd - it does indeed work without the prefix. Could have sworn that was giving me trouble. Well, nevermind then.

bcoe:
@jamestalmage can we ship this?

jamestalmage:
LGTM

Is there a good resource to link to on using `node_modules/.bin/` for local installs?

@sindresorhus If there's not something out there you should write one and add it to https://github.com/sindresorhus/guides. This issue comes up all the time, and it would be nice to have a concise clearly written resource to point people at.

@bcoe Ship it. If we find a good link, we'll just add it when we do.

sindresorhus:
@jamestalmage I cover it briefly [here](https://github.com/sindresorhus/awesome-npm#run-scripts), but I'll look into doing something more comprehensive.

PR #145, opened 2016-01-19
allow the directory that coverage reports are outputted to to be configured

### LINKED ISSUE #63: Specify output folder
Is there any way to control where it spits out the generated report without involving `.istanbul.yml`?

bcoe:
fixes #63 

reviewers: @jbcpollak, @mateatslc, @jamestalmage

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -21,6 +21,10 @@ var yargs = require('yargs')
         describe: 'coverage reporter(s) to use',
         default: 'text'
       })
+      .option('report-dir', {
+        describe: 'default directory to output coverage reports in',
+        default: 'coverage'
+      })
       .help('h')
       .alias('h', 'help')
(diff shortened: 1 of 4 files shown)


jamestalmage:
The code looks fine. I'm still not sold as to why it's necessary. Do people have real reasons they need to rename the directory? I haven't heard anybody say "Without this option it's impossible/difficult to use NYC, because...".

This sort of option isn't bad in itself, but allowing them to accumulate unchecked just bloats the codebase, creates more potential failure points, and just slows development of meaningful features. If there is legitimate need, then by all means :+1:.

jbcpollak:
The reason I needed to relocate the output was to conform to the
requirements of our CI service (circleci) which needs reports in a specific
directory.

jamestalmage:
Sold. :shipit:

mateatslc:
My reason for this is similar to @jbcpollak's, help CI process find the report. This does exactly what I needed, thanks!

PR #150, opened 2016-01-22
guard against illegal positions

novemberborn:
Istanbul can sometimes return illegal position values. Clamp them so they can be used by the source map consumer.

See https://github.com/bcoe/nyc/pull/143#issuecomment-173554681 for background.


bcoe:
@novemberborn nice elegant patch :+1: 

@albertogasparin would you be able to give this branch a try, and see if it solves your problems?

PR #160, opened 2016-02-01
Don't instrument files outside of the current working directory

### LINKED ISSUE #154: Covers `npm link`ed modules.
If I use `npm link` and then run a test suite using NYC, the linked module ends up generating coverage info. I'm guessing because there is no `node_modules` folder in the parent path for linked modules (at least when using `nvm`).

### LINKED ISSUE #157: Empty coverage output
I created a fake module https://gist.github.com/micnic/f815c0bee18e99805318 and added a test for it, I am using `tap` which uses `nyc` for code coverage. As you can see, If I do not add any additional argument I get an empty output, in case I add `--all` argument the output adds the `index.js` file but it has 0% coverage which is false as the exported function was executed inside the test file.

Does `nyc` avoid somehow my `index.js` file or it is required in the way `nyc` does not expect the file to be required or I am missing something?

novemberborn:
Files outside of the current working directory (meaning higher up in the file hierarchy) should not be instrumented.

Fixes #157 where nyc is run from inside a `node_modules` directory.

Fixes #154 where 'npm link' is used to fulfil dependencies.

This removes matching the include/exclude patterns to the absolute filename, the patterns are now applied only to the relative portion within the current working directory.

`addAllFiles()` has been changed to explicitly add files in the current working directory only.

Some other tests related to include/exclude behavior have been cleaned up.

---

Note that I'm not entirely certain `filename` in `shouldInstrumentFile()` is always absolute. Review on that front would be especially appreciated.

This changes the include/exclude matching behavior, but restricting it to the cwd makes more sense to me.

### DIFF
--- a/index.js
+++ b/index.js
@@ -183,7 +183,7 @@ NYC.prototype.shouldInstrumentFile = function (filename, relFile) {
   // Don't instrument files that are outside of the current working directory.
   if (/^\.\./.test(path.relative(this.cwd, filename))) return false
 
-  relFile = relFile.replace(/^\.\//, '') // remove leading './'.
+  relFile = relFile.replace(/^\.[\\\/]/, '') // remove leading './' or '.\'.
   return (!this.include || micromatch.any(relFile, this.include)) && !micromatch.any(relFile, this.exclude)
 }
 
--- a/index.js
+++ b/index.js
@@ -180,10 +180,11 @@ NYC.prototype._readTranspiledSource = function (path) {
 }
 
 NYC.prototype.shouldInstrumentFile = function (filename, relFile) {
-  relFile = relFile.replace(/^\.\//, '') // remove leading './'.
+  // Don't instrument files that are outside of the current working directory.
+  if (/^\.\./.test(path.relative(this.cwd, filename))) return false
 
-  return (!this.include || micromatch.any(filename, this.include) || micromatch.any(relFile, this.include)) &&
-    !(micromatch.any(filename, this.exclude) || micromatch.any(relFile, this.exclude))
+  relFile = relFile.replace(/^\.\//, '') // remove leading './'.
+  return (!this.include || micromatch.any(relFile, this.include)) && !micromatch.any(relFile, this.exclude)
 }
 
 NYC.prototype.addAllFiles = function () {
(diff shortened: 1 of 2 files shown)


bcoe:
@novemberborn  great work; I'll do a thorough test on a few platforms/projects and we can release this to a `next` tag ASAP.

novemberborn:
> I'll do a thorough test on a few platforms/projects and we can release this to a next tag ASAP.

Great!

bcoe:
@novemberborn starting doing a little bit of testing, this patch doesn't seem to be working for a linked module:

screen shot 2016-02-02 at 10 58 56 pm

One other thing we might want to be mindful of is this logic:

```js
relFile = relFile.replace(/^\.\//, '') // remove leading './'.
```

We should probably be using `path.sep` for the benefit of Windows.

bcoe:
@novemberborn the good news, is I saw that this patch solves the issues @micnic was having \o/

jamestalmage:
> We should probably be using `path.sep`

Or just account for either slash in the regexp

```js
str.replace(/\.[\\\/]/, '')
```

Using `path.sep` in a regexp is painfull

sindresorhus:
> Or just account for either slash in the regexp

:+1:

novemberborn:
@bcoe 

> starting doing a little bit of testing, this patch doesn't seem to be working for a linked module

Ha, this is a nice one! I reproduced this by cloning <https://github.com/bcoe/yargs/tree/split-out-parser> and <https://github.com/yargs/yargs-parser/tree/require.main.filename> into sibling directories, then `npm link`ing `yargs-parser` into `yargs`.

Turns out when you have paths like these:

```
/var/folders/_6/p8qxp_3n62zg9081tvb0lcc80000gn/T/tmp.NOCrQG1U/yargs
/var/folders/_6/p8qxp_3n62zg9081tvb0lcc80000gn/T/tmp.NOCrQG1U/yargs-parser
```

`'yars-parser'.indexOf('yars')` does return `0`, so `yargs-parser` was included. I've (force-)pushed a fix which checks if `path.relative(this.cwd, filename)` returns a string that starts with `..`. I was being lazy with `indexOf`.

> We should probably be using path.sep for the benefit of Windows.

Pushed a commit which fixes that, using @jamestalmage's method.

bcoe:
lol @novemberborn that's great, as usual I manage to be a nightmare user of our own library and find a crazy edge-case:

golden

PR #163, opened 2016-02-08
Add support for configurable file extensions

lloydcotten:
This adds support for allowing files with extensions other than '.js' to be covered.  It can be configured through package.json config, or command line args.  

As per my comment in issue https://github.com/bcoe/nyc/issues/161#issuecomment-180747165, there aren't any additional tests yet. If you feel some additional tests to cover this functionality would be helpful, I'd appreciate some guidance as to how you think they could be done (I'm not quite used to the way the existing tests are done).  Even just pointing me to existing tests that I could pattern after, would be great.

I'd be glad to address any other comments also.

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,6 +17,7 @@ var md5hex = require('md5-hex')
 var findCacheDir = require('find-cache-dir')
 var pkgUp = require('pkg-up')
 var readPkg = require('read-pkg')
+var js = require('default-require-extensions/js')
 
 /* istanbul ignore next */
 if (/index\.covered\.js$/.test(__filename)) {
@@ -53,7 +54,14 @@ function NYC (opts) {
   // require extensions can be provided as config in package.json.
   this.require = arrify(config.require || opts.require)
 
-  this.transform = this._createTransform()
+  this.extensions = arrify(config.extension || opts.extension).concat('.js').map(function (ext) {
+    return ext.toLowerCase()
+  })
+
+  this.transforms = this.extensions.reduce(function (transforms, ext) {
+    transforms[ext] = this._createTransform(ext)
+    return transforms
+  }.bind(this), {})
 
   this.sourceMapCache = new SourceMapCache()
 
@@ -79,7 +87,7 @@ NYC.prototype._loadConfig = function (opts) {
   return config
 }
 
-NYC.prototype._createTransform = function () {
+NYC.prototype._createTransform = function (ext) {
   var _this = this
   return cachingTransform({
     salt: JSON.stringify({
@@ -94,7 +102,7 @@ NYC.prototype._createTransform = function () {
     factory: this._transformFactory.bind(this),
     cacheDir: this.cacheDirectory,
     disableCache: !this.enableCache,
-    ext: '.js'
+    ext: ext
   })
 }
 
@@ -212,7 +220,15 @@ NYC.prototype._maybeInstrumentSource = function (code, filename, relFile) {
     return null
   }
 
-  return this.transform(code, {filename: filename, relFile: relFile})
+  var ext, transform
+  for (ext in this.transforms) {
+    if (filename.toLowerCase().substr(-ext.length) === ext) {
+      transform = this.transforms[ext]
+      break
+    }
+  }
+
+  return transform ? transform(code, {filename: filename, relFile: relFile}) : null
 }
 
 NYC.prototype._transformFactory = function (cacheDir) {
@@ -238,11 +254,17 @@ NYC.prototype._transformFactory = function (cacheDir) {
   }
 }
 
+NYC.prototype._handleJs = function (code, filename) {
+  var relFile = path.relative(this.cwd, filename)
+  return this._maybeInstrumentSource(code, filename, relFile) || code
+}
+
 NYC.prototype._wrapRequire = function () {
-  var _this = this
-  appendTransform(function (code, filename) {
-    var relFile = path.relative(_this.cwd, filename)
-    return _this._maybeInstrumentSource(code, filename, relFile) || code
+  var handleJs = this._handleJs.bind(this)
+
+  this.extensions.forEach(function (ext) {
+    require.extensions[ext] = js
+    appendTransform(handleJs, ext)
   })
 }
 
--- a/test/fixtures/package.json
+++ b/test/fixtures/package.json
@@ -12,7 +12,8 @@
       "exclude": [
         "**/blarg",
         "**/blerg"
-      ]
+      ],
+      "extension": [".es6", ".foo.BAR"]
     }
   }
 }
--- a/test/src/nyc-test.js
+++ b/test/src/nyc-test.js
@@ -83,6 +83,22 @@ describe('nyc', function () {
 
       nyc.exclude.length.should.eql(19)
     })
+
+    it("loads 'extension' patterns from package.json#config.nyc", function () {
+      var nyc = new NYC({
+        cwd: path.resolve(__dirname, '../fixtures')
+      })
+
+      nyc.extensions.length.should.eql(3)
+    })
+
+    it("loads 'extension' from package.json#nyc", function () {
+      var nyc = new NYC({
+        cwd: path.resolve(__dirname, '../..')
+      })
+
+      nyc.extensions.length.should.eql(2)
+    })
   })
 
   describe('_prepGlobPatterns', function () {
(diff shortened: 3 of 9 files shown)


index.js
```diff
@@ -240,9 +252,17 @@ NYC.prototype._transformFactory = function (cacheDir) {
...
+
+  function handleJs (code, filename) {
     var relFile = path.relative(_this.cwd, filename)
     return _this._maybeInstrumentSource(code, filename, relFile) || code
+  }
+
+  appendTransform(handleJs)
+
+  this.extensions.forEach(function (ext) {
+    require.extensions[ext] = require.extensions['.js']
```
jamestalmage: Maybe we should use [`default-require-extensions`](https://www.npmjs.com/package/default-require-extensions) here? Just in case `require.extensions['.js']` has been messed with already by some other module?

novemberborn:
The new extensions are pointed to `.js`. This assumes files with those new extensions can be treated as `.js`, right? E.g. it wouldn't add support for CoffeeScript or TypeScript.

index.js
```diff
@@ -212,7 +217,14 @@ NYC.prototype._maybeInstrumentSource = function (code, filename, relFile) {
     return null
   }
 
-  return this.transform(code, {filename: filename, relFile: relFile})
+  var ext = path.extname(filename)
```
novemberborn: Note that `extname` only returns the *last* extension. Looping through the registered transforms to see if they match the filename would let us support `.foo.bar`.

bcoe:
@lloydcotten would it be possible to add a test for this functionality? I differ to @jamestalmage and @novemberborn's other comments -- they've become the subject matter experts on the codebases that facilitate the extension overrides.

index.js
lloydcotten: I've updated to use this.

lloydcotten:
@novemberborn This would not support CoffeeScript or TypeScript.  We would have to somehow allow custom hooks to be added.  Wouldn't that work the same way as it already does for Babel anyway?

index.js
lloydcotten: I've updated to use the Array `filter` function to find matching extensions.

lloydcotten:
@novemberborn I took some time today to try to figure out the best way to add some tests around this.  I've taken a stab at it.  Not sure if it's accomplished the best way possible, but the functionality is covered and passes.

Any feedback is welcome.  I'll be glad to try to refactor anything.  Once every one is OK with it, I'll squash the commits.  I'm removing the [WIP] tag.

lloydcotten:
- [ ] Squash commits

index.js
```diff
@@ -53,7 +54,12 @@ function NYC (opts) {
   // require extensions can be provided as config in package.json.
   this.require = arrify(config.require || opts.require)
 
-  this.transform = this._createTransform()
+  this.extensions = arrify(config.extensions || opts.extension)
```
novemberborn: This should access `config.extension` I think. Like `--require` results in `opts.require` and `config.require`, so should `--extension`.

index.js
```diff
@@ -53,7 +54,12 @@ function NYC (opts) {
...
+
+  this.transform = this.extensions.concat(['.js']).reduce(function (transform, ext) {
```
novemberborn: `this.transforms` might be better, since it may now hold multiple transforms.

index.js
```diff
@@ -53,7 +54,12 @@ function NYC (opts) {
...
+    transform[ext] = this._createTransform(ext)
```
novemberborn: Perhaps lowercase `ext` here to make the later comparison easier.

I'm assuming `require.extensions` is case insensitive. If it isn't, well that would be weird.

index.js
novemberborn: Perhaps add `.js` onto `this.extensions`. That way you can use it in `_maybeInstrumentSource()`. Could also map it to ensure all extensions are lowercased.

index.js
```diff
@@ -212,7 +218,12 @@ NYC.prototype._maybeInstrumentSource = function (code, filename, relFile) {
     return null
   }
 
-  return this.transform(code, {filename: filename, relFile: relFile})
+  var ext = Object.keys(this.transform).filter(function (ext) {
+    return filename.toUpperCase().substr(-ext.length) === ext.toUpperCase()
+  })
```
novemberborn: This function gets called *a lot*. Using `this.extensions` should help. Perhaps you could use a `for` loop and break once you've matched an extension, rather than filtering a full array?

test/src/nyc-test.js
```diff
@@ -83,6 +83,22 @@ describe('nyc', function () {
...
+
+    it("loads 'extensions' patterns from package.json#config.nyc", function () {
+      var nyc = new NYC({
+        cwd: path.resolve(__dirname, '../fixtures')
+      })
+
+      nyc.extensions.length.should.eql(2)
+    })
+
+    it("loads 'extensions' from package.json#nyc", function () {
```
novemberborn: @bcoe I'm thinking this (and similar tests) should really be testing another fixture directory, rather than the top level `package.json`. Kinda weird to allow `.es6` for the nyc project itself.

test/src/nyc-test.js
```diff
@@ -232,6 +248,39 @@ describe('nyc', function () {
       })
     })
 
+    describe('compile handlers for custom extensions are assigned', function () {
+      it('assigns a function to custom extensions', function () {
+        var nyc = new NYC({
+          cwd: process.cwd()
```
novemberborn: Better to provide `extension: '.es6'` here than rely on any `package.json` config. More explicit that way.

novemberborn:
@lloydcotten 

> This would not support CoffeeScript or TypeScript. We would have to somehow allow custom hooks to be added. Wouldn't that work the same way as it already does for Babel anyway?

I'm not sure really. Not the end of the world if that doesn't work right away.

> I took some time today to try to figure out the best way to add some tests around this. I've taken a stab at it. Not sure if it's accomplished the best way possible, but the functionality is covered and passes.

Tests look pretty good! Should test with multiple extensions since there is looping logic. The case insensitive matching needs covering as well.

test/src/nyc-test.js
lloydcotten: The reason I didn't is because the tests will use package.json by default over the `opts` argument.  I patterned this after the `config.require` / `opts.require` that already exists.

I agree that it is not intuitive (I would expect `opts` to override `config`), but I figured the most important thing is consistency.  If you want, I can swap this around so that `opts` does override `config` for the `extension`.  I could submit a separate PR for the `require` config so that it behaves the same way.  Let me know your thoughts.

lloydcotten: Reference for my previous comment: https://github.com/bcoe/nyc/blob/master/index.js#L54

lloydcotten:
@novemberborn I updated the PR with some refactoring based on your comments.  When I hear your feedback on my question about the tests, I'll modify the tests as necessary and add one to cover case insensitive matching.

test/src/nyc-test.js
novemberborn: > The reason I didn't is because the tests will use package.json by default over the opts argument. I patterned this after the config.require / opts.require that already exists.
> 
> I agree that it is not intuitive (I would expect opts to override config), but I figured the most important thing is consistency.

Oh I hadn't even realized it worked that way. Could you open an issue so @bcoe can weigh in? I suppose technically that'd be a breaking change.

---

That aside my larger concern is using the main `package.json` for these tests. It was kind of convenient for the `exclude` option but it seems strange to enable the `.es6` extension within the nyc codebase.

I suppose it's OK for now, could refactor into using a separate fixture `package.json` later.

index.js
```diff
@@ -53,7 +54,14 @@ function NYC (opts) {
   // require extensions can be provided as config in package.json.
   this.require = arrify(config.require || opts.require)
 
-  this.transform = this._createTransform()
+  this.extensions = arrify(config.extension || opts.extension).concat(['.js']).map(function (ext) {
```
novemberborn: You can concat with strings. In other words `concat('.js')` is sufficient.

index.js
```diff
@@ -212,7 +220,15 @@ NYC.prototype._maybeInstrumentSource = function (code, filename, relFile) {
     return null
   }
 
-  return this.transform(code, {filename: filename, relFile: relFile})
+  var ext, transform
+  for (ext in this.transforms) {
```
novemberborn: I was thinking you could iterate over `this.extensions` here, though we're veering quite far into premature optimization land.

novemberborn:
@lloydcotten nice work so far! We can clean up the main `package.json` later I think, and let's discuss `opts` over `config` separately.

index.js
lloydcotten: I agree - it's more concise in the code to just use `for...in`.  And in most cases I would think we would have 2 or 3 extensions to iterate through at the most.  Yes it could be many files, but still I think it has to get into the 100s of thousands before any noticeable difference between `for..in` and `for`.

index.js
lloydcotten: Ha - learn something new everyday.  I did not know that.

test/fixtures/package.json
```diff
@@ -12,7 +12,8 @@
       "exclude": [
         "**/blarg",
         "**/blerg"
-      ]
+      ],
+      "extension": [".es6", ".foo.BAR"]
```
lloydcotten: Until it is easier to use opts over package.json config, I mixed the casing here a bit to test case-insensitive extension matching.

test/src/nyc-test.js
```diff
@@ -211,6 +248,42 @@ describe('nyc', function () {
       })
     })
 
+    describe('compile handlers for custom extensions are assigned', function () {
+      it('assigns a function to custom extensions', function () {
+        var nyc = new NYC({
+          cwd: path.resolve(__dirname, '../fixtures')
```
lloydcotten: I changed the `cwd` to use the fixtures `package.json` for these two tests, so that the looping can be tested as well as the case-insensitive matching in the next one.

bcoe:
@lloydcotten thanks for the hard work, I've been swamped this week but will get a release out Saturday.

novemberborn:
@lloydcotten this is great, thanks!

Looks like your PR now contains some duplicate commits that are already in `bcoe:master` though.

lloydcotten:
@bco, @novemberborn you're welcome.  Thanks for the cooperation and feedback on the PR.

Not sure why the duplicate commits - must've goofed with my squash somehow.  I'll take a look.

lloydcotten:
OK - I've cleaned up the commit history.  Should be good to go now.  Let me know if you need anything else on this PR.

bcoe:
@lloydcotten thanks for the wonderful contribution \o/

index.js
caub: why not making a ext => transform object/Map?

PR #167, opened 2016-02-14
all of nyc's configuration options can now also be set in package.json

bcoe:
This will need to be a major bump, because I've retired `config.nyc`.

This upgrades us to `yargs 4.x` which is a pretty major change (so that I can use the new `pkgConf` feature), this is a pretty major change and I would love for folks to take an extra close look to make sure I haven't broken anything.

Note that I could delete a bunch of code with yargs' improved command functionality \o/

Reviewers: @novemberborn, @jamestalmage, @jameswomack, @lloydcotten

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,9 +15,9 @@ var SourceMapCache = require('./lib/source-map-cache')
 var convertSourceMap = require('convert-source-map')
 var md5hex = require('md5-hex')
 var findCacheDir = require('find-cache-dir')
-var pkgUp = require('pkg-up')
-var readPkg = require('read-pkg')
 var js = require('default-require-extensions/js')
+var pkgUp = require('pkg-up')
+var yargs = require('yargs')
 
 /* istanbul ignore next */
 if (/index\.covered\.js$/.test(__filename)) {
@@ -25,17 +25,15 @@ if (/index\.covered\.js$/.test(__filename)) {
 }
 
 function NYC (opts) {
-  opts = opts || {}
+  var config = this._loadConfig(opts || {})
 
-  this._istanbul = opts.istanbul
-  this.subprocessBin = opts.subprocessBin || path.resolve(__dirname, './bin/nyc.js')
-  this._tempDirectory = opts.tempDirectory || './.nyc_output'
-  this._reportDir = opts.reportDir
-
-  var config = this._loadConfig(opts)
+  this._istanbul = config.istanbul
+  this.subprocessBin = config.subprocessBin || path.resolve(__dirname, './bin/nyc.js')
+  this._tempDirectory = config.tempDirectory || './.nyc_output'
+  this._reportDir = config.reportDir
   this.cwd = config.cwd
 
-  this.reporter = arrify(opts.reporter || 'text')
+  this.reporter = arrify(config.reporter || 'text')
 
   // load exclude stanza from config.
   this.include = false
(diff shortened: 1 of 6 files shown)


lloydcotten:
I took a look through this, looking mostly at how it might affect the extensions feature.  I think I like the changes.  I did try it out on on of my projects, but I'm getting an error from istanbul, I think that was introduced in this recent merge https://github.com/bcoe/nyc/commit/9fb7da46878fae2a335c213324ba492e1f950289 from @novemberborn 

I'll try to drill down to see what's causing it more specifically and comment over on that PR.

bcoe:
@lloydcotten if you delete `node_modules/.cache` does it still crash?

I was seeing a similar problem, we should get to the bottom of this before we publish.

lloydcotten:
Hmm... you are right.  That does resolve it.  I was seeing this:
```
/home/lloyd/dev/github/nyc/node_modules/istanbul/lib/report/html.js:288
                    text = structuredText[startLine].text;
                                                    ^

TypeError: Cannot read property 'text' of undefined
    at /home/lloyd/dev/github/nyc/node_modules/istanbul/lib/report/html.js:288:53
    at Array.forEach (native)
    at annotateBranches (/home/lloyd/dev/github/nyc/node_modules/istanbul/lib/report/html.js:255:30)
    at HtmlReport.Report.mix.writeDetailPage (/home/lloyd/dev/github/nyc/node_modules/istanbul/lib/report/html.js:426:9)
    at /home/lloyd/dev/github/nyc/node_modules/istanbul/lib/report/html.js:489:26
    at SyncFileWriter.extend.writeFile (/home/lloyd/dev/github/nyc/node_modules/istanbul/lib/util/file-writer.js:57:9)
    at FileWriter.extend.writeFile (/home/lloyd/dev/github/nyc/node_modules/istanbul/lib/util/file-writer.js:147:23)
    at /home/lloyd/dev/github/nyc/node_modules/istanbul/lib/report/html.js:488:24
    at Array.forEach (native)
    at HtmlReport.Report.mix.writeFiles (/home/lloyd/dev/github/nyc/node_modules/istanbul/lib/report/html.js:482:23)
```

bcoe:
@lloydcotten do you caching turned on?

lloydcotten:
Yes, I had caching on.

novemberborn:
@bcoe @lloydcotten the cache uses a salt derived from NYC's and Istanbul's version numbers. The version hasn't yet changed from before my path changes (it's still `5.5.0`) so it may have used something that was actually too old. The risk of using code from GitHub branches I suppose.

novemberborn:
@bcoe looks good!

I took the liberty of pushing a commit which fixes some duplicate config tests. We're now also using `fixtures/package.json` to test config, allowing me to remove `.es6` extension from the actual `package.json`.

bcoe:
@novemberborn from your discussion with @lloydcotten, it sounds like the act of bumping the version will fix the cache issue?

I'll make this `nyc@6.0.0`, since it changes config's functionality quite a bit.

novemberborn:
@bcoe I'm not sure why cache entries from an older version don't work, but if clearing the cache fixes the issue, then yes bumping the version should preclude any older cache entries from being used.

index.js
```diff
@@ -25,17 +25,15 @@ if (/index\.covered\.js$/.test(__filename)) {
...
-  this._istanbul = opts.istanbul
-  this.subprocessBin = opts.subprocessBin || path.resolve(__dirname, './bin/nyc.js')
-  this._tempDirectory = opts.tempDirectory || './.nyc_output'
-  this._reportDir = opts.reportDir
+  var config = this._loadConfig(opts || {})
 
-  var config = this._loadConfig(opts)
+  this._istanbul = config.istanbul
+  this.subprocessBin = config.subprocessBin || path.resolve(__dirname, './bin/nyc.js')
+  this._tempDirectory = config.tempDirectory || './.nyc_output'
```
jameswomack: Why is `path.resolve` used on `subprocessBin` but not `subprocessBin`?

bcoe: @jamestalmage I think the reasoning was that:

* if you explicitly specify the path to the bin that should be run, we assume you know the path that you want.
* otherwise we default to the bin that we know will be relative to `index.js`.

PR #170, opened 2016-02-20
update changelog for 6.x release

bcoe:


bcoe:
the problem turned out to be the context of the `require` statement, I fixed the problem like so:

https://github.com/yargs/yargs/commit/28954bbac5b6c860fd7a8908fd2670856ef95d1d

PR #177, opened 2016-02-29
Add an EditorConfig and clean up readme

JaKXz:
Note: introducing these changes raises a question in the coveralls
example at the bottom: does `nyc` really need to be run again for the
report to be available? and wouldn’t it be better just to have the lcov
report generated alongside the regular text display, and have the
`after_success` command just `cat` and pipe it to `coveralls`?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+
+# Change these settings to your own preference
+indent_style = space
+indent_size = 2
+
+# We recommend you to keep these unchanged
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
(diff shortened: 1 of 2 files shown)


JaKXz:
@bcoe please advise and I'll update :)

bcoe:
It's important to note that running the tests outputs coverage info as JSON in `.nyc_output`, and that running:

`nyc report --reporter=text-lcov | coveralls`

Does not re-run tests. So this approach is fast, and quite similar to simply using `cat` to pipe to `coveralls`.

One lovely advantage of letting the `nyc` bin output `lcov` info directly, is that the approach is cross-platform (I don't believe that Windows ships with `cat`); tldr; I'm tempted to stick with this as the example code.

Thank you for this contribution \o/ I hope that you'll continue to contribute more as you work with `nyc` (I saw the additional tickets that you opened :smile:).

bcoe:
@JaKXz give this a shot:

`npm i nyc@next`

shipped with your editor config changes.

Issue #174, opened 2016-02-28
--all doesn't use configured file extensions


novemberborn:
#163 introduced configurable file extensions however `--all` only selects `**/*.js` files. We could generate that pattern for all configured extensions and then use [`globby`](https://www.npmjs.com/package/globby) to load the files.

novemberborn:
@lloydcotten perhaps you'd be interested in improving this? No obligation of course :smiley_cat:

lloydcotten:
Hmm... I see.  Yes, perhaps I could look at that.

PR #178, opened 2016-03-01
Include extra extensions in glob for --all switch

lloydcotten:
This adds support to automatically include configured extensions when using the --all switch as per issue #174.

I simply modified an existing test to cover this scenario.


bcoe:
@lloydcotten great catch/fix :+1:

my only comment, is that I think it would be worth adding an additional test. something like:

`outputs an empty coverage report for all supported extensions`

rather than tacking this logic onto the existing test.

novemberborn:
Cool. +1 on making it a separate test.

lloydcotten:
@bcoe @novemberborn I sat there and debated whether I should do that extra test or not.  I suppose I should know by now that if in doubt, I should just do a new test.  :-)

It's updated now.

novemberborn:
:shipit:

lloydcotten:
@bcoe Any idea why `master` branch is failing after the merge?  Seems to be failing on an unrelated test.

lloydcotten:
Should mention that is on Travis - passes OK locally for me.

bcoe:
@lloydcotten mind giving this branch a shot:

https://github.com/bcoe/nyc/pull/180

should fix the broken build on master ... not quite sure why that cropped up when it did.

bcoe:
@lloydcotten give this a shot:

`npm i nyc@next`

shipped with your `--all` fixes. I'll release it to `latest` soon if everyone's testing goes well.

PR #180, opened 2016-03-04
do not infer cwd from pkgUp, if pkgUp walks all the way to /

bcoe:
The library `pkgUp` is used to find the closest package.json, which nyc uses as `cwd`.

If `pkgUp` walks all the way to `/`, it returns the string `package.json`. In this edge-case, we should not switch to `cwd` to the path of the package.json found by `pkgUp`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -71,7 +71,7 @@ NYC.prototype._loadConfig = function (opts) {
   var cwd = opts.cwd || process.env.NYC_CWD || process.cwd()
   var pkgPath = pkgUp.sync(cwd)
 
-  if (pkgPath) {
+  if (pkgPath && pkgPath !== 'package.json') {
     cwd = path.dirname(pkgPath)
   }
 
--- a/index.js
+++ b/index.js
@@ -71,7 +71,7 @@ NYC.prototype._loadConfig = function (opts) {
   var cwd = opts.cwd || process.env.NYC_CWD || process.cwd()
   var pkgPath = pkgUp.sync(cwd)
 
-  if (pkgPath && pkgPath !== 'package.json') {
+  if (pkgPath) {
     cwd = path.dirname(pkgPath)
   }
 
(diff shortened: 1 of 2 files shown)


jamestalmage:
This almost seems like a pkgUp bug. Shouldn't it return null?

novemberborn:
> This almost seems like a pkgUp bug. Shouldn't it return null?

It will. [pkg-up](https://github.com/sindresorhus/pkg-up/blob/2b890bcb11632eea38c01075104fcdd330e6c65d/index.js#L9) --> [find-up](https://github.com/sindresorhus/find-up/blob/bc536485bfddd2d960f472964e65afef9895f475/index.js#L53).

Any chance you have a `/package.json` on your machine @bcoe?

bcoe:
@novemberborn @jamestalmage I have no `/package.json`, and we were seeing the same behavior on Travis/Appveyor which was breaking tests.

I'm really confused because it feels like this should have always been broken, and `pkg-up` and `find-up` only have one release (and haven't changed in a while). This fixes things, but I'm trying to figure out which dependency of a dependency caused behavior to change  ¯\_(ツ)_/¯

novemberborn:
It's a bug in `find-up`, introduced by <https://github.com/sindresorhus/find-up/pull/3> and released 3 days ago. Once it runs out of path parts to try it accidentally uses the working directory and returns a relative path.

novemberborn:
This will be fixed with <https://github.com/sindresorhus/find-up/pull/4>.

sindresorhus:
find-up@1.1.2 is out with @novemberborn's fix. Sorry about the trouble.

novemberborn:
@bcoe why merge this? `find-up` had a bug but that's resolved now.

bcoe:
@novemberborn I figured it was worth merging an explicit bump of `find-up`:

` "find-up": "^1.1.2",`

novemberborn:
@bcoe ah sorry, hadn't checked the diff :blush:

Issue #184, opened 2016-03-06
`reporter` configuration in `package.json` not working


rapzo:
When provided an array or even a string of reporters they are not taking into account due to this override: https://github.com/bcoe/nyc/blob/master/bin/nyc.js#L152
Taking it out fixes the issue, afaic.

This happens even in version 6.0.0, which branch i couldn't find, and master seems to be at v5.5.0 still, otherwise i'd provide the PR.

Or am i missing something?

Thanks guys.

rapzo:
Fixed. #185

PR #185, opened 2016-03-06
Removed yargs default for reporters so the ones defined in  are taken first

rapzo:
As pointed in [issue #184](https://github.com/bcoe/nyc/issues/184).

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -23,8 +23,8 @@ var yargs = require('yargs')
       .usage('$0 check-coverage [options]')
       .example('$0 check-coverage --lines 95', "check whether the JSON in nyc's output folder meets the thresholds provided")
   })
-  .option('r', {
-    alias: 'reporter',
+  .option('reporter', {
+    alias: 'r',
     describe: 'coverage reporter(s) to use',
     default: 'text',
     global: true


bcoe:
wouldn't this approach make it so that it's impossible to override the reporter:

`nyc --reporter=lcov-text`

the goal is a solution that uses the reporter defined in `package.json` `iff` the default reporter has been set by yargs?

rapzo:
Yeah, you're right...
Well, i thought the expected behaviour was to run through the list of reporters defined in `package.json` when no reporter was provided by yargs. Shouldn't it be?

rapzo:
Went a bit up and removed the yargs' default for reporters since it is set [here](https://github.com/bcoe/nyc/blob/master/index.js#L36) when none provided. Seems to work both ways.

bcoe:
@rapzo I dug into this a bit more, there's a slightly better solution that allows us to continue having a default value:

```js
  .option('reporter', {
    alias: 'r',
    describe: 'coverage reporter(s) to use',
    global: true
  })
```

there appears to be a bug with yargs, which is taking into account aliases when loading defaults from the `nyc` stanza in the package.json. A temporary fix is to edit `bin/nyc.js` so that the long option name is used for the option rather than the alias.

I'd happily land this patch, but we should also dig into this bug in yargs.

rapzo:
Sorry for the delay. Here's the patch. A better solution indeed.

bcoe:
@rapzo give this a shot, shipped with your fixes for package configuration:

`npm i nyc@next`

If everything is looking good I'll promote the change to `latest` soon.

rapzo:
I will! Thanks man!

Issue #209, opened 2016-04-01
Return non-zero exit code when coverage is below required minimum


tp:
Would you be open to discuss the possibility to return a non-zero exit code when the coverage is below the minimum thresholds set?

Or do you feel like this should be handled by external systems, working on the coverage data?

addaleax:
Is [this](https://github.com/bcoe/nyc#checking-coverage) what you are looking for?

tp:
Yes, adding `--check-coverage=true` gets the behavior that I wanted. :-) Thanks!

tp:
The only issue I have found is that this still returns 0 even when ava (the test command I use) returns a non-zero exit code because tests failed.

Wondering how to best check the test result and coverage in one go, or whether I need to analyze the output of the different tools to get a combined result.

addaleax:
I think that’s actually a bug, yep. For now, you should be able to run `nyc ava …` and `nyc check-coverage` separately to check their exit codes.

bcoe:
@tp shoot, yes we should patch this! please open an issue if you have a second, and patches very much welcome.

PR #210, opened 2016-04-02
Use test program’s exit code even with `--check-coverage`

addaleax:
Resolve the issue mentioned in #209 that, when running with `--check-coverage`, the coverage check’s exit code overrides the test command’s. This depends on tapjs/foreground-child#7 in order to work (without it, the current behaviour is simply retained).

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -129,9 +129,17 @@ if (argv._[0] === 'report') {
   }
   sw([wrapper], env)
 
+  // Both running the test script invocation and the check-coverage run may
+  // set process.exitCode. Keep track so that both children are run, but
+  // a non-zero exit codes in either one leads to an overall non-zero exit code.
+  process.exitCode = 0
   foreground(nyc.mungeArgs(argv), function (done) {
+    var mainChildExitCode = process.exitCode
+
     if (argv.checkCoverage) {
       checkCoverage(argv, function (done) {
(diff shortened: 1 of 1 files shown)


bcoe:
@addaleax amazing, I'm at a bar mitzvah this morning :wine_glass: :tada: but I'm hoping to lock myself in a room tonight to work on yargs and nyc.

addaleax:
@bcoe Take your time! I don’t think this is very urgent for anyone, so enjoy yourself :smile:

bcoe:
@addaleax this is looking great... I'm feeling like it's getting to the point where we might want to start adding some simple tests for the CLI though -- given that the logic continues getting more and more complex. I wonder if we could get away with simply spawning a subprocess and then asserting against the output?

bcoe:
@addaleax give this a spin:

`npm i nyc@next`

this release candidate has landed your spawn-wrap and foreground-child changes.

novemberborn:
> I'm feeling like it's getting to the point where we might want to start adding some simple tests for the CLI though -- given that the logic continues getting more and more complex. I wonder if we could get away with simply spawning a subprocess and then asserting against the output?

This is what AVA does. See <https://github.com/sindresorhus/ava/blob/master/test/cli.js>.

addaleax:
Yup, everything seems to work now with manual testing. :tada:

I think I can get tests for this and maybe the `npm` stuff together.

Issue #207, opened 2016-03-28
`include` option doesn't supported in CLI


rpominov:
Can it be added to CLI options? I can submit a PR if it'll help.

jamestalmage:
Do you have a use case where you are unable to just use the config in `package.json`?

rpominov:
Yeah, I'm building a tool that basically gives you an opinionated dev environment. It runs test compiles js etc. Also it does coverage, so nyc is just implementation detail of that tool, and nyc config have to be hidden from the tool's user. Currently I just pass all options via CLI, which works fine, except `include` can't be specified via CLI options.

Here is the link if you're interested https://github.com/rpominov/lobot . I just run `lobot coverage` or `lobot coveralls` and it does stuff using `nyc` under the hood.

PR #214, opened 2016-04-03
Add `include` to CLI options

rpominov:
See #207

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -116,7 +116,8 @@ if (argv._[0] === 'report') {
   if (!Array.isArray(argv.extension)) argv.extension = [argv.extension]
 
   var nyc = (new NYC({
-    require: argv.require
+    require: argv.require,
+    include: argv.include
   }))
   nyc.reset()
 
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -47,13 +47,13 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
     describe: 'whether or not to instrument all files of the project (not just the ones touched by your test suite)'
   })
   .option('exclude', {
-    alias: 'ex',
+    alias: 'x',
     default: [],
     describe: 'a list of specific files and directories that should be excluded from coverage, glob patterns are supported, node_modules is always excluded'
   })
   .option('include', {
-    alias: 'in',
-    default: ['**'],
+    alias: 'n',
+    default: [],
     describe: 'a list of specific files that should be covered, glob patterns are supported'
   })
   .option('require', {
@@ -120,6 +120,7 @@ if (argv._[0] === 'report') {
   if (!Array.isArray(argv.require)) argv.require = [argv.require]
   if (!Array.isArray(argv.extension)) argv.extension = [argv.extension]
   if (!Array.isArray(argv.exclude)) argv.exclude = [argv.exclude]
+  if (!Array.isArray(argv.include)) argv.include = [argv.include]
 
   var nyc = (new NYC({
     require: argv.require,
@@ -142,6 +143,9 @@ if (argv._[0] === 'report') {
   }
   if (argv.exclude.length) {
     env.NYC_EXCLUDE = argv.exclude.join(',')
+  }
+  if (argv.include.length) {
+    env.NYC_INCLUDE = argv.include.join(',')
   }
   sw([wrapper], env)
 
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -46,6 +46,11 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
     type: 'boolean',
     describe: 'whether or not to instrument all files of the project (not just the ones touched by your test suite)'
   })
+  .option('include', {
+    alias: 'in',
+    default: [],
+    describe: 'a list of files that should be covered'
+  })
   .option('require', {
     alias: 'i',
     default: [],
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -46,6 +46,11 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
     type: 'boolean',
     describe: 'whether or not to instrument all files of the project (not just the ones touched by your test suite)'
   })
+  .option('exclude', {
+    alias: 'ex',
+    default: [],
+    describe: 'a list of specific files and directories that should be excluded from coverage, glob patterns are supported, node_modules is always excluded'
+  })
   .option('include', {
     alias: 'in',
     default: ['**'],
@@ -117,7 +122,8 @@ if (argv._[0] === 'report') {
 
   var nyc = (new NYC({
     require: argv.require,
-    include: argv.include
+    include: argv.include,
+    exclude: argv.exclude
   }))
   nyc.reset()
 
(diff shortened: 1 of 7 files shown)


bin/nyc.js
```diff
@@ -46,6 +46,11 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
     type: 'boolean',
     describe: 'whether or not to instrument all files of the project (not just the ones touched by your test suite)'
   })
+  .option('include', {
+    alias: 'in',
```
rpominov: `i` is taken unfortunately.

bin/nyc.js
```diff
@@ -46,6 +46,11 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
...
+    default: [],
+    describe: 'a list of files that should be covered'
```
rpominov: Not sure whether this description is good enough.

rpominov:
Also not 100% I did everything right, but I tested it in my project and it works fine.

bin/nyc.js
novemberborn: [From the readme](https://github.com/bcoe/nyc#including-files), perhaps "specific files that should be covered".

The readme then shows glob patterns in an example without explicitly mentioning pattern support. So maybe add that too.

novemberborn:
If we add `include` we should also add `exclude`.

bin/nyc.js
rpominov: In readme the description is based on `exclude`, should we do the similar here after we add `exclude` too?

bin/nyc.js
```diff
@@ -46,6 +46,11 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
     type: 'boolean',
     describe: 'whether or not to instrument all files of the project (not just the ones touched by your test suite)'
   })
+  .option('include', {
+    alias: 'in',
+    default: [],
```
rpominov: Also from readme:

> Note: include defaults to ['**']

But this default value probably defined somewhere else in the code. Should we remove it there, and specify `['**']` here. Although we need default value when it's loaded from package.json too, so maybe only specify here and not remove in other place?

bin/nyc.js
novemberborn: Not sure. The include section in the readme follows the exclude section. I suspect users may view CLI usage information somewhat differently.

bin/nyc.js
novemberborn: The code arrives at that `**` default implicitly. I think we'd need a guard [here](https://github.com/bcoe/nyc/blob/deee980b965cbe70b2da2c70b28ce9fc0b4a65d5/index.js#L40) for empty `config.include` arrays.

bin/nyc.js
```diff
@@ -46,6 +46,16 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
...
     describe: 'whether or not to instrument all files of the project (not just the ones touched by your test suite)'
   })
+  .option('exclude', {
+    alias: 'ex',
+    default: [],
+    describe: 'a list of specific files and directories that should be excluded from coverage, glob patterns are supported, node_modules is always excluded'
+  })
+  .option('include', {
+    alias: 'in',
+    default: ['**'],
```
rpominov: I've set `['**']` so the help would like this:

```
--include, --in   a list of specific files that should be covered, glob
                  patterns are supported                     [default: ["**"]]
```

rpominov:
I haven't check yet that the latest changes work. Will check in my project a bit later.

rpominov:
Checked:

- [x] `--all` and `include` not set: all files included
- [x] `--all --include **/src/**`: only src included
- [x] `--all` and `include: "**/src/**"` in package.json: only src included 
- [x] `--all` and `exclude: "**/src/**"` in package.json: src excluded
- [ ] `--all --exclude **/src/**`: src excluded

Looking into the last case.

rpominov:
The last commit fixes last case.

bcoe:
@rpominov I really like this work :+1: 

would it be possible to add add a test; no need to add one for the CLI (since we don't have any tests for this yet, although it would be great to have some eventually). But it would be nice to have tests for the new logic that appends a list of excludes provided to the default excludes, and logic for the behavior around an include that's an empty list.

Also, mind rebasing to resolve conflicts :+1:

rpominov:
@bcoe Ok, no problem. I think I'll be able to do this at the end of the week.

rpominov:
I've rebased and added the tests, hope they're ok.

Also I wasn't able to make all current tests pass on my machine (in master, without changes of this PR). In particular [this](https://github.com/bcoe/nyc/blob/3403ca17991fded623a05fda3e3398375320e1d9/test/src/nyc-test.js#L266) assertion fails, and [this](https://github.com/bcoe/nyc/blob/3403ca17991fded623a05fda3e3398375320e1d9/test/src/nyc-bin.js#L130) code fails with "file already exists" error. Probably this is just something wrong with my setup...

Also, man, running those tests splitted to individual files takes forever :hushed:

bin/nyc.js
novemberborn: `include` is used in [this line](https://github.com/bcoe/nyc/blob/3403ca17991fded623a05fda3e3398375320e1d9/index.js#L191):

```js
return (!this.include || micromatch.any(relFile, this.include)) && !micromatch.any(relFile, this.exclude)
```

I think with this change, when `nyc` is used through the CLI, it'll always call `micromatch.any`. Without a default it would just short-circuit the logic. We should definitely have the default in the documentation but maybe we should also [set in the constructor](https://github.com/bcoe/nyc/blob/3403ca17991fded623a05fda3e3398375320e1d9/index.js#L39:L42) so we can remove that logic, *or* set `this.include` to `false` if it's just `**`, so we can avoid the unnecessary match.

The difference between the CLI codepath and the embedding code path bothers me a little :smile:

novemberborn:
LGTMO!

bin/nyc.js
```diff
@@ -109,9 +119,12 @@ if (argv._[0] === 'report') {
   // wrap subprocesses and execute argv[1]
   if (!Array.isArray(argv.require)) argv.require = [argv.require]
   if (!Array.isArray(argv.extension)) argv.extension = [argv.extension]
+  if (!Array.isArray(argv.exclude)) argv.exclude = [argv.exclude]
 
   var nyc = (new NYC({
-    require: argv.require
+    require: argv.require,
+    include: argv.include,
```
rpominov: @novemberborn Maybe we could check for `**` here before passing to `new NYC`?

novemberborn: Yea, or we change the `NYC` implementation. Let's see what @bcoe suggests.

bcoe:
@novemberborn @rpominov I made a couple small tweaks, and added tests. see:

https://github.com/rpominov/nyc/pull/4

rpominov:
@bcoe Thanks, merged. The commits log of this PR gets a little messy, should I squash the commits?

bcoe:
@rpominov no worries 👍 I've started using the squash button when I merge commits, along with this slick new library:

https://github.com/conventional-changelog/standard-version

rpominov:
Ah, right, forgot that github now can do squash. And standard-version looks very cool :+1:

bcoe:
@rpominov this is now released in `nyc@6.4.0`.

rpominov:
@bcoe Great. Thanks!

PR #216, opened 2016-04-03
bundle dependencies to solve singleton issues we've seen

bcoe:
* I've started bundling dependencies in an attempt to address some of the singleton issues that we've seen in `npm@3.x`:
  * https://github.com/bcoe/nyc/issues/205
* I'm also floating a patch that addresses an issue with `nyc npm` on `node@5.8.x`:
  * https://github.com/bcoe/nyc/issues/190


novemberborn:
Big diff! :wink: 

Any specific feedback you're looking for @bcoe? Seems like a good approach to avoid nyc from interfering with application or test code.

jamestalmage:
@bcoe - Is there away to deploy bundled dependencies to npm without adding them to the git repo?

bcoe:
@jamestalmage we can bundle dependencies without checking in, all the files are a patch I'm currently floating for `spawn-wrap`.

Issue #198, opened 2016-03-20
Bad application of source maps to coverage reports for ternary expressions


skozin:
I don't know for sure where I should report this issue, so sorry if it doesn't belong here.

When I generate coverage report with `nyc ava && nyc report --reporter=html`, the resulting report shows incorrect highlighting of the source code. And, the most critically, overall coverage numbers for branches and functions look incorrect (much less than expected).

I have prepared [the repo](https://github.com/skozin/nyc-babel-issue-repro) with the reproduction case.<br>Also, I've uploaded the report to [this GH pages site](http://skozin.github.io/nyc-babel-issue-repro/nyc-babel-issue-repro/index.html).

Here is a screenshot that shows incorrect highlighting:

![image](https://cloud.githubusercontent.com/assets/1699593/13906228/6f55a302-eee2-11e5-8069-df3ef42d27c5.png)

Annotations on the left suggest that `_takeFromWaitingPublisher()` function has been executed multiple times, and that's indeed the case as the test code is crafted to ensure this. Yet it is highlighted as not being covered.

The `_take()` function also has a similar problem, although its code is highlighted in yellow, which means "branch not covered". However, the highlighted code has clearly been executed multiple times.

Also, the highlighting sometimes starts and ends at positions that make no sense, e.g. between "t" and "h" letters inside `this` keyword.

bcoe:
@skozin does the percentage of coverage seem correct, but the highlighting off? This is most likely a problem with how we remap from the sourcemap to the coverage files used by Istanbul; @novemberborn is a subject matter expert on this part of the codebase.

skozin:
@bcoe no, 

> And, the most critically, overall coverage numbers for branches and functions look incorrect (much less than expected).

skozin:
It reports 75% Branches (33/44) and 66.67% Functions (4/6), although all functions except one have been executed, and almost all branches inside them was taken.

skozin:
I've simplified the case a little. Now nyc reports 73.68% Branches (28/38) and 66.67% Functions (4/6). The function coverage I expect to see is 80%:

```
Covered:

constructor: 1x
take: 24x
anon inside take: 24x
_take: 24x
_takeFromWaitingPublisher: 9x
_triggerWaiters: 9x
_emitDrain: 1x
assert: 49x

Not covered:

anon at the end of _take
nop

Total:

80% (8/10)
```

skozin:
[...]

Issue #183, opened 2016-03-06
Implement --all without requiring modules


novemberborn:
Currently when `--all` is used we require all modules that match a glob pattern. This causes these files to be instrumented and a baseline coverage to be established in the main process.

Requiring all modules has a few problems:

* The files may need to be transpiled, so `--require` needs to be used
* The files may throw when loading because they're in the wrong environment (#175, #172)
* If the subprocess has its own require hook we may end up installing duplicate hooks (#181)

Instead what we should do:

1. Get a list of modules matching the glob pattern, without requiring them
2. Run the subprocesses
3. Built the coverage report for the subprocesses
4. For all modules found in step 1, that are not in the coverage report, synthesize coverage reports directly in nyc without instrumenting the files. We can generate these for the actual source and no transpiling should be necessary

Once we do this we may even be able to remove the `--require` flag, shifting transpilation responsibilities to the subprocess.

bcoe:
this should be addressed in https://github.com/istanbuljs/nyc/pull/286

Issue #168, opened 2016-02-20
Code mapping after an error occurs


aureooms:
I'm running nyc with ava. When a runtime error occurs in a test (e.g. trying to access a key of `undefined`), ava outputs the stacktrace with lineno:col inside the instrumented file, not the original file. Would it be possible to map the instrumented code back to the original code and output the relevant lineno:col?

novemberborn:
@aureooms nyc uses [Istanbul](https://github.com/gotwarlost/istanbul) to instrument the files. You'd probably need Istanbul to include a source map when it instruments a file, and it would have to take into account input source maps in case of ES2015 transpilation. Then it may or may not work automatically with AVA.

I don't think Istanbul supports this currently so you'd have to open an issue there.

bcoe:
@aureooms give this a spin:

https://github.com/bcoe/nyc#use-babel-plugin__coverage__-for-better-es6es7-support

we've released better support for ES6/ES6 coverage using the `__coverage__` instrumentation library.

Issue #117, opened 2015-12-26
Split out the caching and source-map-fixing into separate modules.


jamestalmage:
NYC now does 3 things:
1. Installs an (optionally cacheable) transform via `append-transform` that captures coverage and source-map data.
2. Remaps istanbul reports using source-map data.
3. Uses `spawn-wrap` to make all this happen across forked processes.

I think `1` and `2` need to be split out.

sindresorhus:
:+1:

novemberborn:
> Remaps istanbul reports using source-map data.

There's <https://github.com/SitePen/remap-istanbul> but it already does way too many things besides remapping the reports.

Even if 1 and 2 are split out they may still end up coupled to each other. Not sure I see the point, unless they're moved into Istanbul itself maybe.

jamestalmage:
I think `1` would almost certainly depend on `2` (but not necessarily the other way around).

> Not sure I see the point

It would allow all our hard work getting source-maps and caching working to be used in other contexts.

> unless they are moved into Istanbul itself maybe

That may be the ideal solution, but I have got to imagine that there is a good reason that source-map support hasn't found it's way to master yet over there. 

// @gotwarlost

novemberborn:
> It would allow all our hard work getting source-maps and caching working to be used in other contexts.

Sure, assuming those other contexts discover the modules.

jamestalmage:
> Sure, assuming those other contexts discover the modules.

You take the risk of being ignored every time you type `npm publish`. That's just life.

For my part, I am really curious how much overhead `spawn-wrap` and `foreground-child` add? How much faster would AVA be if we used those split out modules and provided a `--coverage` option that called them in every forked process? Our `babel.js` script runs before any user scripts, so we can already provide many of the guarantees that `spawn-wrap` provides for NYC. Of course, that wouldn't cover users who are going to be forking again from within their test files. In that case - they need to use NYC.

bcoe:
@jamestalmage it looks like @gotwarlost has been working on his own source-map library as part of the istanbuljs org, could we look at merging efforts?

coreyfarrell:
Please correct me if I'm wrong but this is resolved by the existence of `caching-transform` and `istanbul-lib-source-maps`?

coreyfarrell:
I think caching is resolved and any further source-map work should be incorporated to istanbul-lib-source-maps.

PR #217, opened 2016-04-04
Replace Istanbul for instrumentation?

novemberborn:
Issues like #198, #183 and #168 have caused me to reflect on nyc's use of Istanbul. I think we should replace Istanbul for our instrumentation purposes. To this end I've started [Copenhagen](https://github.com/novemberborn/copenhagen). This PR is a proof-of-concept of integrating Copenhagen with nyc. You can find a sample project at <https://github.com/novemberborn/nyc-copenhagen-poc>.

Please read the [Background section](https://github.com/novemberborn/copenhagen#background) for context but the tl;dr is that Copenhagen will enable us to:

* Instrument the new JavaScript syntax, rather than trying to map the coverage report
* Fix stack traces for instrumented code
* Support languages like TypeScript
* Report on all files without having to execute them

Thoughts?


sarbbottam:
>Copenhagen

👍

novemberborn:
@bcoe would love your thoughts on this when you have a chance.

bcoe:
@novemberborn haven't missed this pull request; just want to collect my thoughts and give some good feedback (busy week).

novemberborn:
@bcoe no worries :)

bcoe:
@novemberborn I'm going to dig into this code more thoroughly this weekend, and play. But wanted to give you some initial feedback.

I think the idea of a competing code-coverage library that is suited to ES2015 is smart, and will help push forward both Istanbul and Copenhagen. I also feel, however, that Istanbul has a lot going for it:

* a huge community.
* a lot of time to learn from the community, and address interesting edge-cases.
* some really solid developers supporting the project.

What I'd love to see is this:

* abstract out nyc so that we could plug in an alternative code-coverage library.
* work closely with the Istanbul community to figure out the path forward to creating a solution that has really rock-solid support for ES2015 features.

novemberborn:
> I also feel, however, that Istanbul has a lot going for it

Absolutely! I think it does several things at once though:

* Instrument code and collect usage data
* Interpret coverage data
* Generate various reports
* Provide a CLI

Copenhagen only covers the first item, but with the right breakdown for integrating with tools like nyc and AVA. Plus it provides a way of instrumenting any source that compiles to JavaScript.

bcoe:
@novemberborn could we work on building out a suitable integration point in Istanbul, such that we could configure an alternative instrumenter?

jamestalmage:
I think #117 is a good place to start with this. If we modularize there various components of `nyc`, hopefully @novemberborn can use that to develop `nyc-copenhagen` (or whatever). I would certainly be excited to participate in both projects.

novemberborn:
> could we work on building out a suitable integration point in Istanbul, such that we could configure an alternative instrumenter?

@bcoe, but we're essentially already doing that. nyc decides what files to instrument, calls out to Istanbul to do the actual instrumentation, stores the coverage data, and then combines it all so Istanbul can generate its reports.



> I think #117 is a good place to start with this. If we modularize there various components of nyc, hopefully @novemberborn can use that to develop nyc-copenhagen (or whatever). I would certainly be excited to participate in both projects.

@jamestalmage yea I've thought about that. The reporting code that Copenhagen generates will be able to take input source maps into account, and then it won't be hard to map the raw usage data back to the source files. Of course you'd need to generate reporting code for the original sources.

There would have to be some glue code in nyc (mostly to map sources to the reporting code etc) but that'd be pretty minimal.

bcoe:
@novemberborn would definitely land this if you can get it to the point that tests pass, and we can swap in copenhagen is a drop in replacement for istanbul -- I think probably istanbul should be the default, and we'd point folks at copenhagen for ES2015?

JaKXz:
> Support languages like TypeScript

👍 👍 👍

novemberborn:
@bcoe cool! I'll have a think about structuring the required work across the various projects.

> I think probably istanbul should be the default, and we'd point folks at copenhagen for ES2015?

Yea that'd be a good place to start. We'll have to see how Copenhagen shapes up.

bcoe:
@novemberborn closing this pull for now; waiting with bated breath for the drop in API replacement.

PR #199, opened 2016-03-21
add more exclude glob patterns

kentcdodds:
I personally use (and teach) that applications should co-locate test files and source code, which is where the `*.test.js` comes from.

The React community uses the `__tests__` convention.

Happy to make any changes you'd like to see :-)

PR #218, opened 2016-04-04
fix glob pattern for exclusion and add tests

kentcdodds:
Should have added tests in #199. My bad. I added tests here for both of the globs added in #199 and I tested it manually in my own project. Works great. Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,7 +42,7 @@ function NYC (opts) {
   }
 
   this.exclude = this._prepGlobPatterns(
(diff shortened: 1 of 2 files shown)


bcoe:
@kentcdodds give this a shot:

`npm i nyc@next`

It should support your new exclude patterns, if everything is looking good over the next day or so we can promote it to the stable release.

kentcdodds:
Works great! Thanks! Opened #223 to update the documentation. Sorry I had to do three PRs to get this simple feature in ¯\\\_(ツ)_/¯

netflix-full-house-fuller-3ornjKEREhVsOmQG1q

PR #219, opened 2016-04-05
introducing conventional-changelog for changelog generation

bcoe:
starting to introduce conventional-changelog to [several projects](https://github.com/yargs/yargs/blob/master/CHANGELOG.md), thanks @stevemao \o/


jamestalmage:
@bcoe - It's probably worth adding an intro/link to conventional changelog in the contributing guidelines.

PR #223, opened 2016-04-05
improve documentation on excluding files

kentcdodds:


README.md
```diff
@@ -142,8 +142,9 @@ directory:
 }
 ```
 
-> Note: exclude defaults to `['test', 'test{,-*}.js']`, which would exclude
-the `test` directory as well as `test.js` and `test-*.js` files
+> Note: exclude defaults to `['test', 'test{,-*}.js', '**/*.test.js', '**/__tests__/**']`,
+which would exclude `test`/`__tests__` directories as well as `test.js`, `*.test.js`,
+and `test-*.js` files. By specifying your own `exclude` property, this overrides these defaults
```
jamestalmage: Specifying your own `exclude` property overrides these defaults.

kentcdodds: Thanks! I don't know why I had such a hard time with that copy :-)

The PR's been updated!

PR #233, opened 2016-04-21
Provide codecov instructions

rgbkrk:
Thanks @bcoe for `nyc` and @stevepeak for `codecov`

/cc https://github.com/codecov/codecov-node/issues/13

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,3 +240,42 @@ integrated with coveralls and travis-ci.org:
 That's all there is to it!
 
 > Note: by default coveralls.io adds comments to pull-requests on GitHub, this can feel intrusive. To disable this, click on your repo on coveralls.io and uncheck `LEAVE COMMENTS?`.
+
+
+## Integrating with codecov
+
+nyc npm test && nyc report --reporter=text-lcov > coverage.lcov && codecov
+
+[codecov](https://codecov.io/) is a great tool for adding
+coverage reports to your GitHub project, even viewing them inline on GitHub with a browser extension:
+
+![browser extension](https://d234q63orb21db.cloudfront.net/ad63907877249140772dff929ad1c340e424962a/media/images/next/extension.png)
+
+Here's how to get `nyc` integrated with codecov and travis-ci.org:
+
+1. add the codecov and nyc dependencies to your module:
+
+  ```shell
+  npm install codecov nyc --save-dev
+  ```
+
+2. update the scripts in your package.json to include these bins:
+
+  ```json
+  {
+     "script": {
+       "test": "nyc tap ./test/*.js",
+       "coverage": "nyc report --reporter=text-lcov > coverage.lcov && codecov"
+     }
+  }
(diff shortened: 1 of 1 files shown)


wuct:
Thanks for your doc! I have played around `codecov` for a while and it turned out there it no need to create `coverage.lcov` file. `nyc report --reporter=text-lcov && codecov` is sufficient.

rgbkrk:
If I make sure my workspace is clean (removing a prior `coverage.lcov`, then I get the coverage text to stdout (expected) followed by this response from `codecov`:

```
==> Detecting CI Provider
    No CI Detected. Using git/mercurial
==> Configuration:
    Endpoint: https://codecov.io
{ commit: '43cd11ae757ffdd837ee2ee0c107ecfa2addb6ce',
  branch: 'master',
  package: 'node-v1.0.1' }
==> Building file structure
==> Generating gcov reports (skip via --disable=gcov)
    $ find . -type f -name '*.gcno'  -exec gcov  {} +
==> Scanning for reports
    X Failed to read file at
```

wuct:
@rgbkrk You are right, I saw the same failure on my laptop.
I have used my setting in [this PR](https://github.com/wuct/raf-throttle/pull/6), and it does trigger the `codecov` correctly. I am not sure why this happened, maybe there is a old coverage report on the Travis CI.
However, we should just use your config. Thanks you again for pointing this out :)

bcoe:
@wuct @rgbkrk it would be nice to be able to simply pipe to `codecov`, rather than having the extra step that pipes to a `.lcov` file.

I'll happily land this documentation, but is this a bug on `codecov`'s end? If so, we should open an issue.

rgbkrk:
I'd sure prefer the pipe too. I'm in no way affiliated with codecov, I just like their design and features better than coveralls (on the site and the chrome extension).

wuct:
IMHO it is not a bug of [`codecov`](https://github.com/codecov/codecov-node). `codecov` uses [`find`](https://github.com/codecov/codecov-node/blob/master/lib/codecov.js#L260) to find coverage reports internally, add does not support pipe currently. However, the deprecating [`codecov.io`](https://github.com/cainus/codecov.io) do support piping. Maybe we should open a feature-request issue on `codecov` and close this issue?

rgbkrk:
@wuct do we want some docs that work for what's available now vs. no docs and waiting on `codecov` to get updates, releases, etc.

wuct:
@rgbkrk I would like to have docs that work for what's available now, because it would help people save time (at least for me) :)

bcoe:
@wuct @rgbkrk would one of you mind opening an issue on codecov? I'll happily accept update docs here if they simplify the process.

rgbkrk:
https://github.com/codecov/codecov-node/issues/13 is still open related to this over on codecov

Thanks for `nyc` @bcoe and for accepting the PR!

PR #237, opened 2016-04-27
Strip any duplicate extensions from --extension

### LINKED ISSUE #236: Duplicate '.js' extension breaks coverage report
Like discussed in #235, adding the `.js` extension (either via command line or package.json) breaks the test coverage report, as it results in duplicate array elements.

I've set up a git repository to reproduce this bug here: https://github.com/doque/break_coverage.
You can switch between the branches to see that it works when removing the '.js' extension (or if you use my fork), see the README.

I realize this error can be attributed to the user, but feel like this should be avoided, which is why I've opened #237  that fixes this bug.

doque:
Fixes #236.

### DIFF
--- a/index.js
+++ b/index.js
@@ -58,6 +58,9 @@ function NYC (opts) {
 
   this.extensions = arrify(config.extension).concat('.js').map(function (ext) {
     return ext.toLowerCase()
+  }).filter(function (item, pos, arr) {
+    // avoid duplicate extensions
+    return arr.indexOf(item) === pos
   })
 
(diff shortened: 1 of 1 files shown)


novemberborn:
LGTM. I like the `indexOf` trick within the `filter`!

I think what's going wrong is that nyc is instrumenting the instrumented code, due to [this line](https://github.com/bcoe/nyc/blob/f320bc6ed8fbacf381ba5cb653711dc85783f0f1/index.js#L272).

PR #251, opened 2016-05-08
Test against more relevant versions of node on Travis

JaKXz:
~~allow for `node@unstable` failures~~
- slightly quicker `npm install`


JaKXz:
Which of [these compilers](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements) would be preferred?

bcoe:
@JaKXz I'm cool with dropping `iojs` and upgrading to 6, but since `nyc` is aiming to be fairly backwards compatible could you leave `node@0.10`.

JaKXz:
@bcoe done. I'll leave the Travis warning about GCC compilers to you / the other maintainers if there's some preferred one necessary - or I can just add whichever you guys want. I don't know the differences well enough to make a substantial decision.

.travis.yml
```diff
@@ -1,11 +1,15 @@
...
 node_js:
-  - '0.12'
-  - '0.10'
-  - iojs
+  - 4
+  - "0.10"
+  - "stable"
+
+before_install:
+  - "npm config set progress=false"
```
novemberborn: Does this do anything? npm won't render a progress bar in Travis anyway.

novemberborn:
Should leave `0.12` as well, it's supported to the end of the year (and even afterwards we should refrain from unintentionally breaking compatibility). I'd prefer testing `5` while it's still being maintained.

.travis.yml
JaKXz: @novemberborn I think it does speed up install a bit, it definitely does locally... I don't think there's harm in having it on CI since like you said Travis doesn't render one anyway, but if there's even a slight benefit then it's worth it imho.

novemberborn: I don't think there are any benefits for Travis (and I doubt whether there's significant ones locally). Best not to have invocations like this in the code, it'll only lead to more questions in the future.

bcoe:
@JaKXz thanks for entertaining our 🚲 🏠, will land this today :)

JaKXz:
@bcoe TBPH I don't mind at all because this is really a discussion for you guys to have as maintainers (about which versions you're supporting etc) - I just noticed that 1) the builds were _really_ slow and 2) that the versions of node were out of date so I thought I'd kickstart things :)

JaKXz:
@novemberborn [this](https://news.ycombinator.com/item?id=10974929) and the continuing discussion at https://github.com/npm/npm/issues/11283 are why I suggest the `before_install` command - I don't mind being vetoed here but I don't think it's harmful.

bcoe:
@JaKXz easy enough to test; let's kick off a build with and without the setting, and look at the Travis benchmarks; there's a second count beside each step in the build.

PR #252, opened 2016-05-08
Update AppVeyor config with ~faster npm install

JaKXz:
Only test against stable and LTS versions of node since AppVeyor goes pretty slow.

@bcoe I noticed that after dropping the Travis build times to 1/4th of what they were on Travis in #251, AppVeyor's build takes even longer... so this is an attempt to streamline it by using the version of `npm` that comes with each node distro and only going for stable and LTS versions. Thoughts?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,12 +1,11 @@
 environment:
   matrix:
-    - nodejs_version: '5'
-    - nodejs_version: '4'
-    - nodejs_version: '0.12'
+    - nodejs_version: '6' # stable
+    - nodejs_version: '4' # LTS
 install:
   - ps: Install-Product node $env:nodejs_version
   - set CI=true
-  - npm -g install npm@latest
+  - npm config set progress=false
   - set PATH=%APPDATA%\npm;%PATH%
   - npm install
 matrix:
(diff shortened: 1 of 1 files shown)


bcoe:
@JaKXz this seems to be failing, kicked off the build twice. Perhaps try adding back the npm update, and see if this helps; if not, we might not be able to pull in 6.x yet.

JaKXz:
@bcoe I don't think the `npm` update is what's broken here, since `node@6` comes with a version that I'm pretty sure `npm install -g npm@latest` would install [at time of writing] - see the AppVeyor logs around line 507:

```
node --version
v6.1.0
npm --version
3.8.6
```

Though I'm not sure why it would pass on Travis and not AppVeyor with `node@6`.

bcoe:
@JaKXz I think we'll need to spin up a Windows VM to test this one out.

jamestalmage:
Probably related to:
https://github.com/sindresorhus/ava/issues/814

JaKXz:
whoa it's passing now... 🎱

jamestalmage:
Yes. Node 6.2.0 reverted a regression that almost certainly affected this.

jamestalmage:
`0.12` remains in LTS mode through the end of the year. We should not drop it yet.

bcoe:
@JaKXz interesting that it started working, maybe the version of `6` was updated.

Issue #239, opened 2016-04-27
Cannot read property 'start' of undefined


xjamundx:
## Reproduction Steps

1. Create an `index.js` file which uses babel-register, destructuring and a binary expression:

```js
require('babel-register')
const {x} = 'hi' || true
```

2. Run `nyc` with html reporter and `ava`:

`./node_modules/.bin/nyc --reporter html ./node_modules/.bin/ava ./index.js `

**Error**
```
/Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/report/html.js:276
                startCol = meta.start.column;
                               ^

TypeError: Cannot read property 'start' of undefined
    at /Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/report/html.js:276:32
    at Array.forEach (native)
    at annotateBranches (/Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/report/html.js:255:30)
    at HtmlReport.Report.mix.writeDetailPage (/Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/report/html.js:426:9)
    at /Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/report/html.js:489:26
    at SyncFileWriter.extend.writeFile (/Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/util/file-writer.js:57:9)
    at FileWriter.extend.writeFile (/Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/util/file-writer.js:147:23)
    at /Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/report/html.js:488:24
    at Array.forEach (native)
    at HtmlReport.Report.mix.writeFiles (/Users/jamuferguson/dev/web/node_modules/nyc/node_modules/istanbul/lib/report/html.js:482:23)
```

You can also try this out in my repo here which is hooked up with this problem:
https://github.com/xjamundx/nyc-broke

novemberborn:
Looks like corrupted coverage data is being output by nyc, and it's tripping up the HTML reporter. If you could share your code, or even better an isolated reproduction, that would help in tracking this down.

xjamundx:
Thanks @novemberborn working on doing just that :)

xjamundx:
Okay this is fun.. It was a lot of code, but now we're down to:

**feedback.js**
```
[...]

PR #255, opened 2016-05-09
Discard more bad source map positions

novemberborn:
Spotted positions like these in coverage output when debugging #239:

```json
            "1": {
                "start": {
                    "line": 1,
                    "column": 0
                },
                "end": {
                    "line": null,
                    "column": -1
                }
            },
```

This PR adds a guard against badly adjusted end positions, causing those locations to drop out of the report.


bcoe:
@novemberborn awesome, good work.

PR #266, opened 2016-06-04
Better ES2015 Support

bcoe:
## The Problem

nyc has several open issues related to ES2015 support:

https://github.com/bcoe/nyc/issues/265
https://github.com/bcoe/nyc/issues/264
https://github.com/bcoe/nyc/issues/215
https://github.com/bcoe/nyc/issues/239
https://github.com/tapjs/node-tap/issues/267

## My Goal

Let's work towards adding better ES2015 support to nyc:

1. using this ticket as a catch-all, and the tests I've checked in as a starting point, let's add regression tests for some of the bad behavior that folks have seen.
2. let's fix the behavior! I think this will require coordination between [Istanbul](https://github.com/gotwarlost/istanbul) and nyc.
  * some of the bad line # mappings might be able to be fixed in our source-map logic (CC: @novemberborn )
  * we might need to address some of the issues in the underlying Istanbul coverage library (CC: @gotwarlost).
  * could we have better native ES2015 support, what does this look like (CC: @JaKXz)
3. let's better document best practices for using nyc with ES2015:
  * I think this should probably be a separate document in a docs folder in nyc. My feeling being that some of the issues folks run into with nyc/ES2015 are actually operator errors.

To get started, I would ask that folks who have run into issues submit pull requests with reproductions to this branch (which will act as a catch all for the time being).

We can build out some failing tests, and hopefully get things to a happier place.

CC: @jmdobry, @skozin, @xjamundx, @remy, @JaKXz

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,7 @@
 {
-  "presets": ["es2015"]
+  "presets": ["es2015"],
+  "plugins": [
+    "transform-async-to-generator",
+    "transform-runtime"
+  ]
 }
--- a/package.json
+++ b/package.json
@@ -89,6 +89,10 @@
   },
   "devDependencies": {
     "any-path": "^1.3.0",
+    "babel-core": "^6.9.1",
+    "babel-plugin-transform-async-to-generator": "^6.8.0",
+    "babel-plugin-transform-runtime": "^6.9.0",
+    "babel-preset-es2015": "^6.9.0",
     "chai": "^3.0.0",
     "coveralls": "^2.11.4",
     "exists-sync": "0.0.3",
--- a/test/fixtures/es2015/asyncawait.js
+++ b/test/fixtures/es2015/asyncawait.js
@@ -1,0 +1,13 @@
+const foo = async (opts, opts2) => {
+  let snuh = await bar()
+}
+
+const bar = () => {
+  return new Promise(function (resolve, reject) {
+    setTimeout(() => {
+      resolve(33)
+    }, 10)
+  })
+}
+
+foo()
--- a/test/fixtures/package.json
+++ b/test/fixtures/package.json
@@ -10,7 +10,8 @@
   "nyc": {
     "exclude": [
       "**/blarg",
-      "**/blerg"
+      "**/blerg",
+      "**/es2015"
     ]
   }
 }
--- a/test/src/es2015.js
+++ b/test/src/es2015.js
@@ -1,0 +1,20 @@
+/* global describe, it */
+
+var NYC
+
+try {
+  NYC = require('../../index.covered.js')
+} catch (e) {
+  NYC = require('../../')
+}
+
+(new NYC()).wrap()
+
+require('babel-core/register')
+require('tap').mochaGlobals()
+
+describe('es2015 coverage', function () {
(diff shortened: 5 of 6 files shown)


jamestalmage:
I tried your sample using `nyc` and `__coverage__`

https://github.com/jamestalmage2/nyc-es2015

`__coverage__` handled your example like a champ.

I think it would be great if we could figure out a way to create pluggable coverage instrumentation. Even when `istanbul` fixes this issue, `__coverage__` is still compelling for people already using `babel-register` to instrument sources (it eliminates the current `print ast from transform 1 => parse code into ast` wasted effort that currently happens).

bcoe:
@jamestalmage really slick, should we just document `__coverage__` as our sanctioned way to handle ES2015, and perhaps add some targeted docs for this?

@dtinth anything we can do to make nyc integrate more tightly with __coverage__?

dtinth:
@bcoe You can use `babel-core` with `babel-plugin-__coverage__` instead of `istanbul.Instrumenter`. Make sure to set `babelrc: false` to make sure that it only does coverage instrumentation.

Here’s an example usage:

```js
const Babel = require('babel-core')
const result = Babel.transform(code, {
  babelrc: false,
  plugins: [ require('babel-plugin-__coverage__') ],
  filename: filename
})
```

The result is an ES6 `{ code, map, ast }` with coverage instructions added.

jamestalmage:
I think it would be better to add first class support for __coverage__ . Setting it up now is pretty convoluted, especially when it comes to include/exclude config.

jamestalmage:
Interesting consequence of naming it `__coverage__`. It's bold if you don't surround with back ticks

bcoe:
@jamestalmage @dtinth this was my thinking too, could we create a shim with __coverage__ that is a drop in replacement for istanbu.Instrumenter, and then make the instrumenter a configuration option in nyc?

jamestalmage:
I think that's a good first step, but I'd love to generalize `nyc` further.

Someday, we are going to allow users to compile their sources in the main process in AVA. We currently do that for tests, and not loading up babel in each thread made a HUGE performance increase. The eventual plan is to analyze the dependency graph, much the same way browserify does, and precompile all the required sources for a test before forking. It should provide a huge speedup.

Once that is done, I would love to integrate a first class `--coverage` flag, much as `node-tap` does, but I would want to do the coverage instrumentation in the main thread using `__coverage__`. My hope is that it will make coverage blazing fast.

Additionally, I think it would be awesome if AVA eventually integrated coverage into it's smart `--watch` mode. Keeping coverage data from tests we decided not to rerun, and combining it with new coverage data from tests we did. That will require tracking which tests are causing which files to be written to disk, and selectively deleting or combining them.

Finally, I think @novemberborn has some interesting ideas with `copenhagen`, and I'm sure there will be other clever coverage instrumentation solutions in the future. It would be nice if `nyc` provided a generalized infrastructure that provided coverage implementors a convenient way to make their solutions handle process forking.

Short term it may be better to go with an easy win that just allows for `__coverage__`.

bcoe:
closing in favor of the work being done in https://github.com/bcoe/nyc/pull/268

PR #268, opened 2016-06-05
feat: make `__coverage__` the default approach we advocate for ES2015 coverage

bcoe:
_related to #266_

The reason that setting `include = /` makes `__coverage__` compatible with `nyc` is that `/` results in no files matching `shouldInstrumentFile()`.

When `nyc` is configured with a rule that does match `shouldInstrumentFile()` it results in the source-map information populated by `__coverage__` being clobbered, breaking output.

Rather than relying on this side-effect I propose that we do the following:

- [x] add a configuration option that allows source-maps to be turned off in `nyc`.
- [x] add a noop instrumenter so that we can skip having `nyc` instrument files (`__coverage__` handles this).
- [x] pull nyc's `shouldInstrumentFile` logic into a module that can be used by both `nyc` and `__coverage__`, this way folks can opt to use `nyc`'s `include`/`exclude` functionality, while using `__coverage__` for instrumentation.
  * https://github.com/bcoe/test-exclude
- [x] update nyc's documentation, documenting the usage of `__coverage__` for ES2015 instrumentation.
- [ ] direct folks in #266 towards using `__coverage__` as a fix for their test-coverage issues.

@jamestalmage @dtinth does this sound reasonable?


novemberborn:
Hey @bcoe, apologies for the radio silence, I've been traveling 😄 

This makes sense, though I wonder if we can shove more behavior into the instrumenter. `nyc`'s default logic could take care of the source maps, for instance. We could also power `--all` and the writing of the coverage data to disk.

That would allow tools like `copenhagen` (if I get a chance to work on it) to plug in to `nyc` more neatly.

bcoe:
@novemberborn happy to gradually pull more logic into sub-modules, being mindful of the logic that will probably be shared across instrumentation libraries, e.g., in the case of `--all`, I imagine the iteration logic would remain fairly consistent, but a replacement instrument would be for asking the actual question.

Could you do me a favor and take `nyc@next` for a spin on a few of your projects, the `__coverage__` logic will not be fully functional until @dtinth lands:

https://github.com/dtinth/babel-plugin-__coverage__/pull/35

PR #276, opened 2016-06-13
fix: upgrade foreground-child dependency

bcoe:
@addaleax @jamestalmage mind giving this a spin while builds run?

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "default-require-extensions": "^1.0.0",
     "find-cache-dir": "^0.1.1",
     "find-up": "^1.1.2",
-    "foreground-child": "^1.4.0",
+    "foreground-child": "^1.5.1",
     "glob": "^7.0.3",
     "istanbul": "^0.4.3",
     "md5-hex": "^1.2.0",


jamestalmage:
`spawn-wrap` is still on `foreground-child@1.3`

addaleax:
Looks good locally and I think bundling an npm3-style tree should take care of the old version that `spawn-wrap` depends on, right?

bcoe:
@jamestalmage @addaleax my plan is to bundle on a npm@2.x tree, which solves some singleton issues with our dependencies -- but! I think we should be fine as long as I re-install everything prior to publish? `spawn-wrap` isn't itself bundled or pinned.

addaleax:
I guess that should work. If the copy of `foreground-child` in `spawn-wrap`’s dependencies is still outdated after that for some reason, you can probably also just remove it and let `spawn-wrap` use the top-level one.

jamestalmage:
Yeah - actually, it's a carrot dependency, so it should be deduped. 

Why does bundling on `npm@2` solve problems?

bcoe:
@jamestalmage nyc relies on a few popular dependencies that themselves use a singleton pattern (for various reasons, maintaining state, etc).

Bundling dependencies helped to solve this issue, but I found that deduping dependencies and publishing with npm@3.x continued to introduce some regressions:

https://github.com/bcoe/nyc/issues/246

jamestalmage:
Hmm.

Have we figured out **why** that is happening? Is there a bug somewhere in `npm`? Ideally we wouldn't be so brittle.

PR #278, opened 2016-06-14
feat: add "instrument" option as alternative to specifying noop instrumenter

bcoe:
it felt a little weird providing a noop instrumenter directly in `__coverage__`, this pull introduces the `instrument` flag instead -- making it possible to setup `__coverage__` with the following config:

```json
 "nyc": {
    "include": [
      "index.js"
    ],
    "require": [
      "babel-register"
    ],
    "reporter": [
      "lcov",
      "text"
    ],
    "sourceMap": false,
    "instrument": false
  }
```

which feels cleaner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ the `sourceMaps` option to `inline`.
         "babel-register"
       ],
       "sourceMap": false,
-      "instrumenter": "./lib/instrumenters/noop"
+      "instrument": false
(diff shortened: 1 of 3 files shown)


bcoe:
@dtinth added the `--instrument` flag which, in my opinion, makes configuring `__coverage__` feel less weird.

PR #280, opened 2016-06-15
Docs: Only enable __coverage__ plugin in test env.

dtinth:
I updated the docs, so that it prevents the `__coverage__` plugin from kicking in in development/production mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,10 +70,16 @@ the `sourceMaps` option to `inline`.
     {
       "babel": {
         "presets": ["es2015"],
-        "plugins": ["__coverage__"]
+        "env": {
+          "test": {
+            "plugins": ["__coverage__"]
+          }
+        }
       }
     }
   ```
+
+  Note: With this configuration, the `__coverage__` will only be active when `NODE_ENV` or `BABEL_ENV` is `test`.
 
 2. disable nyc's instrumentation and source-maps:
 


bcoe:
@dtinth awesome can't wait to write a blog post on this!

bcoe:
@dtinth might it be worth drawing attention to [cross-env](https://www.npmjs.com/package/cross-env) -- good way to set the test environment in npm scripts.

jamestalmage:
I like using `development` instead of `test`. Babel assumes `development` if not set explicitly.

Then set it to `production` for deploys:

```json
{
  "scripts": {
    "build": "babel --out-dir lib src",
    "prepublish": "NODE_ENV=production npm run build"
  }
}
```

That seems to be what [babel suggests](http://babeljs.io/docs/usage/babelrc/).

bcoe:
@jamestalmage, if @dtinth is comfortable with this, I think this might be a good modification to the configuration, folks can avoid setting the environment variable.

kentcdodds:
> I like using development instead of test. Babel assumes development if not set explicitly.

I'd recommend against this. Here's how I see it used (and do myself):

- `development` is for running the app locally (no minification etc.)
- `test` is for running the tests
- `production` is for building/deploying

If you mix `development` and/or `test`, then you could wind up in a situation where you're running the app locally with coverage instrumentation. Probably not a good idea :-)

bcoe:
@kentcdodds @jamestalmage @dtinth perhaps compromise, and add a note about `cross-env` like I originally proposed?

kentcdodds:
Yes, that's great

jamestalmage:
What @kentcdodds says makes sense to me. I say we just stick with `test`, then. 

And, yes of course - mention `cross-env`

kentcdodds:
PR created: #281

PR #298, opened 2016-07-06
feat: adds instrument command line option

### LINKED ISSUE #293: add `instrument` command line option
We should add a command line option that allow an arbitrary file to be instrumented and returns the output. @gotwarlost it sounded like this is one of the glaring features missing in nyc, that's required for front-end Istanbul users?

bcoe:
first pass at adding `instrument` command, which can be used to instrument files, or directories of files.

@gotwarlost could you give this a spin and let me know if it meets your needs (@jamestalmage would also love a code review).

fixes #293


bin/nyc.js
```diff
@@ -16,23 +16,47 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
   .command('report', 'run coverage report for .nyc_output', function (yargs) {
     return yargs
       .usage('$0 report [options]')
+      .option('reporter', {
+        alias: 'r',
+        describe: 'coverage reporter(s) to use',
+        default: 'text'
+      })
+      .option('report-dir', {
+        describe: 'default directory to output coverage reports in',
```
kentcdodds: Shouldn't this simply say:

```
directory to output coverage reports in
```

The default is `coverage`. If `report-dir` is specified, it's no longer the "default".

bin/nyc.js
```diff
@@ -16,23 +16,47 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
...
+      .option('branches', {
+        default: 0,
+        description: 'what % of branches must be covered?'
+      })
+      .option('functions', {
+        default: 0,
+        description: 'what % of functions must be covered?'
+      })
+      .option('lines', {
+        default: 90,
```
kentcdodds: Interesting. Is there a reason this is `90` instead of `0`?

bcoe: just picking defaults based on personal preference, I tend to find a project with `90%+` test coverage is pretty healthy.

kentcdodds: I see. Maybe I'm misunderstanding what these numbers mean, but shouldn't that be true of the other thresholds as well?

bcoe: I think you're right, just not as experienced with the other thresholds.

kentcdodds: Yeah, me either. I only think that I sort of understand what they really mean, but I'm not certain...

PR #301, opened 2016-07-06
Avoid pid collisions.

jamestalmage:
With AVA's concurrency flag, it's possible that two processes during a test run will share the same pid.

Avoid this collision by hashing the pid with process.hrtime for a truly unique value.

### DIFF
--- a/index.js
+++ b/index.js
@@ -299,8 +299,12 @@ NYC.prototype.writeCoverageFile = function () {
     coverage = this.sourceMapTransform(coverage)
   }
 
+  var id = md5hex(
+    process.hrtime().concat(process.pid).map(String)
+  )
+
   fs.writeFileSync(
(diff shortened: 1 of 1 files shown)


jamestalmage:
This is almost certainly an issue on Windows. 

Proof:

```js
var childProcess = require('child_process');
var path = require('path');
var p = path.join(__dirname, 'noop.js');

var count = 0;
var pids = {};

function fork() {
  if (count > 200) {
    return;
  }

  count ++;

  var id = childProcess.fork(p).on('exit', fork).pid;

  if (pids[id]) {
    console.log('duplicate use of ' + id);
  }

  pids[id] = true;
}

fork();
```

This prints nothing on Linux, and 50+ lines Windows.

bcoe:
@jamestalmage sounds like a plan to me.

PR #307, opened 2016-07-10
fix: babylon misconfiguration

kentcdodds:
When using ES6 modules in source/test code, I get this message: 

```
failed to instrument /Users/kdodds/Developer/starwars-names/src/index.js with error: 'import' and 'export' may appear only with 'sourceType: module' (1:0)
```

The coverage report is still generated, but I think that we probably don't want this error message. Let me know if I'm misunderstanding something.

### DIFF
--- a/lib/instrumenters/istanbul.js
+++ b/lib/instrumenters/istanbul.js
@@ -3,6 +3,7 @@ function InstrumenterIstanbul (cwd) {
 
   return istanbul.createInstrumenter({
     autoWrap: true,
+    esModules: true,
     coverageVariable: '__coverage__',
     embedSource: true,
     noCompact: false,


gotwarlost:
Is there a situation where this might need to be `false` and, if so, should this be a config option?

kentcdodds:
@gotwarlost, I figured that'd be needed. Thought I'd open the PR with minimal requirements and we can iterate. Any suggestions from anyone on @istanbuljs/nyc about where/how this should work? What should be the default?

JaKXz:
It would be `false` when when instrumenting ES5 code. When instrumenting ES2015+ modules I've been using the `babel-plugin-istanbul` with nyc's instrumentation turned off as specified in the readme without too many issues.

kentcdodds:
Huh... Somehow I missed that requirement of the `7.0.0` breaking changes... This makes me sad. `nyc` went from: "just stick `nyc` before `npm test`" to a whole bunch of config. Oh well...

bcoe:
@kentcdodds for native ES6 stuff (whatever is currently supported on the Node bleeding edge), It should continue being `nyc npm test`. Using `babel-plugin-instanbul` should only be necessary when you're using a compile step to pull in ES2015 features outside what's currently supported by Node.

kentcdodds:
Understood. There are definitely several use cases for transpiling with babel. Almost everything I do uses babel for several features (even when Node6 is available). No worries though. I still love `nyc`.

bcoe:
@kentcdodds I'd love to start a conversation about how we could make things work out of the box. also maybe it's worth adding some documentation that points out that ES6 _can work_ if you're using newer V8 builds...

It seems like we're experiencing a cambrian explosion of approaches to instrumenting ES2015 currently, so I do want to be careful that nyc isn't too prescriptive ... I also want to keep things simple. feels like we need to find a balance.

kentcdodds:
Yeah, honestly, with how powerful and flexible our tools are, it's important that we avoid being prescriptive. I was going to suggest that we have a special binary that sets some defaults (`nyc-babel` for example). That may not be a bad idea. We'd cover a bunch of use cases that way I think...

Issue #171, opened 2016-02-25
Add the ability to specify a custom tempDirectory


kentcdodds:
The `NYC` constructor already accepts the option. But this needs to be exposed and utilized in the CLI. I almost had things working, but then had to get back to work.

My use case is I have frontend and backend code that's tested separately and I want to be able to track coverage on them separately. So I already specify the `--report-dir` for them, but to be able to utilize `check-coverage`, I need to be able to specify a custom `tempDirectory`.

I had to slightly alter both the `bin/nyc.js` and `bin/wrap.js`. Basically this got me most of the way there:

In **nyc.js**

``` diff
if (argv._[0] === 'report') {
  // run a report.
  process.env.NYC_CWD = process.cwd()

  report(argv)
} else if (argv._[0] === 'check-coverage') {
  checkCoverage(argv)
} else if (argv._.length) {
  // wrap subprocesses and execute argv[1]
  if (!Array.isArray(argv.require)) argv.require = [argv.require]

  var nyc = (new NYC({
    require: argv.require,
+    tempDirectory: argv.tempDirectory,
  }))
  nyc.reset()

  if (argv.all) nyc.addAllFiles()

  var env = {
    NYC_CWD: process.cwd(),
    NYC_CACHE: argv.cache ? 'enable' : 'disable',
+    NYC_TEMP_DIRECTORY: argv.tempDirectory,
  }
  if (!argv.tempDirectory) {
      throw new Error('no temp directory in nyc')
  }
  if (argv.require.length) {
    env.NYC_REQUIRE = argv.require.join(',')
  }
  sw([wrapper], env)

  foreground(nyc.mungeArgs(argv), function (done) {
    if (argv.checkCoverage) {
      checkCoverage(argv, function (done) {
        if (!argv.silent) report(argv)
        return done()
      })
    } else {
      if (!argv.silent) report(argv)
      return done()
    }
  })
} else {
  // I don't have a clue what you're doing.
  yargs.showHelp()
}

function report (argv) {
  process.env.NYC_CWD = process.cwd()

  ;(new NYC({
    reporter: argv.reporter,
    reportDir: argv.reportDir,
+    tempDirectory: argv.tempDirectory,
  })).report()
}


function checkCoverage (argv, cb) {
+  var tempDirectory = argv.tempDirectory || './.nyc_output'
  foreground(
    process.execPath,
    [
      require.resolve('istanbul/lib/cli'),
      'check-coverage',
      '--lines=' + argv.lines,
[...]

Issue #292, opened 2016-07-02
combine functionality of append-transform and istanbul-lib-hook


bcoe:
we should figure out how we want to combine the functionality of these two libraries. Currently nyc lacks functionality surrounding the vm module.

_any work we put towards this should be against this branch: https://github.com/istanbuljs/nyc/pull/286_

CC: @gotwarlost, @novemberborn, @jamestalmage

gotwarlost:
I'll take a crack at this. If I get stuck I'll ping you guys for help :)

bcoe:
@gotwarlost awesome, let me know anything I can do to help -- also be liberal with opening additional tickets. I'd love to get this to the point where we're comfortable with pointing the existing Istanbul user-base at the nyc bin.

gotwarlost:
One question: how is the `addFile` method used? Is this part of the public API for `nyc`? No one calls this in `nyc` itself apart from `addAllFiles`

bcoe:
@gotwarlost looks like you're correct. we should rename this `_addFile` I think, and move it so it's nested directly below the method `addAllFiles`? -- make it stand out more as a private helper function, rather than an API method... I think we could do pass through NYC and give several methods this treatment.

bcoe:
@gotwarlost this is out the door \o/

PR #308, opened 2016-07-10
Use istanbul-lib-hook to wrap require and support vm hooks

gotwarlost:
### Interface changes
* Add `--hook-run-in-context` (default: `true`) to the `nyc` command line
* Add `--temp-directory <dir>` to the `report` command line. This is **not** a full implementation of a customizable `tmpdir` but only the bare minimum required to support self-coverage reporting for `nyc`. The full implementation is described in #171

### Misc changes
* Remove `istanbul` dev dependency in favor of APIs that are already being used, so that we don't accidentally use code from the old module any more. (I think the `text` reporter was being pulled in from the old lib). 
* Coverage reporting is now done by `nyc` itself at the end of tests and this necessitates the custom `temp-directory` option for reporting.

### Implementation
* Use istanbul-lib-hook to wrap `require` and hook `vm.runInThisContext`
* Add  `--hook-run-in-context` option and propagate to subprocesses using the `NYC_HOOK_RUN_IN_CONTEXT` environment var
* Add a test that will succeed only if `runInThisContext` is successfully hooked. 
* Replace `istanbul` API use with `istanbul-lib-instrument` and command line use with `nyc` itself.


gotwarlost:
Some questions:

* I added `requirejs` as a dev dependency although it is actually a dependency of the test project. Is this correct or should we be doing an `npm install` from the test project directory before running its test?
* there is no equivalent test for these hooks apart from the CLI test. Is this needed?

Link main issue for cross-ref: https://github.com/istanbuljs/nyc/issues/292

gotwarlost:
No idea why the appveyor build is failing. Need help.

bcoe:
@gotwarlost AppVeyor just fails a good chunk of the time, nothing to do with this pull as long as I can tell. re-kicked it off.

bin/nyc.js
```diff
@@ -127,6 +131,16 @@ var yargs = require('yargs/yargs')(process.argv.slice(2))
     type: 'boolean',
     description: 'should nyc handle instrumentation?'
   })
+  .option('hook-run-in-context', {
+    default: false,
+    type: 'boolean',
+    description: 'should nyc wrap vm.runInThisContext?'
+  })
+  .option('hook-create-script', {
+    default: false,
```
bcoe: we've tried to be fairly batteries included so far, for instance we already instrument subprocesses right out of the gate.  I might be tempted to default this to `true`. I'm happy with the argument name 👍

bin/wrap.js
```diff
@@ -13,7 +13,9 @@ try {
   include: process.env.NYC_INCLUDE ? process.env.NYC_INCLUDE.split(',') : [],
   enableCache: process.env.NYC_CACHE === 'enable',
   sourceMap: process.env.NYC_SOURCE_MAP === 'enable',
-  instrumenter: process.env.NYC_INSTRUMENTER
+  instrumenter: process.env.NYC_INSTRUMENTER,
+  hookRunInContext: process.env.NYC_HOOK_RUN_IN_CONTEXT === 'true',
```
bcoe: For consistency, I might stick with setting the environment variable to `enable`; I think this approach originated from not wanting to get confused by variable type (`!!'false' === true`).

build-self-coverage.js
```diff
@@ -1,4 +1,4 @@
-var istanbul = require('istanbul')
+var istanbul = require('istanbul-lib-instrument')
```
bcoe: I really like this fix. it felt weird leaving in the Istanbul dependency.

test/src/nyc-bin.js
```diff
@@ -388,4 +389,25 @@ describe('the nyc cli', function () {
       })
     })
   })
+
+  describe('hooks', function () {
+    it('provides coverage for requireJS and AMD modules', function (done) {
+      var args = [bin, '--hook-run-in-context', process.execPath, './index.js']
```
bcoe: lovely! I'm definitely finding myself writing more and more tests that directly exercise the bin which I don't think is a bad thing -- for one catches regressions I've introduced in the actual command line parsing.

bcoe:
This is looking great 🎉 to land this I think all we'd need to to do would be:

* add a test for `vm.createScript`, I don't think I saw one?
* address the couple small issues:
  * I think that enabling the new instrumentation settings by default would be smart (@jamestalmage what are your thoughts?).
  * I'd be tempted to use `'enable'` rather than `'true'` when passing configuration around in the `env`.
* _finally_, I'd also like to spin up my Windows VM and run this against a few projects once we're happy.

gotwarlost:
Thanks for the detailed feedback.

* I think I'm going to remove the `vm.createScript` hook. The problem is that there is no place to stash the original code that was used for `createScript`. This is because, the instrumenter no longer supports the notion of embedding the source as part of the file coverage object. So if we ended up hooking `vm.createScript` the HTML report won't be able to find the original source. If someone really, really needs this we can add back the notion of embedding the source into the instrumenter and then implementing this. Even in `istanbul` this is a fairly exotic case and I don't believe it works correctly even today.
* I will change the setting to `enable|disable` instead of `true|false` and enable it by default

bcoe:
@gotwarlost awesome, are you feeling like we're getting close to being able to point existing Istanbul users towards the Istanbul 1.0 APIs, and nyc? I've been having great luck as I test the new libraries against everything I throw at them, awesome work :) I wish we'd kicked this in motion s

gotwarlost:
@bcoe I think we are getting real close, it's probably already much better than `istanbul`. One problem is that it is not backwards compatible and we need to go through a documentation exercise and more importantly, provide a conversion program that will take a `istanbul.yaml` and turn it into a `nyc` config file, showing warnings with unsupported features etc.

I think a big gap from `istanbul.yaml` is the inability to configure reporting details in nyc (e.g. set the `text` reporter to write to a file, for example), but those are minority use-cases.

gotwarlost:
OK, the latest commit has all the PR comments included in it. Please take a look. I can squash commits after things look ok. Specifically, please check bundle dependency changes in `package.json` - I'm not quite sure abut what I did there.

gotwarlost:
Also, I'm not sure if `coveralls` not being able to tell how coverage changed is a regression in the way lcov info is being fed to it (lcov now generated using nyc instead of istanbul)

bcoe:
@gotwarlost this is feeling like a coveralls.io issue, I find there's a lot of weirdness with various services when I transfer repo ownership.

PR #318, opened 2016-07-16
docs(README): Add more complete configuration example [ci skip]

JaKXz:
also fix a typo :)


bcoe:
@JaKXz thanks for the contribution \o/ 

going forward, let's start thinking about how we can chop this documentation into multiple knowledge-base articles here:

https://github.com/istanbuljs/istanbuljs.github.io/issues

PR #321, opened 2016-07-17
Handles correct path casing. Fixes #269.

### LINKED ISSUE #269: Files reported twice with different case
I created a yeoman generator that creates Typescript projects with support for `nyc` and `ava` out of the box. Below is the output I am receiving when trying to get the coverage results (run `npm test` or `gulp coverage`):

image

The same thing also happens when opening an lcov report in the browser. See [this project as the current example](https://github.com/JimmyBoh/bio-sim). It is currently reporting a lowercase `c:\` path and an uppercase `C:\` path.

You can install the template with `npm i -g generator-jb-module`.

JimmyBoh:
A simple fix that forces the provided absolute path to use `this.cwd` instead. This is a windows only fix, so let me know if it should be wrapped in a `if` statement for platform checks.

### DIFF
--- a/index.js
+++ b/index.js
@@ -287,6 +287,7 @@ NYC.prototype._handleSourceMap = function (cacheDir, code, hash, filename) {
 
 NYC.prototype._handleJs = function (code, filename) {
   var relFile = path.relative(this.cwd, filename)
+  filename = path.resolve(this.cwd, relFile) // ensure the path has correct casing (see https://github.com/istanbuljs/nyc/issues/269)
   return this._maybeInstrumentSource(code, filename, relFile) || code
 }
 
--- a/index.js
+++ b/index.js
@@ -287,7 +287,8 @@ NYC.prototype._handleSourceMap = function (cacheDir, code, hash, filename) {
 
 NYC.prototype._handleJs = function (code, filename) {
   var relFile = path.relative(this.cwd, filename)
-  filename = path.resolve(this.cwd, relFile) // ensure the path has correct casing (see https://github.com/istanbuljs/nyc/issues/269)
+  // ensure the path has correct casing (see istanbuljs/nyc#269 and nodejs/node#6624)
+  filename = path.resolve(this.cwd, relFile)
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -287,6 +287,7 @@ NYC.prototype._handleSourceMap = function (cacheDir, code, hash, filename) {
 
 NYC.prototype._handleJs = function (code, filename) {
   var relFile = path.relative(this.cwd, filename)
+  filename = path.resolve(this.cwd, relFile) // ensure the path has correct casing (see https://github.com/istanbuljs/nyc/issues/269)
```
addaleax: I’m not sure whether there are some hard rules around line lengths around here, but I’d recommend placing the comment on its own line. ;)

JimmyBoh: No problem, I can make that change. What are your thoughts on the "re-resolve"?

bcoe:
this _seems_ like a reasonable fix; I seem to screw up path resolution logic `9/10` times, so appreciate the extra set of eyes @addaleax, @gotwarlost.

index.js
addaleax: > What are your thoughts on the "re-resolve"?

tbh, they are kind limited to “this is not obviously wrong” and “the CI seems to be happy”, it’s not really a part of the codebase I have worked with a lot. :D

bcoe: @JimmyBoh I can spin up my `ievm` and test this against a few different OS.

bcoe: @JimmyBoh I'd also be tempted to add a reference to the underlying issue in node.js, so that we can retire the filename logic in the future.

bcoe:
@JimmyBoh could you give this  a spin:

I've landed your path fixes.

JimmyBoh:
Absolutely! I will have some time later today (approx. 4 hours from now). Thanks!

bcoe:
@JimmyBoh I've actually published this to `latest` now:

`npm cache clear; npm i nyc`

Let me know if you run into any issues.

JimmyBoh:
Alrighty, will do!

Issue #99, opened 2015-12-16
use `--` to prevent `nyc` from processing further args.


jamestalmage:
Trying to use latest (master) `nyc` with a custom tap reporter:

```
$ nyc tap --reporter=spec
```

Things work fine until the end, when `nyc` tries to find a `spec` reporter.

The typical solution is to stop processing after a `--` delimiter is found

```
$ nyc tap -- --reporter=spec
```

Issue #201, opened 2016-03-23
--silent (and probably all option) are read when they should not ?


MoOx:
```
  "test-with-coverage": "nyc npm -s test",
```

I want npm to be silent, not nyc. Minor issue but wanted to report it :)
Maybe option should be first?

danez:
I have the same issue, it also happens when specifying `--reporter xunit` for mocha. nyc thinks its its reporter. 
What seems to work for us is this:

```
nyc --reporter html -- npm -s test -- --reporter xunit
```

JaKXz:
@danez that might actually be expected behaviour, because the `--` delimiter is what lets you pass arguments to `npm run` scripts. 

@MoOx could you try [belatedly]:

```json
"test-with-coverage": "nyc npm run test -- -s"
```

danez:
I think this would supply `-s`to whatever is called in `npm test`

I think it should be:
```
"test-with-coverage": "nyc -- npm -s test"
```

JaKXz:
@danez passing `-s` to the `npm test` command is what I was trying to go for. The `--` delimiter is not recognized by nyc, and I don't think that line will work. It's meant for passing [different arguments to existing npm run scripts](https://docs.npmjs.com/cli/run-script).

bcoe:
@MoOx @JaKXz I think we should definitely make nyc's parsing a bit more clever, so that it doesn't look at flags once after the bin it's instrumenting:

`nyc -f banana npm -s` should not parse `-s`, we just need to get a bit more clever with our parsing algorithm.

bcoe:
@MoOx @danez cleaning up some old issues; I believe this should have been addressed when we pulled in `yargs-parser` and started using it to pass arguments to nyc and the instrumented bin.

PR #322, opened 2016-07-17
feat: nyc no longer tries to run arguments passed to the instrumented bin

bcoe:
This pull request is an attempt to make nyc's argument parsing more clever, in the past:

`nyc npm test --silent`

would have resulted in `nyc` itself reading the `--silent` flag, and failing to print output.

With this fix, nyc instead parses:

`nyc npm test`, dropping the trailing arguments.

fixes: #99, #201 

Reviewers: @MoOx, @jamestalmage, @danez, @JaKXz

### DIFF
--- a/test/src/nyc-bin.js
+++ b/test/src/nyc-bin.js
@@ -410,4 +410,29 @@ describe('the nyc cli', function () {
       })
     })
   })
+
+  describe('args', function () {
+    it('does not interpret args intended for instrumented bin', function (done) {
+      var args = [bin, '--silent', process.execPath, 'args.js', '--help', '--version']
+
+      var proc = spawn(process.execPath, args, {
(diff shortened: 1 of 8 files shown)


test/src/nyc-bin.js
```diff
@@ -410,4 +410,29 @@ describe('the nyc cli', function () {
...
+      proc.stdout.on('data', function (chunk) {
+        stdout += chunk
+      })
+
+      proc.on('close', function (code) {
+        code.should.equal(0)
+        var args = JSON.parse(stdout)
+        args.should.include('--help')
+        args.should.include('--version')
+        args.should.not.include('--silent')
```
MoOx: lgtm!

MoOx:
Nice work! Thank you for this!

PR #326, opened 2016-07-22
Make --all flag work with files with extensions other than .js

rundef:
I'm using TypeScript for a project, and when I run nyc with the *--all* flag, I keep getting errors such as:

*failed to instrument /microservices/ms-gateway/app/entities/UserEndpoint.ts with error: 'import' and 'export' may appear only with 'sourceType: module' (1:0)*

My package.json config looks like this:

```json
"nyc": {
    "include": [
      "app/*.ts",
      "app/**/*.ts"
    ],
    "extension": [
      ".ts"
    ],
    "require": [
      "ts-node/register"
    ]
  }
```
It happens because under the hood, the code was trying to instrument the TypeScript code rather than the transpiled JavaScript code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -146,7 +146,7 @@ NYC.prototype.addFile = function (filename) {
 NYC.prototype._readTranspiledSource = function (filePath) {
   var source = null
   var ext = path.extname(filePath)
-  if (Module._extensions[ext] === 'undefined') {
+  if (typeof Module._extensions[ext] === 'undefined') {
     ext = '.js'
   }
   Module._extensions[ext]({
--- a/index.js
+++ b/index.js
@@ -143,13 +143,14 @@ NYC.prototype.addFile = function (filename) {
   }
 }
 
-NYC.prototype._readTranspiledSource = function (path) {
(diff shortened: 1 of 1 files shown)


rundef:
Sorry about all the commits, the tests weren't working here - for some reason, the test 027 doesn't pass on my computer

bcoe:
@rundef great catch; 

I'm curious why your local tests are failing! might be worth digging into more.

I'll be putting out some releases tonight and tomorrow, and will make sure to land this.

rundef:
The test was failing because I didn't have write access to **node_modules/.cache**
Now everything is working fine !

bcoe:
@rundef I've also landed this change in `next`, if you wouldn't mind giving it a shot:

when you have a moment, and are bored, would love a follow up unit test at some point -- just so we never have a regression of this feature.

bcoe:
@rundef I've actually promoted this to `latest`.

`npm cache clear; npm i nyc` should have your changes.

rundef:
Thanks, it works perfectly for me now.
I will send a unit test for that feature later this week, when I have some free time.

PR #335, opened 2016-07-26
Add unit test: addAllFiles should work correctly with non-.js files

rundef:
Unit test for the PR #326


bcoe:
@rundef give the following a try:

`npm cache clear; npm i nyc@next`, I've released your changes.

I'd also love the second set of eyes on the `include`/`exclude` logic, as I've released some breaking changes to how this is handled.

PR #343, opened 2016-08-09
Make `nyc instrument` work in subdirectories

addaleax:
Hi everyone! Ran into a problem running `nyc instrument` today, and it would be cool if you could take a look at. I’m pretty sure this is a correct fix, but still, removing code always comes with the suspicion that there is a reason it was there in the first place. ;)

------

There is no real reason to use a relative path when performing the I/O for instrumenting. Absolute paths should work just as well and will be correct, even when `nyc.cwd` is off a bit because it is modified based on what `pkgUp` says.

### DIFF
--- a/index.js
+++ b/index.js
@@ -183,7 +183,7 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
   var visitor = function (filename) {
     var ext
     var transform
-    var inFile = path.relative(_this.cwd, path.resolve(inputDir, filename))
+    var inFile = path.resolve(inputDir, filename)
     var code = fs.readFileSync(inFile, 'utf-8')
 
     for (ext in _this.transforms) {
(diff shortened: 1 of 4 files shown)


bcoe:
@addaleax this patch looks quite reasonable to me; I think this path issue was probably more related to me always getting tripped up by the paradox of choice in the `fs` API, rather than a specific issue I was trying to solve.

addaleax:
@bcoe Ok, cool then! Everything’s green, I guess there’s no reason not to merge this?

bcoe:
@addaleax 👍 might take me a couple days to get a release out, I want to carefully land a breaking `test-exclude` API change.

addaleax:
Sure, I guess I’ll just keep an eye on `nyc@next` this week? :)

bcoe:
@addaleax give `nyx@next` a spin, I've released your changes; I'd also love the extra set of eyes on the `exclude`/`include` logic, as I've introduced a breaking change.

PR #348, opened 2016-08-12
feat: upgrade to new test-exclude; with suppport for node_modules

bcoe:
BREAKING CHANGE: node_modules is no longer automatically excluded, and an empty array of exclude rules can be provided.

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -10,6 +10,7 @@ try {
 var processArgs = require('../lib/process-args')
 
 var sw = require('spawn-wrap')
+var testExclude = require('test-exclude')
 var wrapper = require.resolve('./wrap.js')
 var Yargs = require('yargs/yargs')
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
> node_modules is no longer automatically excluded

Why? This is very inconvenient and different from most other tools (AVA, ESLint, etc).

bcoe:
@sindresorhus `node_modules` is still excluded by the default set of rules, but if you opt to provide your own exclude rules, you must provide it yourself.

Several folks were asking to be able to instrument files in their `node_modules` folder, which was impossible with the old API, so I thought this was an okay compromise.

sindresorhus:
@bcoe That's what negated globs are for: `["!node_modules"]` should have re-enabled `node_modules`.

bcoe:
@sindresorhus that's smart, I would definitely be comfortable with moving towards this.

sindresorhus:
Where should I open an issue? Here or `test-exclude`?

bcoe:
@sindresorhus 👍 yeah, seems reasonable; this would be another breaking change; but worthwhile you think? covering `node_modules` is an edge-case, and we can help those people one by one.

sindresorhus:
Yeah, I think it's worth it. Very few would cover `node_modules`, but many would define some excludes, which means they would need to needlessly include `node_modules`.

bcoe:
@sindresorhus sorry, only half answered your question; let's open an issue on `test-exclude`.

PR #352, opened 2016-08-14
docs(README): `node_modules` is no longer excluded

gr2m:
via [v8.0.0](https://github.com/istanbuljs/nyc/blob/master/CHANGELOG.md#800-2016-08-12)

Please let me know if I understood this correctly :)

And if you don’t mind me asking here, for a setup like this one:
https://github.com/hoodiehq/spawn-pouchdb-server/blob/master/package.json#L6

Should I know exclude `node_modules`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,9 +170,6 @@ the array is a glob pattern indicating which paths should be omitted.
 
 Globs are matched using [micromatch](https://www.npmjs.com/package/micromatch).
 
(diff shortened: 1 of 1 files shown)


bcoe:
@gr2m you shouldn't need to add exclude rules, `node_modules` is still omitted in the default group of globs; it's just folks writing custom `exclude` logic that will need to update.

gr2m:
Gotcha, thanks Ben!

Issue #158, opened 2016-01-31
Idea: track command, args, cwd, env of each process


isaacs:
Nyc right now creates a really interesting record of not just what lines and files were covered, but also what lines and files were covered *by each individual process*, organized by pid.

I was thinking it might be interesting to have a way to drill into the details of each of those commands.

It might be worth putting off until someone has an idea for how to report on this, but even just for forensics or saving coverage reports for later analysis, it could be interesting.  And it's probably relatively trivial to just dump some process info into the json along with the coverage data.  If it's problematic to put it in the same file, it could be dropped into a file at `.nyc_coverage/processinfo/<pid>.json` containing:

```javascript
{
  "argv": process.argv,
  "execArgv": process.execArgv,
  "env": process.env,
  "cwd": process.cwd()
}
```

jamestalmage:
You could organize/combine hierarchically as well (what are all the pid's that pid X launched)

PR #364, opened 2016-08-24
feat: gather process tree information

addaleax:
Add a `--show-process-tree` that shows a pretty tree of all spawned processes after `nyc` has run.

The data files for that are stored in `processinfo` are stored in `(temp directory)/processinfo` so that they don’t interfere with the fixed format of the coverage files.

Fixes: https://github.com/istanbuljs/nyc/issues/158

Right now this only displays the commands as they are being run, ex:

```
$ nyc --show-process-tree npm run pretest

> nyc@8.1.0 pretest /home/sqrt/src/nyc
> standard

----------|----------|----------|----------|----------|----------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------|----------|----------|----------|----------|----------------|
All files |  Unknown |  Unknown |  Unknown |  Unknown |                |
----------|----------|----------|----------|----------|----------------|
nyc
└─┬ /usr/local/bin/node /usr/local/bin/npm run pretest
  └── /usr/local/bin/node /home/xxxx/src/nyc/node_modules/.bin/standard
```

Note: Unlike described in the above issue, I’ve opted to leave environment variables out of the stored data because they may contain sensitive information and there have been cases of `.nyc_output` accidentally being published.

Any thoughts?

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,13 @@ var pkgUp = require('pkg-up')
 var testExclude = require('test-exclude')
 var yargs = require('yargs')
 
+var ProcessInfo
+try {
+  ProcessInfo = require('./lib/process.covered.js')
+} catch (e) {
+  ProcessInfo = require('./lib/process.js')
+}
+
 /* istanbul ignore next */
 if (/index\.covered\.js$/.test(__filename)) {
   require('./lib/self-coverage-helper')
@@ -36,6 +43,7 @@ function NYC (opts) {
   this._instrumenterLib = require(config.instrumenter || './lib/instrumenters/istanbul')
   this._reportDir = config.reportDir
   this._sourceMap = config.sourceMap
+  this._showProcessTree = config.showProcessTree
   this.cwd = config.cwd
 
   this.reporter = arrify(config.reporter || 'text')
@@ -71,6 +79,9 @@ function NYC (opts) {
   this.hashCache = {}
   this.loadedMaps = null
   this.fakeRequire = null
+
+  this.processInfo = new ProcessInfo(opts && opts._processInfo)
+  this.rootId = this.processInfo.root || this.generateUniqueID()
 }
 
 NYC.prototype._loadConfig = function (opts) {
@@ -327,6 +338,10 @@ NYC.prototype.clearCache = function () {
 
 NYC.prototype.createTempDirectory = function () {
   mkdirp.sync(this.tempDirectory())
+
+  if (this._showProcessTree) {
+    mkdirp.sync(this.processInfoDirectory())
+  }
 }
 
 NYC.prototype.reset = function () {
@@ -352,6 +367,12 @@ NYC.prototype.wrap = function (bin) {
   return this
 }
 
+NYC.prototype.generateUniqueID = function () {
+  return md5hex(
+    process.hrtime().concat(process.pid).map(String)
+  )
+}
+
 NYC.prototype.writeCoverageFile = function () {
   var coverage = coverageFinder()
   if (!coverage) return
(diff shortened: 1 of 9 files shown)


bcoe:
@addaleax I like this as a start to addressing #158! It looks like the tests are failing because Travis is outputting the hierarchy in a different ordering than your local machine, it outputs branch `4` and then branch `3` -- adding a sorting step might be a good fix for this?

A couple thoughts:

* I think It would be worth pulling the process-tree logic into its own module `lib/process.js`?, since the
  logic isn't too coupled with the rest of the logic in index.js.
  * this might mean tweaking our coverage logic a bit, to instrument an additional file.
* I think one thing @isaacs was picturing, was, _could we get the coverage percentage for a given branch?_
  * this could be a cool eventual addition to functionality, and is another argument for starting to isolate the branching logic into its own module.

addaleax:
> adding a sorting step might be a good fix for this?

Done! I definitely should have thought of that… :smile:

> I think It would be worth pulling the process-tree logic into its own module `lib/process.js`?

Sounds like a good idea… I’ve started splitting things up, I guess I’ll see if I get the self-coverage handling right.

> could we get the coverage percentage for a given branch?

Definitely, and I agree that that would a pretty neat thing to have! It might require touching a bit more of the reporting logic, though.

index.js
```diff
@@ -326,7 +336,7 @@ NYC.prototype.clearCache = function () {
 }
 
 NYC.prototype.createTempDirectory = function () {
-  mkdirp.sync(this.tempDirectory())
+  mkdirp.sync(this.processInfoDirectory())
```
bcoe: could we make it so we only create this directory, if we've enabled process tracking; my thinking is, let's make the behavior 100% identical, unless you are outputting the process tree.

index.js
```diff
@@ -22,6 +22,13 @@ var pkgUp = require('pkg-up')
 var testExclude = require('test-exclude')
 var yargs = require('yargs')
 
+var ProcessInfo
+try {
+  ProcessInfo = require('./lib/process.covered.js')
```
bcoe: thanks for doing this, would like us to get in the habit of splitting this repo into more modules.

package.json
```diff
@@ -72,6 +72,7 @@
   "author": "Ben Coe <ben@npmjs.com>",
   "license": "ISC",
   "dependencies": {
+    "archy": "^1.0.0",
```
bcoe: we should also add archy to the bundled dependency list.

index.js
bcoe: do we need to add `process.js` to the list of files in `./build-self-covered.js`, and should it be added to our exclude rules in the packages.json, I noticed `index.covered.js` is.

bcoe:
> Definitely, and I agree that that would a pretty neat thing to have! It might require touching a bit more of the reporting logic, though.

Let's hold off on this until V2 💃 

I made a few more comments, also we should definitely figure out a good place to document this in the README; this having been said, I'm really loving this feature.

index.js
addaleax: > do we need to add `process.js` to the list of files in `./build-self-covered.js`, and should it be added to our exclude rules in the packages.json, I noticed `index.covered.js` is.

Well, yeah, but it is? I did change `build-self-coverage.js` here, and both `.gitignore` and the `files` field of the `package.json` contain rules to exclude *any* `*covered.js` file.

bcoe: @addaleax shoot, missed the updated _ignore me!_

index.js
addaleax: hmm, I’ll try that

addaleax:
> Let's hold off on this until V2

You mean, this PR, or per-subtree coverage reporting? I’m fine with either, but if you’re talking about the latter, I’d want to try and implement per-subtree reporting to make sure everything’s wired up in order  for that to work

bcoe:
> You mean, this PR, or per-subtree coverage reporting?

@addaleax I think this pull request is great, and we should land it ASAP -- just want to get the behavior identical if it's not enabled, and add some documentation to the README 👍

addaleax:
> just want to get the behavior identical if it's not enabled

I think that’s the case. :)

> and add some documentation to the README

Done! Please make sure this PR is in the state you’d like it to see in :)

bcoe:
@addaleax awesome, on a short vacation to NYC, but will hopefully land some pull requests tomorrow or Monday.

isaacs:
> on a short vacation to NYC, but will hopefully land some pull requests tomorrow or Monday.

That's right, @bcoe is so committed to Open Source, he takes vacations just to visit his node modules.

addaleax:
@isaacs … glad to hear your dad joke license arrived ;) (ok truth is i have been giggling at this for five minutes straight)

bcoe:
@isaacs is this what you were picturing with #158, seems like a great start.

isaacs:
@bcoe @addaleax Yeah, this looks really cool.  It strikes me now that maybe storing the environment variables is a bad idea, since people will occasionally publish a `coverage` or `.nyc_output` folder to npm or github, and environs often have passwords in them.

bcoe:
@addaleax mind giving the `next` release a spin:

if QA is looking good, we'll get this promoted soon.

addaleax:
@bcoe Everything looks good so far :)

PR #366, opened 2016-08-24
Update README.md

matheusrocha89:
Adding note that since version 8.0 the node_modules is no more excluded automatically

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ the array is a glob pattern indicating which paths should be omitted.
 
 Globs are matched using [micromatch](https://www.npmjs.com/package/micromatch).
 
-For example, the following config will exclude any files with the extension `.spec.js`, 
+For example, the following config will exclude any files with the extension `.spec.js`,
 and anything in the `build` directory:
 
 ```json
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -185,6 +184,8 @@ directory:
   }
 }
 ```
+> Note: Since version 8.0 the `node_modules` folder is not automatically excluded,
+you need o explicit add in your exclude array
```
JaKXz: `s/o/to`

README.md
```diff
@@ -185,6 +184,8 @@ directory:
   }
 }
 ```
+> Note: Since version 8.0 the `node_modules` folder is not automatically excluded,
```
JaKXz: We should note this only if you specify anything for `exclude`.

README.md
```diff
@@ -185,6 +184,8 @@ directory:
   }
 }
 ```
+> Note: Since version 8.0 if you add a "exclude" array the `node_modules`
+folder is not automatically excluded, you need to explicit add it to your exclude array
```
bcoe: > you will need to explicitly add it to your exclude array

bcoe:
@matheusrocha89 one grammar nit, thanks for the contribution!

PR #376, opened 2016-09-01
update istanbul-lib-source-maps dependency version

sc0ttyd:


sc0ttyd:
Just noticed that this is already in @bcoe 's previous PR. Closing.

PR #385, opened 2016-09-08
Added report types to readme

4uto:
stolen from "istanbul help reports"

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ you can use any reporters that are supported by istanbul and even use several at
 ```bash
 nyc report --reporter=<format> --reporter=<another format>
 ```
-<format> is one of:
+\<format\> is one of:
 * clover
  * XML coverage report that can be consumed by the clover tool
 * cobertura
--- a/README.md
+++ b/README.md
@@ -157,10 +157,44 @@ To view your coverage report:
 
 <img width="500" src="screen.png">
 
-you can use any reporters that are supported by istanbul:
+you can use any reporters that are supported by istanbul and even use several at once:
 
 ```bash
-nyc report --reporter=lcov
+nyc report --reporter=<format> --reporter=<another format>
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -157,10 +157,44 @@ To view your coverage report:
...
 <img width="500" src="screen.png">
 
-you can use any reporters that are supported by istanbul:
+you can use any reporters that are supported by istanbul and even use several at once:
 
 ```bash
-nyc report --reporter=lcov
+nyc report --reporter=<format> --reporter=<another format>
+```
+<format> is one of:
```
addaleax: This `format` should probably be escaped here, e.g. `` `format` ``? Otherwise it doesn’t display in the rendered README. :)

4uto:
I escaped a <format>, thanks Anna Henningsen

bcoe:
@4uto @addaleax sorry to be _that nit-picky OSS maintainer_, but I think I would rather see this new content in this README.md:

https://github.com/istanbuljs/istanbul-reports

With a section added to the `nyc` README that says, `see a full list of reports here`.

I think this would make it more likely that we keep documentation up to date as we add additional reports. What do you think?

bcoe:
@4uto just checking in, are you alright with this change?

4uto:
@bcoe Sorry for the delay. My new baby is keeping me away from the computer.

Your thinking is sound. I was just under the impression that nyc was the new black and that istanbul wasn't "in the front" anymore. I'll go make the change over there and then come back here with an update like you suggested.

bcoe:
@4uto thanks, and babies are more important that JavaScript test coverage! 👶 

My personal goal, is to keep the individual READMEs fairly concise, but to gradually flesh out http://istanbul.js.org/ further:

https://github.com/istanbuljs/istanbuljs.github.io

Issue #379, opened 2016-09-04
Can't override package.json reporters via CLI


gustavohenke:
## Expected Behavior
Override `reporter` configuration in package.json via CLI `--reporter` argument.

## Observed Behavior
The CLI argument `--reporter` doesn't work when there's something under nyc's `reporter` key in package.json.

## Code (or Repository) that Reproduces Issue
gustavohenke/firebase-stub@21487a9af497a8bfcf5015be019da28bcfdc3483

I just started the project and haven't decided how to document it yet, so...

```
npm install
npm test
nyc report --reporter=text-lcov
```

## Forensic Information

**Operating System:**
Windows 10

**Environment Information:**
```
λ node --version
v6.4.0

λ npm --version
3.10.6
```

gustavohenke:
To keep the issue template clean, I forgot to add one thing: thanks for such amazing tool!
Much better than doing various hacks to get some coverage out of mocha tests.

bcoe:
@gustavohenke I agree that `--reporter` should take precedence over the reporter in package.json.

edmundito:
I just ran into the same issue as well. Another thing I noticed is that if you have this in your configuration:

```json
"reporters": []
```

Then `nyc report` does not do anything.

trainerbill:
Yeah this really bit me when using coveralls:

```
"coverage": "nyc report --reporter=text-lcov | coveralls",
```

having reporters specified in the package.json overrides the coverage script and you get an ugly error everytime on travis:

![image](https://cloud.githubusercontent.com/assets/3578259/18462725/84ec4c64-794a-11e6-8c91-c13e8e28c71d.png)

looks like nyc report --reporter=text-lcov is passing in the summary and not the lcov.  removing reporters from package.json -> nyc -> reporters solved the problem.

bcoe:
@trainerbill see #388 would love the extra eyes.

bcoe:
@gustavohenke this is now deployed \o/

gustavohenke:
Yay!

PR #388, opened 2016-09-13
feat: refactored config to fix precedence of config vs. args

bcoe:
see #379 

* refactors config to simplify logic and fix order that args vs. config is applied.
* should improve performance since yargs is only applied once now.

@mourner I will be landing your performance fixes to lib-instrument along with these other fixes, in the next nyc release.

reviewers: @addaleax, @JaKXz, @mourner

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-var arrify = require('arrify')
+var Config = require('../lib/config')
 var foreground = require('foreground-child')
 var NYC
 try {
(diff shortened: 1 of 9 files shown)


mourner:
@bcoe the config refactor looks good, the only thing that doesn't fit usual JS conventions is the `Config` object, which is both a factory and a namespace, although typically capitalized names in JS are classes which are instantiated with `new`. For clarity, I'd turn the factory into another function (`exports.makeConfig`), and make the namespace lowercase (or maybe rename to `config-util`?).

PR #389, opened 2016-09-13
fix: upgrade to newer versions of source-map, signal-exit, and instrument

bcoe:
a few somewhat frightening dependencies being upgraded, we should test thoroughly.

@mourner mind taking this for a spin, it's what actually lands your performance fixes (along with some performance fixes I snuck into #388, yargs should now execute much less frequently).


mourner:
@bcoe I'll try this out, thanks!

mourner:
@bcoe this branch works great for me.

bcoe:
@mourner seeing any significant performance gains, vs. `latest`?

mourner:
@bcoe seeing the same performance gains I described in https://github.com/istanbuljs/istanbul-lib-instrument/pull/22

bcoe:
@mourner cool, was curious if spawning yargs less frequently (which I believe has a 100ms overhead) would improve performance further -- this would specifically target people using `nyc` for instrumentation, rather than `babel-plugin-istanbul`.

mourner:
@bcoe it probably shaved off a few hundred ms, but it's not really noticeable with a test suite that runs for 6+ minutes. :)

PR #394, opened 2016-09-15
fix: add a feature which allows us to bust the cache when breaking changes are introduced

bcoe:
when we make potentially breaking changes to the cache, e.g.,

https://github.com/istanbuljs/istanbul-lib-instrument/pull/22

we should increment the cache version, forcing an invalidation so that folks don't see issues like #392

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,7 +27,9 @@ try {
   ProcessInfo = require('./lib/process.js')
 }
 
-var CACHE_VERSION = '2'
+// bust cache whenever nyc is upgraded, this prevents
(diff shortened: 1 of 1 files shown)


timoxley:
I have a feeling this CACHE_VERSION should be automatically derived from package versions.

Cost of accidentally not bumping this version, due to forgetting or not realising it broke, is high.
Cost of nyc having to rebuild the cache after each update is low.

i.e. better to be safe and take a few extra seconds once every update than to potentially introduce fatal & hard to diagnose bugs.

PR #411, opened 2016-10-04
Istanbul-lib-source-maps 1.0.2

mourner:
> mind submitting the corresponding pull to nyc; I think that's the only other place we need to update?

@bcoe I assume you referred to a dependency bump?

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "istanbul-lib-hook": "^1.0.0-alpha.4",
     "istanbul-lib-instrument": "^1.1.3",
     "istanbul-lib-report": "^1.0.0-alpha.3",
-    "istanbul-lib-source-maps": "^1.0.1",
+    "istanbul-lib-source-maps": "^1.0.2",
     "istanbul-reports": "^1.0.0-alpha.8",
     "md5-hex": "^1.2.0",
     "micromatch": "^2.3.11",


JaKXz:
Since it's a patch shouldn't it just be picked up automatically?

mourner:
@JaKXz yes, but this should also in theory force an update if you do `npm install`, whereas without this you would need `npm update`. I think it doesn't hurt to update a patch dep if it contains a critical fix.

JaKXz:
@mourner I agree - was just curious :)

bcoe:
@JaKXz @mourner fun fact, we bundle-dependencies on this repo:

1. to improve install time (because our dependencies were getting pretty hefty).
2. it fixes some singleton related issues; by moving things into a sub-folder (you end up with two copies of libraries, rather than one when using npm 3).

Long story short, we need to explicitly bump dependencies on nyc.

evilebottnawi:
@JaKXz @bcoe some problems with install `nyc@8.3.0` and `istanbul-lib-source-maps@1.0.2`
I am try
```
rm -r node_modules/nyc
npm cache clean
npm i nyc
```
I get `nyc@8.3.0` but with `istanbul-lib-source-maps@1.0.1`, maybe someone from packages require `istanbul-lib-source-maps@1.0.1` and we have `istanbul-lib-source-maps: ^1.0.1`, so I get not last version

mourner:
@evilebottnawi there were no `nyc` release after https://github.com/istanbuljs/nyc/pull/411 yet. See @bcoe's last comment there.

evilebottnawi:
@mourner  understand it, and therefore I say that when you try to install `nyc@8.3.0`, i got `istanbul-lib-source-maps@1.0.1`, not `istanbul-lib-source-maps@1.0.2`

mourner:
@evilebottnawi which is expected due to bundling of dependencies, as per https://github.com/istanbuljs/nyc/pull/411#issuecomment-251514484

evilebottnawi:
@mourner I should wait for a new version or it may be to solve?

JaKXz:
@evilebottnawi I would suggest installing it from `nyc@next` to get the bug fix in and give it a spin :)

`npm cache clear; npm i -D nyc@next`

PR #416, opened 2016-10-08
feat: implement coverage details for process tree

addaleax:
Show line coverage for each subtree of the tree of spawned processes.

This is just a start, but while I definitely wanted to begin working in that direction as planned, I’m not too certain about how this is best implemented when it comes to visual display… In theory, a full per-file coverage table could be shown for each subtree, but that seems like overkill, so I’ve just implemented something small. Also, maybe printing this extra information should be optional/configurable, too?

What do you think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -413,7 +413,7 @@ function coverageFinder () {
 NYC.prototype._getCoverageMapFromAllCoverageFiles = function () {
   var map = libCoverage.createCoverageMap({})
 
-  this._loadReports().forEach(function (report) {
+  this.loadReports().forEach(function (report) {
     map.merge(report)
   })
 
@@ -439,7 +439,9 @@ NYC.prototype.report = function () {
 }
 
 NYC.prototype.showProcessTree = function () {
-  console.log(this._loadProcessInfoTree().render())
+  var processTree = ProcessInfo.buildProcessTree(this._loadProcessInfos())
+
+  console.log(processTree.render(this))
 }
 
 NYC.prototype.checkCoverage = function (thresholds) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -460,7 +460,24 @@ NYC.prototype.checkCoverage = function (thresholds) {
 }
 
 NYC.prototype._loadProcessInfoTree = function () {
-  return ProcessInfo.buildProcessTree(this._loadProcessInfos())
+  var _this = this
+  var processTree = ProcessInfo.buildProcessTree(this._loadProcessInfos())
```
bcoe: I'd be tempted to encapsulate this logic inside `process.js`; and all that `index.js` would need to do would be:

`console.log(processInfo.render())`.

index.js
```diff
@@ -460,7 +460,24 @@ NYC.prototype.checkCoverage = function (thresholds) {
...
+
+  processTree.getCoverageMap(function (filenames, maps) {
+    var map = libCoverage.createCoverageMap({})
+
+    _this._loadReports(filenames).forEach(function (report) {
```
bcoe: we can rename `_loadReports` to `loadReports`, and `process.js` can call this method on `index.js`.

lib/process.js
```diff
@@ -57,6 +63,20 @@ ProcessInfo.buildProcessTree = function (infos) {
   return treeRoot
 }
 
+ProcessInfo.prototype.getCoverageMap = function (merger) {
```
bcoe: This is cool! it works because we already output the coverage for each subprocess to a separate file named after the pid?

index.js
addaleax: @bcoe I just updated this PR which what I *think* you had in mind – maybe you meant something totally different, this feels a bit icky now (as in, this doesn’t seem quite right from a separation of concerns pov?).

bcoe:
@addaleax love this refactor 👍

bcoe:
@addaleax give this a spin when you have a sec:

`8.4.0` is a pretty big change to how the underlying instrumentation works, so I would appreciate extra eyes.

PR #430, opened 2016-10-30
feat: add --config-file option for .nycrc

laggingreflex:


bcoe:
@laggingreflex what's the specific use-case you have for this; I'm leaning towards not adding an additional configuration option, but would love to better understand where you're coming from 👍

laggingreflex:
Ah. Never mind in that case. It probably wouldn't have been too convincing anyways, just a personal preference to keep all tests related stuff in the tests folder. All the other stuff was configurable, like mocha.opts, karma.conf, even the coverage folders, it was just this nycrc sticking out. No biggie though.

bcoe:
@laggingreflex thanks for being understanding 👍 I'm trying to, within reason, be aggressive about not adding too much more configuration to nyc, since it's already got _a ton_ of dials and knobs.

PR #437, opened 2016-11-07
Update README.md

thepont:
There is very little chance people will need coverage as more then a dev dependency.


bcoe:
:shipit: thanks for the contribution @thepont.

Issue #438, opened 2016-11-07
add more tests for produce-source-map logic


bcoe:
#393 adds logic around merging source-maps, introducing the `merge-source-map` dependency; we should add some tests that exercise this new dependency.

CC: @kpdecker, @JaKXz

PR #393, opened 2016-09-15
Add produceSourceMap option

kpdecker:
This allows the user to control the source map output post instrumentation.

Not fully straightforward right now, but when combined with a fork of source-map-support that has inline support (or placing the files in the proper location), end to end covered source maps can be achieved.

For runtime source-map-support, see https://github.com/evanw/node-source-map-support/pull/118. Will be forking to @kpdecker/source-map-support soon, since the canonical project appears to be dead.

Fix for #285 (When combined with the above)

PR #439, opened 2016-11-07
feat: adds support for source-map production

bcoe:
@kpdecker I had some refactoring on my mind, that I opted to tack onto your work so that we can get this landed -- rather than overly marking up #393.

I made a few very minor changes:

* I think that source-map production and the application of source maps, are different enough behavior that it's worth introducing the additional flag `--produce-source-map`; I also had some performance concerns, and think it's better that folks using `source-map-support` opt in.
* I've been trying to gradually pull more and more logic out of `index.js`, so that it's ultimately just calling a bunch of pretty easy to follow libraries and helpers. tldr; I pulled the source-map logic into the instrumenter.
* I added a couple `TODOs` around adding additional tests, and opened #438.

CC: @JaKXz would love to have you take one final look at this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,14 +51,12 @@ and a `text-lcov` coverage report.
 nyc --reporter=lcov --reporter=text-lcov npm test
 ```
 
-## Stack Traces
+## Accurate stack traces using source maps
 
-When `source-map` handling enabled, then the instrumented source files will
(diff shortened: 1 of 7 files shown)


kpdecker:
Awesome. LGTM

Issue #456, opened 2016-11-27
No colors when not executed via a tty


ngryman:
## Expected Behavior

To have a colored output.

## Observed Behavior

Colors have gone 👻 

## Proposed fix

I think you do that on purpose when no `tty` is detected, which is perfectly fine because it makes easier to process the output if needed.
That said, in some use cases (like mine) one might just want to forward your output. To do so it would be nice to have a flag to force colored output (others do `--ansi` or `--colors`).

Thanks!

## Bonus Points! Code (or Repository) that Reproduces Issue

```js
import { spawn } from 'child_process'

spawn('nyc ava', { stdio: 'inherit' })
```

## Environmentt

```
$ node --version
v6.9.1
$ npm --version
3.10.8
```

mlrawlings:
Have you tried passing `--color` as a command line flag? Or `FORCE_COLOR` as an environment variable?

https://github.com/chalk/supports-color#info

If that does work, it would probably be worth adding this to the docs somewhere. Maybe under [Running Reports](https://github.com/istanbuljs/nyc#running-reports)?

ngryman:
@mlrawlings `--color` works indeed! Thanks 👌
It would be worth mentioning it in the docs.

bcoe:
@ngryman 👍 if you're bored feel free to take a stab at the docs :) otherwise let's leave this open as a feature request.

xenithorb:
`--color` doesn't work anymore. Where's the love for CI pipelines?

electrovir:
I ran into this same issue. Apparently the flags are order sensitive. I'm only using two CLI flags, `--colors` and `--nycrc-path`:

- If I put `--colors` after my `--nycrc-path` cli flag, `nyc` errors out
- If I put `--colors` _before_ my `--nycrc-path` cli flag, everything works great and I get colors.

PR #457, opened 2016-11-30
docs: add color option (#456)

ngryman:
I think mentioning it in the `cli` usage should be enough.

### DIFF
--- a/lib/config-util.js
+++ b/lib/config-util.js
@@ -110,6 +110,11 @@ Config.buildYargs = function (cwd) {
       type: 'boolean',
       describe: "don't output a report after tests finish running"
     })
+    .option('color', {
+      default: false,
+      type: 'boolean',
+      describe: 'force colored output even when no tty is detected'
+    })
     .option('all', {
       alias: 'a',
       default: false,


bcoe:
@ngryman thanks \o/ the failing tests are a standard nit, you just need to remove the double quotes from the description string.

I'm surprised passing `--color` bubbles appropriately up to the sub-process; what if we made it so setting `--color` also results in nyc setting the `FORCE_COLOR` both in its own environment variables and in subprocesses?

CC: @mlrawlings does this seem like a reasonable approach?

ngryman:
@bcoe Fixed.

mlrawlings:
@bcoe 
> I'm surprised passing --color bubbles appropriately up to the sub-process;
1. `nyc` passes arguments down to the sub process at least one level (`nyc ava -s`).  I haven't had a chance to see exactly what is happening at subsequent levels.

2. Do you pass all arguments, or only those after the initial command?  In the case of  `nyc -asc ava -s`, I'm guessing `-asc` does _not_ get passed to `ava`.

    Is there a difference between `nyc --color ava` and  `nyc ava --color`?  Do both work?

> what if we made it so setting --color also results in nyc setting the FORCE_COLOR both in its own environment variables and in subprocesses?
3. Using `FORCE_COLOR` is only good for ensuring color as the presence of it with **any** value forces colors to be used.  But the lack of the env variable doesn't mean "no colors".  In any case, this may be enough in this situation.  Another option could be to open a PR against `supports-color` for a `FORCE_NO_COLOR` option.

bcoe:
@ngryman could you provide an example of how you're using the `--color` flag, if you're doing something like @mlrawlings indicates:

`nyc ava --color`, there's no need for us to add `--color` to nyc's help. I think if we do have a reason for `nyc` itself to handle color, we should be doing something like this:

```js
if (argv.color) {
  process.env.FORCE_COLOR = 'true'
  env.FORCE_COLOR = 'true' // where env is the variable we pass to subprocesses.
}
```

ngryman:
@bcoe Sure, here it is: https://github.com/ngryman/meta-dev/blob/master/bin/meta.js#L58.

bcoe:
@ngryman 👍 given that we already pass any unknown command line arguments to the bin that `nyc` is instrumenting, I think we can avoid adding this additional option to the CLI.

Having said this, given the wide variety of techniques and approaches folks use for instrumenting code at this point, I think it would worth considering adding a section to the website that links to example projects, and provides tips and tricks like this:

I've been thinking a lot about documentation lately, as the other project I work on yargs is also very much in need of a documentation overhaul -- would love any thoughts folks have as to how we could make features, like passing arguments to the instrumented bin, stand out more.

The website, and corresponding repo can be found here:

https://istanbul.js.org/

and

https://github.com/istanbuljs/istanbuljs.github.io

respectively, if anyone wants to strike up a conversation about how we move forward with documentation.

PR #469, opened 2016-12-10
feat: reporting watermarks

xdissent:
Passes watermarks from the nyc config to `istanbul-lib-report`. This should allow custom watermarks for all reporters in `istanbul-reports` that use them. In the old `.istanbul.yml` the watermarks lived under a `reporting` key, so maybe this should be there too, but I read thru some of the other PRs about the new config and thought this might make more sense in the context of those discussions. Plus `reportDir` was already top level too.

### DIFF
--- a/index.js
+++ b/index.js
@@ -433,7 +433,8 @@ NYC.prototype.report = function () {
   var tree
   var map = this._getCoverageMapFromAllCoverageFiles()
   var context = libReport.createContext({
-    dir: this._reportDir
+    dir: this._reportDir,
+    watermarks: this.config.watermarks
   })
 
   tree = libReport.summarizers.pkg(map)


bcoe:
@xdissent looking good to me. two asks for you:

1. would it be possible to add a test? an integration test using the bin test suite is great.
2. I think we should document this config setting in https://github.com/istanbuljs/nyc/blob/master/lib/config-util.js#L66

xdissent:
@bcoe I looked at the tests yesterday and I didn't really see a way to test the watermarks config's effect on anything. It's kinda difficult to add in a custom reporter at the moment, and none of the stock reporters seem to output anything differently based on watermarks - except colors, and colors don't appear to be used in the test output. Any suggestions? For the documentation, I also hit a snag because the watermarks config is an object and yargs isn't very helpful with object args IIRC. Istanbul didn't have cli options for the watermarks either, so I might need some guidance on how that should look if they need to be cli opts now too.

bcoe:
@xdissent haven't forgotten about this pull, but am digging myself out of a backlog of open-source work; hope to take a look at how one would test this soon 👍

bcoe:
@xdissent thanks for the contribution, hopefully the first of many. I'm going to try to have a candidate release out today.

bcoe:
@xdissent would love your help testing:

If everything looks good in our smoke tests I will promote the release soon. Thanks for the contribution, I look forward to more in the future 👍

PR #474, opened 2016-12-17
chore(package): update spawn-wrap to 1.3.2

TheAlphaNerd:
This fixes the wrapping of custom node binaries

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "resolve-from": "^2.0.0",
     "rimraf": "^2.5.4",
     "signal-exit": "^3.0.1",
-    "spawn-wrap": "^1.2.4",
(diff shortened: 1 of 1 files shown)


bcoe:
@TheAlphaNerd nice, thanks for updating this.

bcoe:
@TheAlphaNerd I will need to spin up a VM and dig into why this is breaking on Windows.

bcoe:
@isaacs @TheAlphaNerd here's the bug:

https://github.com/tapjs/spawn-wrap/blob/master/index.js#L133

Should be `.test()` not `.match()` 👍

TheAlphaNerd:
sent in a PR to spawn-wrap... will update this PR when it lands

PR #482, opened 2017-01-06
Fix use of --all when instrumentation is disabled

wbyoung:
This allows the use of `--all` with `babel-transform-istanbul`, for instance.

### DIFF
--- a/lib/instrumenters/noop.js
+++ b/lib/instrumenters/noop.js
@@ -3,12 +3,12 @@ var readInitialCoverage = require('istanbul-lib-instrument').readInitialCoverage
 
 function NOOP () {
   return {
-    instrumentSync: function (code) {
+    instrumentSync: function (code, filename) {
       var extracted = readInitialCoverage(code)
       if (extracted) {
         this.fileCoverage = new FileCoverage(extracted.coverageData)
       } else {
-        this.fileCoverage = null
+        this.fileCoverage = new FileCoverage(filename)
       }
       return code
     },


bcoe:
@wbyoung this looks good; seems like adding empty coverage is pretty cheap, and if it corrects some upstream bugs worth doing.

how hard would it be to add an integration test; even if it's just in the `cli` test suite.

wbyoung:
@bcoe I gave it a shot, but after copying one of the other tests that seemed related, I wasn't able to get it failing as expected with the current code.

I can try again when I have some free time if it's important. Any suggestions on which test to model it after?

bcoe:
@wbyoung I don't think this needs to be a blocker; could you provide an example using `babel-transform-istanbul` that fails without the FileCoverage object being instantiated?

wbyoung:
@bcoe see https://github.com/wbyoung/nyc-issue-482-example and the fact that `src/not-included.js` is not part of the test result files.

wbyoung:
FYI, the comment I left [over here](https://github.com/istanbuljs/nyc/pull/483#issuecomment-273378074) about the builds https://github.com/wbyoung/relateify/issues/1 and https://github.com/wbyoung/babel-plugin-transform-postcss/issues/3 failing appears to be "fixed" by removing the `all` config option, so it's probably more related to this PR.

bcoe:
see: https://github.com/istanbuljs/istanbul-lib-coverage/pull/5

landing a hot-fix now.

bcoe:
@wbyoung this hot-fix didn't do the trick; it seems as though defaulting to an empty FileCoverage report causes quite a few issues when the file is then, in turn, covered later on.

I'm rolling back this change for the time being -- I don't think it should actually be necessary if we can get `--all` walking all the files in a project appropriately.

bcoe:
@wbyoung `nyc@latest` should be fixed 👍

PR #490, opened 2017-01-13
Allow eager instantiation of instrumenter

evantorrie:
This change adds an `--eager` option to nyc which forces instantiation
of the instrumentation library prior to walking any of the files.

Addresses #491

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,6 +48,7 @@ function NYC (config) {
   this._reportDir = config.reportDir || 'coverage'
   this._sourceMap = typeof config.sourceMap === 'boolean' ? config.sourceMap : true
   this._showProcessTree = config.showProcessTree || false
+  this._eagerInstantiation = config.eager || false
   this.cwd = config.cwd || process.cwd()
 
   this.reporter = arrify(config.reporter || 'text')
@@ -91,7 +92,7 @@ function NYC (config) {
 NYC.prototype._createTransform = function (ext) {
   var _this = this
 
-  return cachingTransform({
+  var opts = {
     salt: JSON.stringify({
       istanbul: require('istanbul-lib-coverage/package.json').version,
       nyc: require('./package.json').version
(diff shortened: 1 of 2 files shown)


JaKXz:
@bcoe this is the PR for #490 - 

Do you guys think it's possible/valid to write a test for this? A full acceptance test with a library that mocks requires would be thorough but possibly overkill.

bcoe:
@evantorrie another thought comes to mind; is there any reason to not always enable `eager`? I'm correct in thinking that it's a noop, and should improve performance?

CC: @jamestalmage, @novemberborn

novemberborn:
> is there any reason to not always enable eager? I'm correct in thinking that it's a noop, and should improve performance?

It'll decrease performance in case the required modules have previously been instrumented. That's what `CachingTransform` is there for, so we don't have to load Istanbul and all it's dependencies on every test run.

bcoe:
@JaKXz realizing now that this is a feature of caching-transform, I think we can probably do without the test; trusting that it's sufficiently tested in the dependent library.

@evantorrie this is looking good to me; this is also enough of a power feature that I could be convinced that the command line argument is sufficient documentation; perhaps we could switch it to:

```
instantiate the instrumenter at startup (see https://git.io/vMKZ9)
```

I'll also be quite curious to see the benchmarks of using both approaches on a few large projects after we land this -- given that we have some expensive dependencies like `babel`, I wonder if average projects will see some performance gains (we might want to enable this by default at some point).

evantorrie:
I've changed the command line description to reference `https://git.io/vMKZ9`.

Per discussion above, my use-case for this is mostly in running continuous integration via Jenkins which has a fresh checkout from source on each build.  Thus, I never get any benefit from the caching transforms (although obviously I recognize you do if this is run in your dev directory/env).

bcoe:
@evantorrie looks great, I'll look into landing this ASAP.

bcoe:
@evantorrie I've released this feature in the `next` release of `nyc`. I would love help testing:

Look forward to more contributions in the future, thanks! 🎉

PR #493, opened 2017-01-16
chore: add test and docs for high and low watermarks

bcoe:
added some docs and a test to @xdissent's high and low watermark work.

would love your 👀 on it @xdissent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,29 @@ nyc
     └── /usr/local/bin/node /path/to/your/project/node_modules/ava/lib/test-worker.js …
 ```
 
+## High and low watermarks
+
+Several of the coverage reporters supported by nyc display special information
+for high and low watermarks:
+
+* high-watermarks represent healthy test coverage (in many reports
+  this is represented with green highlighting).
+* low-watermarks represent sub-optimal coverage levels (in many reports
+  this is represented with red highlighting).
+
+You can specify custom high and low watermarks in nyc's configuration:
+
+```json
+{"nyc": {
+  "watermarks": {
+    "lines": [80, 95],
+    "functions": [80, 95],
+    "branches": [80, 95],
+    "statements": [80, 95]
+  }
+}}
+```
+
 ## Integrating with coveralls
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -312,6 +312,29 @@ nyc
...
+
+* high-watermarks represent healthy test coverage (in many reports
+  this is represented with green highlighting).
+* low-watermarks represent sub-optimal coverage levels (in many reports
+  this is represented with red highlighting).
+
+You can specify custom high and low watermarks in nyc's configuration:
+
+```json
+{"nyc": {
```
JaKXz: style nit: I think it's better to have this be consistent with the other json formatting in the readme [i.e. with newlines and indenting].

test/src/nyc-bin.js
```diff
@@ -689,4 +689,39 @@ describe('the nyc cli', function () {
...
     })
   })
+
+  it('allows alternative high and low watermarks to be configured', function (done) {
+    var args = [
+      bin,
+      '--watermarks.lines=90',
+      '--watermarks.lines=100',
+      '--watermarks.statements=30',
+      '--watermarks.statements=40',
```
JaKXz: I think I can follow the logic here of setting the "low" watermark as the first arg and high as the second, but I'd prefer the DX to be more explicit, e.g.

```
--watermarks.lines.low=90
--watermarks.lines.high=100
```

bcoe: @JaKXz I see your point of view on this, and this would probably have been my inclination; however, this implementation matches `istanbul`'s:

https://github.com/gotwarlost/istanbul/issues/116

so I think it might be less surprising, it also allows us to kick the configuration directly along to the istanbul internals.

JaKXz: I guess it extends the scope too far to change the underlying API eh...

bcoe: @JaKXz that's my feeling; I think it's an advanced enough option that we're probably okay with sticking with the historical implementation.

xdissent:
Looks good to me! Thanks @bcoe

PR #303, opened 2016-07-07
fix: disable the babel/nyc cache

bcoe:
the interactions between the ava, babel, and nyc cache were causing a lot of confusion for folks, see:

https://github.com/istanbuljs/nyc/issues/296
https://github.com/istanbuljs/nyc/issues/294

In this pull request:

* we leave nyc's cache as opt in (I was considering making it run by default in 7.x).
* we disable babel's cache when running tests with nyc (this cache caused a lot of issues with babel-plugin-istanbul).

PR #508, opened 2017-01-28
feat: allow user to turn on babel cache

Arturszott:
Hey! Recently we have switched to nyc and I was a bit surprised when I saw that babel cache is turned off. Then I've found this PR https://github.com/istanbuljs/nyc/pull/303. Unfortunately it does not explain the reason why nyc works sometimes incorrect when babel cache is turned on. 

I believe that it all depend on the rest of the test stack. I've made a couple tests and there was 50% of speed improvement for my project, so this gave me an idea. 

Why not to allow babel cache to be optional?

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -1,8 +1,5 @@
 #!/usr/bin/env node
 
-// the babel cache does not play nicely with nyc.
-process.env.BABEL_DISABLE_CACHE = '1'
-
 var configUtil = require('../lib/config-util')
 var foreground = require('foreground-child')
 var NYC
@@ -24,6 +21,11 @@ var yargs = configUtil.decorateYargs(configUtil.buildYargs())
 var instrumenterArgs = processArgs.hideInstrumenteeArgs()
 var argv = yargs.parse(instrumenterArgs)
 var config = configUtil.loadConfig(instrumenterArgs)
+
+if (config['babel-cache'] === false) {
+  // the babel cache in few cases does not play nicely with nyc.
+  process.env.BABEL_DISABLE_CACHE = 1
+}
 
 if (argv._[0] === 'report') {
   // run a report.
@@ -50,8 +52,8 @@ if (argv._[0] === 'report') {
     NYC_CONFIG: JSON.stringify(config),
     NYC_CWD: process.cwd(),
     NYC_ROOT_ID: nyc.rootId,
-    BABEL_DISABLE_CACHE: 1,
-    NYC_INSTRUMENTER: config.instrumenter
+    NYC_INSTRUMENTER: config.instrumenter,
+    BABEL_DISABLE_CACHE: process.env.BABEL_DISABLE_CACHE
   }
   sw([wrapper], env)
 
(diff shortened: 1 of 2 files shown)


bcoe:
@Arturszott when you're using nyc with its own cache turned on, the two caches would interfere with each other; causing test coverage not to reload, and other weirdness.

I turned off babel-cache before [babel-plugin-istanbul](https://www.npmjs.com/package/babel-plugin-istanbul) existed ... which I think would work better with the cache enabled.

I agree with you that it would be good to make this a configuration setting, with it defaulting to disabled.

bin/nyc.js
```diff
@@ -50,7 +51,7 @@ if (argv._[0] === 'report') {
     NYC_CONFIG: JSON.stringify(config),
     NYC_CWD: process.cwd(),
     NYC_ROOT_ID: nyc.rootId,
-    BABEL_DISABLE_CACHE: 1,
+    BABEL_DISABLE_CACHE: config['babel-cache'],
```
JaKXz: This is my only doubt about this change: this variable was a `number` before, and now it will be a `boolean`. 

Should the config accept a number? or default to a string number (`'1'`)?

JaKXz:
@Arturszott great work, thanks for the contribution. Just a couple questions for you, and then this should be mergeable :)

bin/nyc.js
bcoe: I'd be tempted to do:

```js
if (config['babel-cache'] === false) {
  // the babel cache in few cases does not play nicely with nyc.
  process.env.BABEL_DISABLE_CACHE = '1'
} else {
  process.env.BABEL_DISABLE_CACHE = '0'
}
```

followed by:

`BABEL_DISABLE_CACHE: process.env.BABEL_DISABLE_CACHE`

rather than:

`BABEL_DISABLE_CACHE: config['babel-cache']`

Definitely make sure this approach works on the repo that you've been benchmarking 👍

JaKXz: Good suggestion, though I'd flip the boolean there so that the cache is defaulted to off.

JaKXz: I suppose since yargs handles setting the default this is also just fine.

Arturszott:
Thanks for advices. I've made minor adjustments and checked if it still works with my project. Setting flag to `0` or `false` still evaluates value to `string`. Based on that I decided to set flag value only if it should be disabled.

Is it okay now?

Arturszott:
@bcoe is that okay for you?

bcoe:
@Arturszott @JaKXz this looks good to me 👍 sorry for the slow turnaround have been slightly swamped.

JaKXz:
@bcoe I'll let you hit the merge button, in case you want to tweak anything.

bcoe:
@JaKXz sounds good 👍 I'm hoping to get this landed this evening.

bcoe:
@Arturszott @JaKXz made one slight tweak, and am landing this work in https://github.com/istanbuljs/nyc/pull/517; will keep you updated here.

PR #517, opened 2017-02-07
feat: allow babel cache to be enabled

bcoe:
Allow the babel cache to be disabled, thanks for the patch @Arturszott.

In this pull I make one tiny tweak, the environment variable is now simply not set if you set `babel-cache=true`.

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -22,11 +22,6 @@ var instrumenterArgs = processArgs.hideInstrumenteeArgs()
 var argv = yargs.parse(instrumenterArgs)
 var config = configUtil.loadConfig(instrumenterArgs)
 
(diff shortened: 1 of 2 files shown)


Arturszott:
Thanks! Looks great. Simplicity 🔥

bcoe:
@Arturszott did you ever get a chance to try this out? You should be able to use it if you run:

`npm i nyc@next --save-dev`

karel-3d:
I have found this PR when debugging issue with nyc and babel

What exactly is the problem with babel cache and nyc? I can't find a reference to any issues, only this PR :) no matter how much I look

karel-3d:
...hm, but it's true, there is some issue with nyc and babel cache

Any idea what is the underlying issue? :(

PR #528, opened 2017-02-25
feat: upgrade to version of yargs that introduces extends feature

### LINKED ISSUE #503: Extend .nycrc configurations
Other tools that have similar uses of `.rc` configuration files, like Babel and ESLint, allow you to extend other configuration files. This is a useful option for configuration reuse, when you want to keep mostly the same configuration in multiple projects, and only override a few options.

It doesn't appear that nyc supports this currently. Below is an example of what I would like to do:

node_modules/my-lib/.nycrc
```json
{
    "lines": 95,
    "statements": 95,
    "functions": 95,
    "branches": 95,
    "reporter": [
        "cobertura",
        "lcov"
    ],
    "all": false,
    "check-coverage": true,
    "report-dir": "./coverage",
    "cache": false
}
```

.nycrc
```json
{
  "extends": "./node_modules/my-lib/.nycrc",
  "lines": 80
}
```

bcoe:
config files can now be extended using a key `extends`, which points to a configuration file to inherit from.

fixes #503

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "signal-exit": "^3.0.1",
     "spawn-wrap": "1.2.4",
     "test-exclude": "^4.0.0",
-    "yargs": "^6.6.0",
+    "yargs": "^7.0.0-alpha.3",
     "yargs-parser": "^4.0.2"
   },
   "devDependencies": {


JaKXz:
@bcoe should we just go to v7.0.2 and merge this? :)

bcoe:
@JaKXz we need to do a little bit of refactoring to because all the top-level options will be global now; I think to keep the CLI identical we should just set `global: false` for these.

there might be a couple more tweaks required.

bcoe:
closed in favor of: https://github.com/istanbuljs/nyc/pull/541

@JaKXz would love help testing (`npm i nyc@next` if/when you get two seconds).

PR #532, opened 2017-03-14
feat: allow NYC_NO_INSTRUMENT to be provided, to disable instrumentation

bcoe:
I'm working in istanbuljs' monorepo, and would like to be able to turn off instrumentation for just one of the repos.


popomore:
Do this feature have been implemented in other PR?

PR #534, opened 2017-03-18
Add possibility to filter during report generation

schutm:
Have the --exclude and --include also work when the report command is being used and not only during instrumentation.

E.G. Instrument and write to the cache directory using `nyc --silent ava` and later create specific report using e.g. `nyc report --include '**/bin/*.js` (ignoring all other files located in e.g. **/lib/**

In my specific this is required because my files are precompiled using rollup. This includes also files from external libraries in the resulting files. These files get tested (other the tests will fail due to import statements). However nyc will only exclude files actually present. It doesn't perform this action based on the  sourcemaps. Therefor I filter the report itself.

### DIFF
--- a/index.js
+++ b/index.js
@@ -426,11 +426,13 @@ function coverageFinder () {
 }
 
 NYC.prototype._getCoverageMapFromAllCoverageFiles = function () {
+  var _this = this;
   var map = libCoverage.createCoverageMap({})
 
   this.loadReports().forEach(function (report) {
     map.merge(report)
   })
+  map.filter(function(filename) { return _this.exclude.shouldReport(filename) })
 
   return map
 }
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "find-up": "^1.1.2",
     "foreground-child": "^1.5.3",
     "glob": "^7.0.6",
-    "istanbul-lib-coverage": "^1.0.1",
+    "istanbul-lib-coverage": "schutm/istanbul-lib-coverage",
     "istanbul-lib-hook": "^1.0.0",
     "istanbul-lib-instrument": "^1.4.2",
     "istanbul-lib-report": "^1.0.0-alpha.3",
@@ -96,7 +96,7 @@
     "rimraf": "^2.5.4",
     "signal-exit": "^3.0.1",
     "spawn-wrap": "1.2.4",
-    "test-exclude": "^4.0.0",
+    "test-exclude": "schutm/test-exclude",
     "yargs": "^6.6.0",
     "yargs-parser": "^4.0.2"
   },


schutm:
@bcoe Now with commit https://github.com/istanbuljs/istanbuljs/commit/e1c99d642da3d9372c234b423b5e79bdd0ad3aad I'd like to take the next step and include this possibility into nyc as well. I do see 2 possibilities:
1. Have an additional flag (--absolutepath or something alike) which will make sure the reports but also the default --include and --exclude paths will use absolute paths in stead of relative paths
2. Have report always use absolute paths

Of course other possibilities are also fine with me. What is your preference?

bcoe:
@schutm I think my preference would be to avoid adding the `--absolutepath`; we should be able to rebase the paths in the map to the base directory, so that I can write fairly natural, short include/exclude rules right?

index.js
```diff
@@ -426,11 +426,13 @@ function coverageFinder () {
 }
 
 NYC.prototype._getCoverageMapFromAllCoverageFiles = function () {
+  var _this = this;
   var map = libCoverage.createCoverageMap({})
 
   this.loadReports().forEach(function (report) {
     map.merge(report)
   })
+  map.filter(function(filename) { return _this.exclude.shouldReport(filename) })
```
addaleax: Shouldn’t this be `map = …`? `map.filter` doesn’t change the `map` array, so this looks a bit odd

schutm:
You're 100% correct. Either map = ... or just immediately 'return  
map.filter'
>
> @addaleax commented on this pull request.
>
> In index.js:
>>    var map = libCoverage.createCoverageMap({})
>  this.loadReports().forEach(function (report) {
>     map.merge(report)
>   })
> +  map.filter(function(filename) { return  
> _this.exclude.shouldReport(filename) })
>
> Shouldn’t this be map = …? map.filter doesn’t change the map array, so  
> this looks a bit odd
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.



--

bcoe:
@schutm *bump*, is this still something that you'd like to land?

bcoe:
@schutm going to close this issue for the time being, since it's quickly drifting away from `master`, but _please_ feel free to rebase and reopen as soon as you have some time to see this feature over the finish line.

schutm:
Sorry for the late reply. Feel free to close it. I tried to incorporate it using the stored filenames in the sourcemap. This didn't work out, and afterwards my priorities shifted. Therefor I do not have time, at the moment, to look into this anymore. Once I got time again, I'll rebase it and try again.

bcoe:
@schutm thanks for the work; feel free to revisit any time.

PR #541, opened 2017-03-27
feat: upgrade to version of yargs with extend support

bcoe:
Along with upgrading a few other dependencies, I've finally upgraded us to the version of yargs with @rcoy-v's extends support.

@rcoy-v mind giving this a try; you can test it out by running:

`npm i nyc@next --save-dev`


rcoy-v:
I've tested the following scenarios:

1. `.nycrc` extending another `.nycrc` from node_modules
Works.

2. `package.json "nyc"` extending a `.nycrc` from node_modules
Does not work.


If you could make a checklist or something outlining all the scenarios you expect should work, I will test them.

bcoe:
@rcoy-v huh; I could have sworn we'd tested that scenario in yargs;

I'll land this feature, since we've been waiting quite some time, and it's probably worth us opening an issue on yargs.

rcoy-v:
@bcoe, I opened this [issue](https://github.com/yargs/yargs/issues/843) on yargs with what I believe is the cause.

PR #553, opened 2017-04-12
Test node 6 on travis

SimenB:
node@6 is the current LTS. "stable" is node@7

Might want to consider dropping 0.10, 0.12 and 5 as they are EOL, but as a testing tool, maybe not? 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 node_js:
+  - 6
   - 4
   - 5
   - "0.10"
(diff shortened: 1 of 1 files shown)


JaKXz:
Thanks @SimenB! I also made some updates that clean up our tested versions.

@bcoe didn't we drop support for v0.10 at some point? or not yet?

SimenB:
I tried to look through the changelog, but saw no mention of dropped node versions

JaKXz:
Ok cool, I guess that will happen when `yargs` drops it. Thanks!

PR #556, opened 2017-04-19
fix: respect cache flag in parent process

bcoe:
`--cache` was not being respected for the top-level process, this resulted in coverage information being mapped differently in the parent process than the child-processes, resulting in a fluctuating coverage percentage.

I pulled the source-map logic into its own library while debugging this issue, because it made things simpler to follow.


lib/source-maps.js
```diff
@@ -0,0 +1,56 @@
+var convertSourceMap = require('convert-source-map')
+var libCoverage = require('istanbul-lib-coverage')
+var libSourceMaps = require('istanbul-lib-source-maps')
+var fs = require('fs')
+var path = require('path')
```
JaKXz: You have `const`s elsewhere, but `var`s here. I'm guessing we're going to Node v4+ soon so I'd like to be consistent :)

index.js
```diff
@@ -103,7 +102,9 @@ NYC.prototype._createTransform = function (ext) {
       return hash
     },
     cacheDir: this.cacheDirectory,
-    disableCache: !this.enableCache,
+    // when running --all (in the parent process) we should
+    // not load source from cache.
+    disableCache: !(this.cache && this.config.isChildProcess),
```
JaKXz: We should probably add a test for this statement

lib/source-maps.js
bcoe: 👍 will make more consistent usage of `const`.

Issue #515, opened 2017-02-06
document and test nyc merge functionality


bcoe:
_@addaleax brought up a couple enhancements to nyc, in a [conversation here](https://github.com/nodejs/node/pull/10856#issuecomment-277821130) that would make instrumentation of Node.js itself easier._

The first feature request was functionality similar to [istanbul-merge](https://www.npmjs.com/package/istanbul-merge).

nyc already does this internally, by outputting multiple coverage reports into a `.nyc_output` folder, and then merging them together while reporting. Having said this:

* we don't document an easy way to do this over multiple test runs (I believe this is what you're looking for essentially @addaleax?).
* we haven't benchmarked our approach vs. istanbul-merge (apparently istanbul-merge has great performance).

## Expected Behavior

An easy, fast, documented, way to collect together multiple coverage.json files into a single report.

## Help Wanted

@CurryKitten, @addaleax, do you have a repo we could work from while implementing this; also am I missing anything in this issue?

addaleax:
I think this is pretty much it, yes. Implementation-wise, one thing that I’ve considered is that it might actually make sense to spawn multiple subprocesses for reading the files and merging the coverage reports to have better CPU utilization. (Interestingly, most of the time here was spent running `JSON.parse`! :o)

> do you have a repo we could work from while implementing this

Not sure – I think code-wise pretty much everything we have is in the linked PR.

bcoe:
@yangli1990 this is similar to what you were looking for I believe.

JaKXz:
@ljharb I think we were having a conversation about this in a `react-dates` PR. If I understand this issue correctly this functionality is theoretically possible with `nyc`... but not much beyond theory is proven, I think.

ljharb:
Awesome! I'd love for `istanbul-merge` to no longer have to exist; I use it extensively in https://npmjs.com/mocha-wrap and https://npmjs.com/jest-wrap (where i need to run tests in mocha and jest respectively, but where i can't properly test the mocha/jest hooks unless i use tape).

boneskull:
Mocha could also use this.

boneskull:
sorry; didn't realize this issue is not "implement such a thing" because it already appears implemented.  just needs more tests (I found a problem; see #664) and docs.  otherwise, it seems to work!

Morikko:
For those like than struggled to understand the merge functionality, just some precisions.
There are:
 1. nyc merge (only output json)
 2. nyc report (do it implicitly)

[...]

PR #558, opened 2017-04-20
feat: add support for --no-clean, to disable deleting raw coverage output

bcoe:
This pull request is a follow on to #556, which starts to switch some of our non-integration tests over to mocha (it was driving me bonkers not having --grep functionality in our slow suites).

@addaleax since nyc's own test suite now requires merging multiple test runs into a single coverage report, I revisited https://github.com/istanbuljs/nyc/issues/515

The approach now outputs both the tap integration tests, and the mocha tests to the same temporary folder, and runs report once all tests have run.

Once this is landed, can we close out #515?

### DIFF
--- a/lib/config-util.js
+++ b/lib/config-util.js
@@ -219,6 +219,17 @@ Config.buildYargs = function (cwd) {
       type: 'boolean',
       global: false
     })
+    .option('clean', {
+      describe: 'should the .nyc_output folder be deleted before executing tests',
+      default: true,
+      type: 'boolean',
+      global: false
+    })
+    .option('temp-directory', {
+      describe: 'directory to output raw coverage data in',
+      default: './.nyc_output',
+      global: false
+    })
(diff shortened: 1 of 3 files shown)


lib/config-util.js
```diff
@@ -219,6 +219,17 @@ Config.buildYargs = function (cwd) {
       type: 'boolean',
       global: false
     })
+    .option('clean', {
+      describe: 'should the .nyc_output folder be cleaned before executing tests',
+      default: true,
```
JaKXz: I'm guessing this is because we want to just be explicit and start from a fresh state every time? There aren't really any perf benefits from keeping this folder around right?

bcoe: this allows us to keep the folder around between two runs of `nyc` -- which allows reports to be combined from mocha and tap -- as we start migrating a few tests over.

PR #578, opened 2017-05-24
Updated spawn-wrap to v1.3.5. Fixes #577.

### LINKED ISSUE #577: All spawn() "node" calls use current node version, not specified version
## Expected Behavior

When running `nyc` using Node 7.10.0, my code should be able to `spawn()` Node.js 6.9.5 executable.

## Observed Behavior

All spawned Node.js versions use whatever version is running `nyc`. In other words, the spawned subprocess always uses `process.execPath` instead of the original `filename` to execute.

```js
# version.js
console.log(process.version);
```

```js
# test.js
const spawnSync = require('child_process').spawnSync;
console.log(process.version);
spawnSync('/path/to/some/other/node', ['version.js'], { stdio: 'inherit' });
```

Assuming you have 2 different Node.js executables, the versions displayed should differ.

## Forensic Information

**Operating System:** macOS 10.12.5
**Environment Information:** Node.js 7.10.0 and 6.9.5. NPM version is irrelevant.

## Root Problem

The bug is in `spawn-wrap@1.2.4`, the version `nyc` uses. `spawn-wrap@1.2.4` is almost a year old. The latest version 1.3.5, which was published in May 2017, appears to fix this problem. Simply updating to `spawn-wrap@1.3.5` will resolve this ticket.

## Workaround

In my gulp file, instead of running `nyc` directly, I run the following bootstrap:

```js
# run-nyc.js
const fs = require('fs');
const Module = require('module');
const originalLoader = Module._extensions['.js'];
const spawnSwapRE = /\/spawn-wrap\/index\.js$/;

Module._extensions['.js'] = function (module, filename) {
	if (spawnSwapRE.test(filename)) {
		filename = require.resolve('spawn-wrap');
		module._compile(fs.readFileSync(filename, 'UTF-8'), filename);
	} else {
		originalLoader(module, filename);
	}
};

// remove this file from argv
process.argv.splice(1, 1);

// run nyc
require(process.argv[1]);
```

```js
const args = [
    'run-nyc.js',
    '/path/to/node_modules/.bin/nyc',
    ...
];
spawnSync(process.execPath, args, { stdio: 'inherit' });
```

## Note

I found another bug in `spawn-wrap` when `nyc` spawns `mocha`. `spawn-wrap` attempts to find out if the file being executed is Node.js so it can wrap it. In the case of `mocha`, it will detect it as a file with a shebang, namely `#!/usr/bin/env node`. The problem is `spawn-wrap` will only test the first part `#!/usr/bin/env` to see if it's Node and in this case it will fail and proceed to spawn `mocha` without wrapping it. To fix it, simply need to call `nyc node mocha <...>`. I've filed https://github.com/tapjs/spawn-wrap/issues/54 in hopes it may get resolved.

cb1kenobi:


bcoe:
@cb1kenobi this can't be landed currently, due to:

https://github.com/tapjs/spawn-wrap/issues/44

cb1kenobi:
@bcoe Well that's a bummer. I filed https://github.com/tapjs/spawn-wrap/issues/54, so I suppose I could take a crack at fixing https://github.com/tapjs/spawn-wrap/issues/44 too. Stay tuned.

cb1kenobi:
Now that `spawn-wrap@1.3.6` was just published and it fixes https://github.com/tapjs/spawn-wrap/issues/44, we should close this PR and whichever commit updates the `spawn-wrap` version in the `package.json` can mark https://github.com/istanbuljs/nyc/issues/577 as fixed.

PR #584, opened 2017-05-27
feat: upgrade to version of yargs with extends support

bcoe:
this introduces the slick new feature of `extends`, which will allow folks to publish preset configurations for nyc; see https://github.com/istanbuljs/istanbuljs/tree/master/packages/nyc-config-babel


bcoe:
@rcoy-v I've added docs for the `extends` functionality.

Issue #615, opened 2017-06-22
ENOENT: no such file or directory (nyc_cache)


garthenweb:
## Expected Behavior

NYC should not throw when installing with npm v5 as a dependency.  

## Observed Behavior

Throws errors while running:

```
ENOENT: no such file or directory, open '.nyc_cache/429c5318ec0ecfb4d0b65d8e90260fb2_https:/registry.npmjs.org/nyc/-/nyc-11.0.2.tgz.map'
```

I guess this is an issue on Windows because it's working with Travis: https://travis-ci.org/garthenweb/bubleify

## Bonus Points! Code (or Repository) that Reproduces Issue

https://github.com/garthenweb/bubleify/tree/chore/update-dependencies
Execute `npm install && npm test` in root folder with npm 5 (on windows).

## Forensic Information

**Operating System:** Windows 10 (Powershell/ Bash on Ubuntu on Windows)
**Environment Information:** node v8.1.2/ npm v5.0.3

novemberborn:
I can't reproduce this on macOS. Given your description in https://github.com/istanbuljs/nyc/pull/616:

> npm v5 transforms the version field from the simple number to an URL of
the tgz file in the package.json when installing nyc as a dependency.

This seems like buggy behavior on npm's part. It tends to add its own fields to `package.json` files using an underscore prefix.

garthenweb:
I just checked the behavior one more time and can still tell that it's the same issue after removing `node_modules` and reinstalling again (also with npm v5.0.4). Here is the package.json in `node_modules/nyc` folder in case it might be helpful:
``` json 
{
  "_args": [
    [
      "nyc@https://registry.npmjs.org/nyc/-/nyc-11.0.2.tgz",
      "C:\\Users\\janni\\projects\\npm-packages\\bubleify"
    ]
  ],
  "_from": "nyc@https://registry.npmjs.org/nyc/-/nyc-11.0.2.tgz",
  "_id": "nyc@11.0.2",
  "_inBundle": false,
  "_integrity": "sha1-nlkqaXGGAoJTtmhRbDjwecOcCPM=",
  "_location": "/nyc",
  "_phantomChildren": {},
  "_requested": {
    "type": "remote",
    "raw": "nyc@https://registry.npmjs.org/nyc/-/nyc-11.0.2.tgz",
    "name": "nyc",
    "escapedName": "nyc",
    "rawSpec": "https://registry.npmjs.org/nyc/-/nyc-11.0.2.tgz",
[...]

PR #616, opened 2017-06-22
fix: accept npm v5 package version

garthenweb:
npm v5 transforms the version field from the simple number to an URL of
the tgz file in the package.json when installing nyc as a dependency.
The version is appended to the hash for cache comparison which is also
used to create a cache directory on the filesystem. The URL may include
illegal character and the creation of the file fails.

Fixes https://github.com/istanbuljs/nyc/issues/615

### DIFF
--- a/lib/hash.js
+++ b/lib/hash.js
@@ -1,3 +1,4 @@
+const path = require('path')
 const CACHE_VERSION = require('../package.json').version
 const md5hex = require('md5-hex')
 const salt = JSON.stringify({
@@ -6,7 +7,7 @@ const salt = JSON.stringify({
 })
 
 function Hash (code, filename) {
-  return md5hex([code, filename, salt]) + '_' + CACHE_VERSION
+  return md5hex([code, filename, salt]) + '_' + path.basename(CACHE_VERSION)
 }
 
 Hash.salt = salt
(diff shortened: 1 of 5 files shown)


novemberborn:
https://github.com/istanbuljs/nyc/issues/615

bcoe:
@garthenweb this seems like a reasonable fix, but I'd like to loop in the CLI team and see if this is expected behavior -- one of the main folks who worked on the package cache is on vacation, I will ping them when they're back.

bcoe:
@zkat @iarna is the behavior @garthenweb is seeing withe the version field in npm 5 expected?

Speedy37:
why not putting the version field in the md5 hash data, this would be safer

zkat:
This is not the expected behavior on the latest versions of npm5, and was actually a bug when running npm5 on a previously-npm4-installed directory. The only (non-bug) reason for this to happen is when someone does an _explicit_ `npm install https://registry.npmjs.org/pkg/-/pkg-1.2.3.tgz` install. See [the `package-lock.json` spec](https://github.com/npm/npm/blob/latest/doc/spec/package-lock.md#version-changed)

I wouldn't be surprised if this sort of thing still happens in some corner cases, but it might be good to try and make sure you're running the latest npm before confirming.

bcoe:
@garthenweb sounds like this should be fixed in newer npms 👍 thanks for the feedback @zkat.

I'm going to close this issue, but please feel free to reopen if it turns out you're still bumping into issues; might be worth opening a tracking issue on npm, if this is the case.

Issue #542, opened 2017-03-29
support for --root option


borisovg:
nyc doesn't seem to have support for `--root` option?

I keep all my test stuff in ./tests (including separate package.json and all coverage output) and run istanbul from this directory with `--root ../` set.

bcoe:
@borisovg would definitely entertain adding this 👍

Speedy37:
While implementing the `--root` option, I found that the `NYC_CWD` environment variable does almost exactly that (it still looks upwards for package.json).

Rush:
I'm trying migrate from istanbul and I'm also missing this feature.

borisovg:
Closing since this was fixed by  #620

Rush:
@borisovg since it's closed does it mean it has been released?

borisovg:
@Rush I'm using `nyc` 11.2.1 and `--cwd` is listed in the help output (in the bug I linked to it was decided that it is sufficient replacement for `--root`. I tried it today with my code and it seems to work fine.

bcoe:
@Rush should be working, there are a few bugs being addressed however feel free to join the conversation here:

https://github.com/istanbuljs/nyc/issues/668

we seem to still have some issues when source-maps are remapping the directory structure.

Rush:
I confirm the option itself is working but unfortunately it's useless to me without ability to provide a custom coverage dir for `lcov`. Istanbul has an option --dir custom-coverage-dir, for example --dir coverage/api

PR #620, opened 2017-06-30
Add support for --cwd (#542)

Speedy37:
I added support for --root option
I found out that the `cwd` option does exactly what the --root is supposed to do except:

 - the NYC_CWD environment has no documentation
 - the implicit lookup for upward package.json is not something I would expect

So I added a --root option that simply force cwd to be what I tell it to be.

### DIFF
--- a/lib/config-util.js
+++ b/lib/config-util.js
@@ -41,7 +41,6 @@ Config.loadConfig = function (argv, cwd) {
   config.extension = arrify(config.extension)
   config.exclude = arrify(config.exclude)
   config.include = arrify(config.include)
-  config.cwd = cwd
 
   return config
 }
@@ -130,6 +129,10 @@ Config.buildYargs = function (cwd) {
       alias: 'n',
       default: [],
       describe: 'a list of specific files that should be covered, glob patterns are supported',
+      global: false
+    })
+    .option('cwd', {
+      describe: 'working directory to lookup for files',
       global: false
     })
     .option('require', {
--- a/test/src/nyc-tap.js
+++ b/test/src/nyc-tap.js
@@ -61,6 +61,11 @@ describe('nyc', function () {
       var nyc = new NYC(configUtil.loadConfig([], __dirname))
       nyc.cwd.should.eql(path.join(__dirname, '../..'))
     })
+
+    it('uses --cwd for cwd if it is set (highest priority and no look upwards for package.json) ', function () {
+      var nyc = new NYC(configUtil.loadConfig(['--cwd', __dirname], __dirname))
+      nyc.cwd.should.eql(__dirname)
+    })
   })
 
   describe('config', function () {


bcoe:
@Speedy37 rather than adding a new `root` option, I think I'd rather than we make `cwd` optionally configurable; basically, just renaming `root` to `cwd`.

Speedy37:
NYC_CWD isn't the same as nyc working directory. 
I think NYC_CWD isn't the right name for this, because it change the root path where it looks for files. The generated coverage file path are still relative to nyc working directory.

Speedy37:
I updated this PR to use cwd naming.

But I'm not convinced this is exactly what you want, because cwd doesn't only impact file lookup, it also impact cache location.

bcoe:
@Speedy37 maybe I'm just not fully understanding the problem you're looking to solve. Do you have a repo that benefits from manually setting the root directory, is the problem not addressed by `--cwd`?

Just want to make sure we're solving a specific issue you're running into before we land anything.

bcoe:
@Speedy37 this did the trick for you, for your use-case? will work on landing this soon, have just been pretty swamped triaging old issues (trying to cleanup the project a bit).

Speedy37:
Sry for the delay, I was on vacations.

The initial problem was to be able to run nyc by including files at /mypath/node_modules/mymodule/*.js while beeing able to reject /mypath/node_modules/mymodule/node_modules/**.

I was able to do this only by using the cwd trick because the 'node_modules/**' exclusion is deeply included by nyc. ('!node_modules/**' doesn't help because it prevents me to reject /mypath/node_modules/mymodule/node_modules/**).

The other problem being NYC_CWD modified by nyc by looking upwards for package.json, making it unreliable (I'm building my npm packages outside the source folder).

So being able to set --cwd would make the whole process much more reliable.

NYC_CWD has two main effect:

 - setting the root directory used by nyc to lookup for files
 - being the base of the temporary directory

In my opinion, two arguments (--root and --tmpdir) would be better than --cwd and their explanations would be more specific.

bcoe:
@Speedy37 as long as the default behavior is reasonable, I'd be comfortable with splitting configuration into two parts `--root` and `--tmpdir`.

I've been swamped with life/work stuff lately, and getting a bit buried in OSS work. would love to have you join the community slack I recently stood up:

it's a great way to kick me in the butt and make sure we get work over the finish line.

Speedy37:
After reviewing the whole list of arguments, you already define temp-directory and because you name cwd the base path for file lookup, I think --cwd is the best name (it won't be confusing with code implementation naming).

Speedy37:
Working like a charm

PowerSupply:
This does not seem to work. I can run my tests, but the coverage report is broken.

bcoe:
@Speedy37 have you had a chance to dig into this at all, I would like to make sure that if we add this feature it's working as expected.

@PowerSupply are you specifically having trouble with the `--cwd` feature?

Speedy37:
I've reverted my builds to version v11.1.0 due to #668, but from my tests the --cwd flag is working fine and with it I didn't had to clean old coverage reports (there is a check that disable cleaning .nyc_output if NYC_CWD is defined).

PowerSupply:
@bcoe yes and no.

My problem is that I want to use the --all flag to include "files that have not been required in your tests" (as per the documentation). But if I set the --all flag then all files in my project gets included although I have specified include to only include files in a certain folder.

Running the command from a subfolder does not seem to stifle the reach of --all, it still includes all files from my project root (where node_modules is).

The previous solution for this was to use the --root flag but it is no longer an option.

Therefore I tried to use --cwd (that is still undocumented in the readme) instead but it does not work. The combination of --cwd and --all does not work like --root and --all used to.

Right now I can see no other solution to cover all files in my "source" folder than to use the --all flag and then having to manually exclude all folders/files in my project that are not under the sources directory. 
This is a degradation from previous functionality.

Speedy37:
@PowerSupply without --cwd nyc will select cwd as the first folder in the path that is node_modules, if none then cwd is the current working directory.

--cwd and --all should work like --root and --all used to (I got it working without troubles)

What is the current behavior when you use --cwd ? If you have no instrumented file at all, see #668

PowerSupply:
Ok, so my problem was somehow related to using babel-plugin-istanbul in a way that I dont understand. I do not have time to look into the error right now but may investigate the problem later.

Thanks for your support!

PR #649, opened 2017-08-18
Remove excluded files from coverage before writing

### LINKED ISSUE #581: nyc trying to cover babel plugin even though it's excluded via config
In a project that uses `babel-istanbul` and `babel-register` AND a custom babel plugin, nyc instruments the custom plugin even though it's in a folder excluded from coverage via `.nycrc:exclude`.

## Expected Behavior

100% coverage on src/foobar.js and nothing else because nyc is set to [exclude](https://github.com/NiGhTTraX/nyc-exclude-bug/blob/adb30c42ae3991d359a8f758ec52743825630023/.nycrc#L6) the `tests/` folder which includes the [babel-plugin](https://github.com/NiGhTTraX/nyc-exclude-bug/blob/adb30c42ae3991d359a8f758ec52743825630023/tests/babel-plugin.js) defined in [.babelrc](https://github.com/NiGhTTraX/nyc-exclude-bug/blob/adb30c42ae3991d359a8f758ec52743825630023/.babelrc#L6).

## Observed Behavior

image

image

https://github.com/NiGhTTraX/nyc-exclude-bug

```
{
  "include": [
    "src/**/*.js"
  ],
  "exclude": [
    "tests/**/*.js"
  ],
  "all": true,

  "sourceMap": false,
  "instrument": false,

  "reporter": ["html"],
  "report-dir": "tests/unit/results/coverage",

  "lines": 100,
  "branches": 100,
  "check-coverage": true
}
```

## Forensic Information

**Operating System:** Ubuntu 16.04
**Environment Information:**

Node: v6.9.5
npm: 3.10.10

princjef:
Fixes #581 by scrubbing the coverage object to remove any excluded files for which coverage was generated.

I'll admit to not being entirely sure how the coverage gets in for those files in the first place (the exclude logic appears to be running correctly in `addAllFiles()`), but this provides a catch-all at the end of processing to clean it up.

There's also a question about whether the scrub should occur before or after following source maps. Doing it before seems to make the most sense based on how things work now.

### DIFF
--- a/index.js
+++ b/index.js
@@ -367,6 +367,13 @@ NYC.prototype.writeCoverageFile = function () {
   var coverage = coverageFinder()
   if (!coverage) return
 
+  // Remove any files that should be excluded but snuck into the coverage
+  Object.keys(coverage).forEach(function (absFile) {
+    if (!this.exclude.shouldInstrument(absFile)) {
+      delete coverage[absFile]
+    }
+  }, this)
+
   if (this.cache) {
     Object.keys(coverage).forEach(function (absFile) {
       if (this.hashCache[absFile] && coverage[absFile]) {


bcoe:
@princjef very odd! seems like a reasonable safety to put in place though. I'll work on landing this as soon as possible.

Worth noting, I stood up a community slack for coordinating work on this project if you feel like helping coordinate work around getting another release of nyc out:

bcoe:
@princjef I've released this for you in the latest version of `nyc`, mind testing things out? thanks for the contribution 🎉

Speedy37:
This broke my builds, because I'm using NYC_CWD / --cwd to bypass **/node_modules/** default exclusion in my tests base path, but I still want to exclude **/node_modules/** after it

ie: I want /mybase_path/node_modules/mymodules/**
but not /mybase_path/node_modules/mymodules/node_modules/**

And this can't be expressed with -x -i even with -x "!**/node_modules/**" due to the way they are resolved

Speedy37:
I spoke to soon, it is commit dd40dc5 that broke my build

PR #672, opened 2017-09-09
Caching merge tweak

cindyywssj:


bcoe:
@cindyywssj can you provide a bit of context to this pull, I think we've already landed a lot of these changes in some for or another.

PR #667, opened 2017-09-06
fix: Do not exclude files with sourcemap from coverage

vanduynslagerp:
Fix #666

PR #674, opened 2017-09-12
Create sourcemap in instrumented files

schutm:
If invoked with `nyc instrument` it wasn's possible to attach the sourcemap. This created problems for me in the coverage reports created.

### DIFF
--- a/lib/commands/instrument.js
+++ b/lib/commands/instrument.js
@@ -27,6 +27,11 @@ exports.builder = function (yargs) {
       type: 'boolean',
       description: 'should nyc detect and handle source maps?'
     })
+    .option('produce-source-map', {
+      default: false,
+      type: 'boolean',
+      description: "should nyc's instrumenter produce source maps?"
+    })
     .option('instrument', {
       default: true,
       type: 'boolean',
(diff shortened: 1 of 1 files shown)


bcoe:
@schutm I'm not sure what the best solution is here such that you and @vanduynslagerp don't consistently break each-other's use-cases.

Is there a heuristic approach we could use so that both your applications of `shouldInstrument` work? I think we should make an effort to add a test for both so that we don't keep flip-flopping on this feature.

schutm:
Agreed we shouldn't break each other changes, and test-cases should be there. However since this is in the CLI I didn't think it would break the previous changes.

For now seems I had a dangling comma; which I'll fix

bcoe:
@schutm I think this accidentally reverts https://github.com/istanbuljs/nyc/pull/667/files which would break @vanduynslagerp's use-case.

schutm:
@bcoe I removed the accidentally included index.js, which reverted #667. In addition I tidied the commits up to a single one.

bcoe:
@schutm this looks great 👍 sorry that this took so long to land.

bcoe:
@schutm try this out, and let me know if it does the trick:

`npm i nyc@next --save-dev`

PR #683, opened 2017-10-03
Only apps should have lockfiles

ljharb:
This prevents package-lock.json from being created locally. It's already gitignored, but local devs won't have an accurate experience when the lockfile exists.

This fixes that.


bcoe:
@ljharb thanks for the contribution.

PR #695, opened 2017-10-18
docs: missing options prefix -- in front of check-coverage

clakech:


bcoe:
@clakech thanks for the fix.

PR #696, opened 2017-10-21
External sourcemap fix

thegecko:
This PR contains two fixes:

- a fix for external sourcemaps which weren't being loaded properly
- added a missing callback

### DIFF
--- a/index.js
+++ b/index.js
@@ -230,7 +230,7 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
   } catch (err) {
     return cb(err)
   }
-  cb();
+  cb()
 }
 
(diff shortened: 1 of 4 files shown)


thegecko:
I think a similar fix is needed for picking up the correct files when creating a report. I am on holiday now, bit will investigate upon my return and add another PR then.

thegecko:
@bcoe I can see #697 has gone some way to fixing this issue. There are still a few minor fixes in this PR which I'll pull out into a fresh one.

bcoe:
@thegecko awesome 👍 I'm glad we're starting to get things sorted out better for folks who pre-instrument their files.

PR #697, opened 2017-10-23
feat: add --exclude-after-remap option for users who pre-instrument their codebase

bcoe:
## The Problem Being Addressed

If someone is running tests against files that have been pre-instrumented in a codebase with corresponding source-maps (rather than using a JIT transpiler like babel-register) the a source-map will sometimes result in a source-files path being remapped to a folder that is covered by an exclude rule.

A user can now disable this behavior by setting `--exclude-after-remap` to `false`, which will stop exclude logic from being applied after the application of `source-maps`.

## In This Pull Request

* added/updated docs:
  * adding a note about the `--exclude-after-remap` caveat.
  * documenting istanbul ignore rules (CC: @kentcdodds, I would love your 👀 on this since you just added a new exclude rule).
* added a section to the test-suite specifically for source-maps (which are a common source of headaches for people, let's try to get more tests in place as people bump into issues).
* I also downgraded `spawn-wrap` because it was creating `undefined` folders while I attempted to test things; [I've filed an issue](https://github.com/tapjs/spawn-wrap/issues/64).

_@thegecko  I think there's a good chance that the behavior around source-map remapping was the root cause of your issues. The more I think about #696, I don't think there's any reason why we should need to use the relative path rather than the absolute path to the file ... there's a chance however this might have stepped around the issues described in this pull request._

@ljharb perhaps you'd like to take a quick look at this pull too -- I forget, I think you might have been bumping into something related to this?


README.md
```diff
@@ -320,9 +331,18 @@ You can specify custom high and low watermarks in nyc's configuration:
 }
 ```
 
-## Other advanced features
+## Parsing Hints (Ignoring Lines)
```
kentcdodds: This whole section looks good to me 👍

package.json
```diff
@@ -96,7 +96,7 @@
     "resolve-from": "^2.0.0",
     "rimraf": "^2.5.4",
     "signal-exit": "^3.0.1",
-    "spawn-wrap": "^1.4.0",
+    "spawn-wrap": "1.3.8",
```
ljharb: i've noticed some weirdness around pinning versions in npm 5; using `=1.3.8` seems to avoid it

thegecko:
I can confirm this fixed issues I was seeing with folders being specified for instrumentation/instrumentering

package.json
rumpl: Is there a reason `spawn-wrap` was downgraded? There is a fix in version `1.4.0` that can be useful for people that want to run nyc in a container.

PR #703, opened 2017-10-29
Instrumenter test and callback

thegecko:
@bcoe As #697 fixed the source-map issues I was looking to fix in #696, this PR only contains the other minor fixes included in that PR.

Specifically:

- Added a test to prove sub-directories can be instrumented
- Added a missing callback call and exit code to ensure all tests still pass

### DIFF
--- a/test/nyc-bin.js
+++ b/test/nyc-bin.js
@@ -479,6 +479,23 @@ describe('the nyc cli', function () {
           done()
         })
       })
+
+      it('allows a sub-directory of files to be instrumented', function (done) {
+        var args = [bin, 'instrument', './subdir/input-dir', './output']
+
+        var proc = spawn(process.execPath, args, {
+          cwd: fixturesCLI,
+          env: env
+        })
+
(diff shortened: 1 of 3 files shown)


test/nyc-bin.js
```diff
@@ -479,6 +479,23 @@ describe('the nyc cli', function () {
           done()
         })
       })
+
+      it('allows a sub-directory of files to be instrumented', function (done) {
```
bcoe: awesome \o/ thank you for the test. random aside, I think these integration tests are probably the approach I should have taken all along with nyc, I think I might start retiring some of the older crufty test files that are kicking around.

bcoe:
@thegecko your changes are set to be released in `11.4.0`; would love your help testing:

`npm i nyc@next` -- (you should see `11.4.0` install).

thegecko:
Confirmed!
```
  "dependencies": {
    "nyc": "^11.4.0"
  }
```

All changes LGTM

PR #715, opened 2017-11-07
docs: correct check-coverage shorthand example

alexframe:
The example does not correctly show the shorthand usage, this PR corrects that.


bcoe:
@alexframe thanks for the contribution.

PR #716, opened 2017-11-07
chore: build and test against new node8 (LTS)

alexframe:
Now that [Node 8 is LTS](https://github.com/nodejs/Release), I've added the it to appveyor and travis build configurations.

This makes sure that all LTS versions of node are explicitly being tested against:

- 4 (Maintenance LTS)
- 6 (Active LTS)
- 8 (Active LTS)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 node_js:
+  - 8
   - 6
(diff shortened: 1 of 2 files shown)


alexframe:
Need to change npm command for node8+, will take a look in my fork and open another PR when resolved.

PR #724, opened 2017-11-17
Add a new argument to specify a different path for nycrc

igorlima:
This PR allows us to specify a different path for `.nycrc`.

> NOTE: Some projects need different settings for _backend_ and _frontend_ tests. I'm working on a project that we do need different settings for both _backend_ and _frontend_ tests.

> Also [mocha](https://mochajs.org/) allows us to do it as well, by providing the argument `--opts`.

Thanks for understanding.

### DIFF
--- a/lib/config-util.js
+++ b/lib/config-util.js
@@ -18,8 +18,8 @@ function guessCWD (cwd) {
   return cwd
 }
 
-function loadConfig (argv, cwd) {
-  const rcPath = findUp.sync([argv.nycrcPath || '.nycrc', '.nycrc.json'], {cwd: cwd})
+Config.loadConfig = function (argv, cwd) {
+  const rcPath = findUp.sync([argv.nycrcPath || '.nycrc', '.nycrc.json'], {cwd: guessCWD(cwd)})
   let config = {}
 
   if (rcPath) {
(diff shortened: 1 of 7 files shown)


pedrocelso:
LGTM

tbonfim:
LGTM

lib/config-util.js
```diff
@@ -40,7 +40,11 @@ function loadConfig (argv, cwd) {
 // that would cause the application to exit early.
 Config.buildYargs = function (cwd) {
   cwd = guessCWD(cwd)
-  const config = loadConfig()
+  const argv = Yargs(process.argv)
+    .help(false)
```
bcoe: I think we could avoid needing to call `Yargs(process.argv)` here if we delay calling this line:

https://github.com/igorlima/nyc/blob/486aacc1ba1702c5f10505664cdefaa7368703ef/lib/config-util.js#L220

I think we could then swap out the config path in the `nyc.js` bin:

https://github.com/igorlima/nyc/blob/486aacc1ba1702c5f10505664cdefaa7368703ef/bin/nyc.js#L21

igorlima: @bcoe thanks for the quick _feedback_.

Done. Addressed already at https://github.com/istanbuljs/nyc/pull/724/commits/b01a3f34884ad3fb6c1a73b200ee7e0000443c6a.

- [x] avoid needing to call `Yargs(process.argv)`
- [x] swap out the config path in the `bin/nyc.js`

> _Let me know if there's any other concern._

igorlima:
@bcoe let me know if there's any other concern. Thanks.

bcoe:
@igorlima thank you for the contribution 😄

astorije:
Awesome stuff, @igorlima! @bcoe, when can we expect it in a release? :)) Thanks all!

bcoe:
@astorije @igorlima please try this:

`npm i nyc@next` version `11.4.0` is in the queue, and will be the next release.

astorije:
I'm using the `master` version right now, but will switch to `@next` until 11.4.0 gets out! Thanks!

PR #730, opened 2017-11-27
chore: upgrading to newer spawn-wrap

bcoe:
CC: @chadxz, @gautaz


JaKXz:
to be pedantic @bcoe I think this should be listed as a `fix: upgrade spawn-wrap` in the changelog since spawn-wrap is a production dependency

PR #758, opened 2017-12-27
Fix typo

hkdnet:
https://github.com/yargs/yargs/blob/master/docs/api.md#global

> Options default to being global.

This patch changes `reporter` from global to not global.
It may break something. However, the broken behavior is not intended.


bcoe:
@hkdnet thanks for catching this; always appreciate contributions.

bcoe:
@hkdnet please try `npm i nyc@next` which should have your fix; let me know if you run into any issues.

PR #768, opened 2018-01-23
chore: remove greenkeeper ignore for find-up

realityking:
Originally this was added in #402 but find-up has since been updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,10 +156,5 @@
     "test-exclude",
     "yargs",
     "yargs-parser"
-  ],
-  "greenkeeper": {
-    "ignore": [
-      "find-up"
-    ]
-  }
+  ]
 }


realityking:
Also looks like greenkeeper has been disabled - all the more reason to remove it.

JaKXz:
My opinion is that greenkeeper might become useful again, so this config isn't hurting us. My understanding is that we've disabled greenkeeper temporarily to try to reduce the issue/PR noise while we triage everything and try to get more issues closed/bugs fixed/PRs merged.

realityking:
@JaKXz I agree but since `find-up` has already been updated ignoring it in the Greenkeeper config is actually harmful if you decide to use Greenkeeper again.

bcoe:
@realityking @JaKXz I'm fine with removing this config -- I believe it's an artifact from wanting to be backwards compatible to Node 0.10.

PR #777, opened 2018-02-12
add new --skip-empty option and pass options to reports

moos:
In conjunction with https://github.com/istanbuljs/istanbuljs/pull/140

### DIFF
--- a/index.js
+++ b/index.js
@@ -447,11 +447,12 @@ NYC.prototype.report = function () {
     dir: this.reportDirectory(),
     watermarks: this.config.watermarks
   })
+  var config = this.config
 
   tree = libReport.summarizers.pkg(map)
 
   this.reporter.forEach(function (_reporter) {
-    tree.visit(reports.create(_reporter), context)
+    tree.visit(reports.create(_reporter, config), context)
   })
 
   if (this._showProcessTree) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -447,11 +447,12 @@ NYC.prototype.report = function () {
     dir: this.reportDirectory(),
     watermarks: this.config.watermarks
   })
+  var config = this.config
 
   tree = libReport.summarizers.pkg(map)
 
   this.reporter.forEach(function (_reporter) {
-    tree.visit(reports.create(_reporter), context)
+    tree.visit(reports.create(_reporter, config), context)
```
bcoe: I think it would be better to pass in the specific config option we care about:

```js
{
  skipEmpty: this.config.skipEmpty
}
```

index.js
```diff
@@ -447,11 +447,12 @@ NYC.prototype.report = function () {
     dir: this.reportDirectory(),
     watermarks: this.config.watermarks
   })
+  var config = this.config
```
bcoe: let's just switch to:

```js
this.reporter.forEach((_reporter) => {

}
```

then we don't need the temporary variable.

bcoe:
@moos I've released your work in `nyc@next`; mind trying it out and letting me know if things work as expected.

`npm i nyc@next`

👍 thanks for the contribution.

PR #787, opened 2018-02-26
Removes split-lines dependency.

philnash:
The dependency is not used, so removing it from package.json.


bcoe:
@philnash good catch, not sure how this snuck in.

philnash:
Heh, it's hard to miss an extra dependency! I just happened to be looking at the right thing. Thanks!

PR #860, opened 2018-06-07
Update README.md with babel configuration info

dmwelch:
The current documentation implies that ONLY the environment named `test` is valid for coverage reporting using Babel.  In fact, multiple environments can be used (and should be) if the tests require it.


bcoe:
@dmwelch thank you for the contribution.

Issue #796, opened 2018-03-20
files using `package` word are not included in the coverage report


kevinrambaud:
## Expected Behavior

I expect `nyc` to cover all the files that are required by any modules that are tested by mocha.

## Observed Behavior

When using the `package` word into a file, either as a variable name of a function name, `nyc` just ignore the file and does not report anything for it.

## Bonus Points! Code (or Repository) that Reproduces Issue

https://github.com/kevinrambaud/nyc-package-word-empty-coverage-example

## Forensic Information

**Operating System:** the operating system you observed the issue on.
**Environment Information:** information about your project's environment, see instructions below:

1. run the following script:

sh -c 'node --version; npm --version; npm ls' > output.txt

2. share a gist with the contents of output.txt.

https://gist.github.com/kevinrambaud/400048f2bef65a061e6a2859613a23bc

coreyfarrell:
@kevinrambaud I've posted a PR to add an option to allow this, but `package` is a reserved word.  Be aware the following script will fail to run:
```js
'use strict';
function package() {
  return 'package is reserved in strict mode';
}
console.log(package());
```
The addition of `'use strict'` causes node.js to refuse the script due to `package` being reserved.

isaacs:
This just popped up again, with the `interface` keyword.

I think there should never be a case where nyc both (a) doesn't cover a file because of a transpilation error, and (b) says nothing about it.  At the very least, a warning should be printed to stderr that coverage could not be added because of `${reason}`.

https://github.com/tapjs/node-tap/issues/527

PR #863, opened 2018-06-09
fix: Add flag to allow instrumentation of non-strict scripts.

coreyfarrell:
nyc was hard-coded to use `esModules: true` when running the
instrumenter.  This could result in certain scripts not being
instrumented.

```js
function package() {
  return 'package is a reserved keyword'
}

console.log(package())
```

Using `--es-modules false` will allow this script to be instrumented.

Issue #796

### DIFF
--- a/lib/config-util.js
+++ b/lib/config-util.js
@@ -115,6 +115,12 @@ Config.buildYargs = function (cwd) {
       describe: 'cache babel transpilation results for improved performance',
       global: false
     })
+    .option('es-modules', {
+      default: false,
+      type: 'boolean',
+      describe: 'tell the instrumenter to treat files as ES Modules',
+      global: false
+    })
     .option('extension', {
       alias: 'e',
       default: [],
--- a/index.js
+++ b/index.js
@@ -141,7 +141,8 @@ NYC.prototype._createInstrumenter = function () {
     ignoreClassMethods: [].concat(this.config.ignoreClassMethod).filter(a => a),
     produceSourceMap: this.config.produceSourceMap,
     compact: this.config.compact,
-    preserveComments: this.config.preserveComments
+    preserveComments: this.config.preserveComments,
+    esModules: this.config.esModules
   })
 }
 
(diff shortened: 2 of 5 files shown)


lib/config-util.js
```diff
@@ -115,6 +115,12 @@ Config.buildYargs = function (cwd) {
       describe: 'cache babel transpilation results for improved performance',
       global: false
     })
+    .option('es-modules', {
+      default: true,
+      type: 'boolean',
+      describe: 'Tell the instrumenter to treat files as ES Modules.',
```
bcoe: no capital on `Tell` or `.` needed, to match style of other options.

lib/config-util.js
```diff
@@ -115,6 +115,12 @@ Config.buildYargs = function (cwd) {
       describe: 'cache babel transpilation results for improved performance',
       global: false
     })
+    .option('es-modules', {
+      default: true,
```
bcoe: Is there any reason to not default this to `false`, I guess I'm not fully understanding what this flag buys us -- since ES-modules aren't supported in vanilla JS yet.

coreyfarrell: I set the default true to match the existing behavior.  The biggest effect is that it puts the Babel parser into strict mode.

I can default to false if you prefer?

bcoe: If it makes us less sensitive, without sacrificing our ability to instrument anything, my temptation would be to default it to `false` ... what do you think?

perhaps someone from the babel project can chime in, CC: @loganfsmyth

coreyfarrell: A couple relevant details:
1. The istanbul-lib-instrument default for esModules is false.
2. This is translated into the babelrc option sourcetype `'module'` or `'script'`.  The default for babel 6 and 7 is `sourceType: 'module'`.

So honestly I think `istanbul-lib-instrument` should also have `esModules: true` by default so that not specifying any option causes babel to behave as normal.  I also understand the argument for being less restrictive, I don't really see this as a huge issue either way.

https://babeljs.io/docs/en/babel-core/#options
https://babeljs.io/docs/en/next/babel-core#options

loganfsmyth: Personally I think the JS ecosystem is in a bit of an unfortunate place here, since there's simply no way to 100% guarantee that you have the type of a file right without additional data. That said, in the context of `nyc`, and now that #863 has landed, it probably makes the most sense for `nyc` to default to `unambiguous` which you'll see in [the `next` docs]( babeljs.io/docs/en/next/babel-core#options). It essentially means that the parser will guess the type.

Does Istanbul actually use the strictness of code the do anything during instrumentation? I'm assuming it doesn't. If that's the case, guessing wrong doesn't actually affect anything in Istanbul and just makes things more likely to parse.

coreyfarrell: @bcoe Set to false and updated the tests.

PR #866, opened 2018-06-18
chore: update dependencies

LinusU:
This PR updates both `caching-transform`, `find-cache-dir` and `find-up` to new major versions. The breaking changes in all three of them are that they now require at least Node.js 6.x. Since `nyc` already requires Node.js 6.x, this shouldn't affect this module.


bcoe:
@LinusU thank you for the contribution.

Issue #869, opened 2018-06-20
Running nyc on a FIPS enabled system fails due to use of weak md5 algorithm.


kresss:
## Expected Behavior
nyc should work on FIPS enabled systems.

## Observed Behavior
nyc fails.

```
$ ./node_modules/.bin/nyc ./node_modules/.bin/mocha "test/init-test-env.js" "test/unit/**/*.js"
crypto.js:81
  this._handle = new binding.Hash(algorithm);

Error: error:060800A3:digital envelope routines:EVP_DigestInit_ex:disabled for fips
    at new Hash (crypto.js:81:18)
    at Object.createHash (crypto.js:611:10)
    at module.exports (/home/kresss/GIT/wsm/mcs/node_modules/nyc/node_modules/md5-hex/index.js:5:20)
    at NYC.generateUniqueID (/home/kresss/GIT/wsm/mcs/node_modules/nyc/index.js:373:10)
    at new NYC (/home/kresss/GIT/wsm/mcs/node_modules/nyc/index.js:88:47)
    at Object.<anonymous> (/home/kresss/GIT/wsm/mcs/node_modules/nyc/bin/nyc.js:36:14)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
```

## Forensic Information

RHEL 7.5 with FIPS enabled at the kernel level. 
Node v8.9.4
npm v5.6.0

kresss:
I made an attempt at fixing this. See #870

PR #870, opened 2018-06-20
Use sha256 instead of md5. (#869)

kresss:
MD5 is not FIPS compliant and makes nyc inoperable on FIPS enabled systems. 

The md5-hex library that was used for creating hashes uses the node crypto library unless it is run in a browser. I don't think that nyc has a browser use case so I removed the extra dependency of the md5-hex module and made calls directly to node's crypto library. 

I arbitrarily picked sha256 as the new hashing algorithm. It is FIPS compliant and is expected to remain so for a while. sha1 could be used if performance is a large concern.


bcoe:
@kresss this seems reasonable to me, very much appreciate the contribution -- will let @addaleax chime in with a second set of eyes.

kresss:
@coreyfarrell I can take a look into it later tonight. Do you know if that project has code coverage and linting like nyc? As far as caching-transform#4 goes, I would probably ask from some help from someone with better domain knowledge. If I get into it tonight I will create an issue under caching-transform for moving from md5.

Thank you all.

coreyfarrell:
Don't worry about caching-transform#4, I'll handle that.  That module has linting and testing, the JavaScript code style is different but `npm test` will let you know if your changes pose any problems.

lib/hash.js
```diff
@@ -1,12 +1,18 @@
 const CACHE_VERSION = require('../package.json').version
-const md5hex = require('md5-hex')
+const crypto = require('crypto')
 const salt = JSON.stringify({
   istanbul: require('istanbul-lib-coverage/package.json').version,
   nyc: CACHE_VERSION
 })
 
 function Hash (code, filename) {
-  return md5hex([code, filename, salt]) + '_' + CACHE_VERSION
```
coreyfarrell: I've looked into this further, the `Hash` function is never actually run.  The only reference to `lib/hash.js` is in `index.js` from `NYC.prototype._createTransform`.  caching-transform does not use `opts.hash` so I think we should remove it.  `opts.salt` is used so I think `lib/hash.js` should `module.exports = {salt}`, delete the Hash function completely.

index.js
```diff
@@ -366,9 +366,13 @@ NYC.prototype.wrap = function (bin) {
 }
 
 NYC.prototype.generateUniqueID = function () {
-  return md5hex(
-    process.hrtime().concat(process.pid).map(String)
-  )
+  const hash = crypto.createHash('sha256')
```
coreyfarrell: This is valid and I don't have an issue with it, just have to ask what about using the `uuid` module instead?  That might actually be more correct for this purpose but I'm not sure it's too important.

bcoe: @coreyfarrell @addaleax great point, in this context it seems like using `uuid.v4()` would be reasonable? this doesn't solve the issue with caching-transform performance, which I think is worth discussing further in that thread.

lib/hash.js
coreyfarrell: I see what happened, this is unused due to the update from caching-transform@1 to caching-transform@2.  I'm working on a patch to address this part correctly, the Hash function isn't entirely unneeded, it needs to be replaced with other code.

bcoe:
@kresss, perhaps you and @coreyfarrell can work together to see this over the finish line? sounds like there's a little bit of juggling required to retire the Hash function ... and I agree with @coreyfarrell that we should use a `uuid.v4()` rather than a hash if it's actually a UUID we want.

http://devtoolscommunity.herokuapp.com/ 👈 there's a slack where we can all chat.

coreyfarrell:
@kresss I've posted #883 which uses uuid instead of hashing.  That patch should make it possible to run `nyc --cache false` on FIPS systems.  Note depending on your tests disabling nyc cache could make it run much slower.  Once we get caching-transform updated it will be unnecessary to disable caching.

kresss:
@coreyfarrell thanks for the update. Sorry for the silence. I went on holiday. The uuid solution sounds good. I will pull your branch and do some testing today on a FIPS system. If I have any issues I will report back on #883 

Thank you.

coreyfarrell:
Since #883 we no longer directly use md5-hex from nyc, so I'm closing this PR as nothing needs to be done here.  You should be able to use the next release of nyc with `--cache false` until the related issue in caching-transform and dependency is resolved.

PR #872, opened 2018-06-27
chore: update dependencies

LinusU:
Getting closer to remove all copies of p-try 1.x, p-limit 1.x, find-up 2.x, etc:

screen shot 2018-06-27 at 12 10 25

The only breaking change is dropped support for Node.js 4.x which should be cool 😎


coreyfarrell:
Thanks for the patch @LinusU.

PR #873, opened 2018-06-27
fix: Update caching-transform options.

coreyfarrell:
When updating caching--transform it was missed that the `hash` option
has been refactored.  This means that two files with identical contents
would be treated as identical.

Add a test to verify that two sources with identical contents are
both instrumented.

### DIFF
--- a/test/fixtures/identical-file-runner.js
+++ b/test/fixtures/identical-file-runner.js
@@ -1,0 +1,14 @@
+const path = require('path')
+const assert = require('assert')
+const file1 = require('./identical-file1.js')
+const file2 = require('./identical-file2.js')
+
+assert.equal(file1(), file2())
+
+const cov = (new Function('return this.__coverage__'))()
+
+assert.deepEqual(Object.keys(cov).sort(), [
+  __filename,
+  path.resolve('identical-file1.js'),
+  path.resolve('identical-file2.js')
+])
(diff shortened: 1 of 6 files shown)


test/fixtures/identical-file-runner.js
```diff
@@ -0,0 +1,14 @@
+const path = require('path')
+const assert = require('assert')
+const file1 = require('./identical-file1.js')
+const file2 = require('./identical-file2.js')
+
+assert.equal(file1(), file2())
+
+const cov = (new Function('return this.__coverage__'))()
```
coreyfarrell: I'm open to better ways to test that the coverage data includes the 3 files.

bcoe: @coreyfarrell this assertion seems reasonable to me _but_, out of curiosity, what would the behavior have been before your fixes to the caching transform, would we have only had one of the two files listed in the coverage object?

mind testing running this test prior to your fix and confirming that it breaks as expected?

coreyfarrell: I did test it, without the fixes to caching-transform this test would not record results for identical-file2.js.  This is because both identical files have the same hash value so caching-transform gave the instrumented code from file1 to file2.  This meant file2 recorded it's results to file1.

Issue #890, opened 2018-07-16
Instrumenter should ignore excluded files and folders


AndrewFinlay:
## Expected Behavior
The instrumenter should ignore files specified in the list of exclude globs

`$ nyc instrument content instrumented `

with .nycrc
```
{
    "exclude": [
        "**/dist/**",
        "**/3rd-party/**"
    ]
}
```

## Observed Behavior
The instrumenter ignores files specified in the defaultExclude instead, (istanbuljs/packages/test-exclude/index.js)
## Bonus Points! Code (or Repository) that Reproduces Issue
[https://github.com/AndrewFinlay/nyc-instrument-exclude-example](https://github.com/AndrewFinlay/nyc-instrument-exclude-example)
## Forensic Information
Operating System:  macOS High Sierra
Environment Information:
node v8.11.1
npm v6.1.0
├── angular-mocks@1.4.9
├── findup-sync@2.0.0
├── istanbul@0.4.5
├── jest-cli@23.0.0
├── jest-localstorage-mock@2.2.0
├── jsdom@11.11.0
├── lodash@4.17.10
├── nyc@13.0.0
├── protractor-istanbul-plugin@2.0.0
├── protractor-jasmine2-screenshot-reporter@0.5.0
├── sinon@5.0.10
├── slug@0.9.1
└── source-map-support@0.5.6

After doing some digging it seems that the nyc instrumenter simply isn't using the include/exclude parameters from the config.  I'm pretty certain that it's the same issue referenced in #596, and related to #573

When NYC is configured in "commands/instrument.js" it excludes any configuration setup in "config-util.js".  This means that it will fallback onto the defaults set in "istanbul/test-exclude/index.js".

The good news is it looks like a pretty easy fix, the bad news is I'm assuming that specifying exclude/include paths in the instrument code is desired behaviour.  So I have no idea what would break by changing this.  It may be best to play it safe and add a pair of new config options, instrumentInclude and instrumentExclude.

Let me know what you think of this PR #893.

felixfbecker:
I think I am seeing the same issue. I am trying to `exclude` test files because the instrumentation breaks Puppeteer's `page.evaluate()` (because it calls `function.toString()` and sends it to the browser to evaluate, which is missing the coverage globals). `nyc` is properly ignoring the files from the final report, but they are still being instrumented, which breaks Puppeteer. `/* istanbul ignore file */` at the top of the file doesn't work either.

AndrewFinlay:
Bump, stale bot

Issue #596, opened 2017-06-06
Globs are not working for $ istanbul instrument -x "**glob**"


ORESoftware:
I am trying to do this:

```
public
    /lib     # this is library code (angular etc.), I want to ignore this dir
    /pages   # this is my code
```
    
Here is the command I use:

`istanbul instrument public --output public-coverage  -x "public/lib/**/*"`

Unfortunately, the public/lib directory still gets instrumented, which is a bummer because it contains a ton of files that are 3rd party libraries which I really do not want to instrument.

I have tried every glob combination under the sun and it's not working - the lib directory and its contents never get ignored, and they are always copied over to public-coverage and instrumented.

I consistently have problems with getting globs to behave as desired...it's not just this library. Not sure why that is, but maybe people should consider other ways of ignoring dirs, etc.

ORESoftware:
bump pls

JaKXz:
@ORESoftware I think you're using the wrong executable, but you can configure `nyc` to ignore directories like this: https://github.com/istanbuljs/nyc#excluding-files

Please share your repo or a sample reproduction with `nyc`.  I'd go as far as to unofficially say that the `istanbul` CLI is no longer supported, since `nyc` has replaced it. You can see the docs on the `instrument` command here: https://istanbul.js.org/docs/advanced/instrument/

ORESoftware:
Thanks! I wonder why nyc didn't choose to stick with original CLI:

before:

`istanbul instrument ./lib -o ./output`

after?

`nyc instrument ./lib ./output`

ORESoftware:
Unfortunately this doesn't work:

```
"nyc": {
    "exclude": [
      "public/lib/**/*"
    ]
  }
```

nyc still attempts to instrument the public/lib contents

this is driving me INSANE

bcoe:
@ORESoftware could you please provide a link to a repository that is running into this issue?

Here's a unit test we have for logic very similar to what you're attempting:

https://github.com/istanbuljs/istanbuljs/blob/master/packages/test-exclude/test/test-exclude.js#L219

We're not doing anything exciting with our `test-exclude` logic, it simply runs a glob with a few nice defaults applied, e.g., always excluding `node_modules`.

If you are having trouble excluding `/public/lib` my guess is that there's something else odd happening, such as the path getting munged at some point during the instrumentation process.

coreyfarrell:
Fix being worked on in #1007

Issue #309, opened 2016-07-11
Brain dump of istanbul - nyc deltas


gotwarlost:
Differences and gaps in `istanbul` versus `nyc` as a brainstorm doc. We should create individual issues for these in a second pass and tag them as `now` versus `later` for the ones for which something needs to be done.

In general, the highest priority is for gaps that affect node.js projects. Browser use-cases are second class citizens for the first release.
## istanbul configuration deltas

Type `istanbul help config` for full list. We posit a migration script that can turn a `istanbul.yaml` file to an `nyc` config.
### instrumentation (instrument and cover commands)
1. instrumentation root: root below which to instrument files (default: process.cwd, I think this is _rarely_ used)
2. default-excludes [bool]: exclude `node_modules/*`, `* test/**`, `tests/**` (default: true, false means consider everything)
3. compact [bool] - emit compact code (default: true).
4. preserve-comments [bool] whether to preserve comments (default: false)
5. save-baseline [bool] - whether the `istanbul instrument` command should save a baseline coverage file
#### Conclusions

default-excludes MUST be solved. This is a BIG diff for people migrating from istanbul to nyc. The simplest approach is to let the migration script handle it by adding appropriate -x options in the nyc config.  This doesn't pollute the nyc codebase and provides a backwards-compatible migration path.

No need to allow files from under `node_modules` to be instrumented for now (although people in istanbul-land have specifically wanted this).

(1), (3) and (4) can be opinions in nyc for now until we find valid use-cases for exposing these options.

(5) may be an issue for people wanting a baseline for instrumentation done for browser tests (since it is basically the  `--all` implementation for browser testing). We can let people complain before we add it. The browser use-case needs some thinking anyway and we can plod along with old versions of istanbul until we close the gaps.
### hooks
- [x] handle-sigint (nyc does this by default?) **this is handled**
- post-require-hook (do NOT support in nyc. This is some legacy garbage for supporting YUI's loader)
#### Conclusions

I don't think any changes are needed here as long as signals are handled by nyc out of the box.
### reporting
1. summarizer [pkg|flat|nested] - summarizer to use for reporting (new feature in 1.0, default: pkg)
2. watermarks: { statements: [50, 80], lines: [50, 80], branches: [50, 80], lines: [50, 80]}
   - low and high watermarks for reporting low|medium|high coverage. Affects report colorization etc.
3. report-config: object containing properties keyed by report name (detailed config for each report)
#### Conclusions
[...]

PR #893, opened 2018-07-18
WIP: Enable existing options for stand-alone instrument command

AndrewFinlay:
Make include, exclude, cwd and es-modules options available for the stand-alone instrument command to make the behaviour more closely resemble that of the standard nyc command. 

This is aimed at fixing #596 and #890. 

This change allows access to these config options making stand alone instrumentation more flexible.

These options I've added here are taken from config-util.js, with a few changes made to the descriptions.  Also, I've removed the global flag in keeping with the other instrument options but I'm not entirely sure of the effect of this so it may need to change.


JaKXz:
🙌

AndrewFinlay:
@coreyfarrell - It seems that the PR has gone stale with regards to the default es-modules option, will fix that.  Regarding having the options declared only once, it's a good idea but that sounds like a major redesign of the current nyc configuration system and is well beyond the scope of this PR.

@JaKXz I've had another dig though the issue log and couldn't find much more history that I hadn't already referenced.  I did find a discussion between @bcoe and @gotwarlost that providing coverage tools for client-side coverage was a lower priority in terms of the initial nyc release, #293 and #309, but I don't think it's really relevant now.  Off the top of your head is there anything in particular that you know is relevant to this PR?

Regarding the robustness of the instrument command, I haven't dug too deeply but it does seem to have a slightly different implementation to that used in the general nyc command.  Not sure about testing for this, I'll have another look at bcoe's earlier work

AndrewFinlay:
Hi all, this PR/issue seems to have lost momentum and I was hoping to get it moving again.  What do I need to do to get this over the line?

bcoe:
@AndrewFinlay @JaKXz will defer to @coreyfarrell's review, a unit test exercising the new behavior would be nice.

very much appreciate the contribution, as an aside feel free to join the community [here](http://devtoolscommunity.herokuapp.com/), if we're ever taking a long time to get back ahold of you.

AndrewFinlay:
Bump, stale bot

JaKXz:
Have the comments on this PR been addressed? The change makes sense to me at a high level

CHANGELOG.md
```diff
@@ -19,7 +19,7 @@ All notable changes to this project will be documented in this file. See [standa
 
 ### Bug Fixes
 
-* add flag to allow control of intrumenter esModules option, default to looser parsing ([#863](https://github.com/istanbuljs/nyc/issues/863)) ([6b6cd5e](https://github.com/istanbuljs/nyc/commit/6b6cd5e))
+* add flag to allow control of instrumenter esModules option, default to looser parsing ([#863](https://github.com/istanbuljs/nyc/issues/863)) ([6b6cd5e](https://github.com/istanbuljs/nyc/commit/6b6cd5e))
```
coreyfarrell: @bcoe does this create a problem with the utility which generates the CHANGELOG.md?

Really the problem is that the typo is in the commit log, I'm not sure this can be fixed?

test/nyc-bin.js
```diff
@@ -561,10 +563,130 @@ describe('the nyc cli', function () {
...
+
+          var stderr = ''
+          proc.stderr.on('data', function (chunk) {
+            stderr += chunk
+          })
+
+          proc.on('close', function (code) {
+            code.should.equal(1)
+            stdoutShouldEqual(stderr, `
+        Failed to instrument ./not-strict.js`)
```
coreyfarrell: nit: Could you increase the indent so that the line with `Failed to instrument` is indented 2 spaces more than `stdoutShouldEqual`?

CHANGELOG.md
AndrewFinlay: Sorry, didn't realize that would be a problem and think to look if it was generated, I was on typo autopilot there.  
If it's just going to disappear next time the tool runs then we'll remove it

bcoe: the changelog generation looks at the header, doesn't re-generate the whole document, so the edit should be fine.

index.js
```diff
@@ -200,6 +200,11 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
     var ext
     var transform
     var inFile = path.resolve(inputDir, filename)
+
+    if (!_this.exclude.shouldInstrument(filename)) {
```
bcoe: refresh me if I'm wrong about how source-maps behave, but is it possible that a filename on disk would be excluded by our exclude rule, but has a filename the source-map maps to that isn't excluded?

we should look at the tests around excludeAfterInstrument behavior, which addresses some of these edge-cases, and make sure we continue to support these edge-cases.

coreyfarrell: Maybe this should be a separate change?  I'm not sure how it's related to the main purpose of this PR 'enable existing options for stand-alone instrument command'.

test/nyc-bin.js
AndrewFinlay: Didn't realise that was running through a strip indent, will fix in the next commit.

AndrewFinlay:
I've added the clean option to this PR, the only issue is that it's a breaking change, changing the current default behaviour from leaving the existing output dir untouched to removing any existing output and starting again.  I figure having 'clean' default to true is the expected behaviour, but that's not really my decision to make so feedback on that would be appreciated.

I also implemented this in a pretty simple way, I figured I should check to see if you want this to use the clean command on the Nyc object.  If so there's a few hoops to jump through to get that to work.

lib/commands/instrument.js
```diff
@@ -51,14 +56,37 @@ exports.builder = function (yargs) {
...
+      default: cwd
+    })
+    .option('es-modules', {
+      default: true,
+      type: 'boolean',
+      describe: 'tell the instrumenter to treat files as ES Modules'
+    })
+    .option('clean', {
+      describe: 'should the output folder be cleaned before instrumenting files?',
+      default: true,
```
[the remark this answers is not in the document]
AndrewFinlay: Hi Corey, I've tried to write this a few times but it was getting a little wordy. So quite simply, I don't feel there's much point in releasing this without the exclude, include and includeNegated changes.  So if it needs to wait for a major release then that's what should happen.  That way it can be part of a comprehensive effort to address missing functionality in the instrument command and be a PR that you're satisfied doesn't have any issues.

index.js
AndrewFinlay: @coreyfarrell This change is pretty much the core of enabling include, exclude, excludeNegated functionality.  An assessment of whether the file should be instrumented is conducted in the function `_this.exclude.shouldInstrument(filename)`.  If it shouldn't be instrumented we return, if it should we continue and instrument the file.

@bcoe I've been looking over anything source-maps related and I'm still working on getting a good understanding of how everything hangs together.  So far I can see that the 'produce-source-maps' adds a source map while we're instrumenting, which shouldn't be an issue and seems unrelated to the 'source-map' option. 
I'm a little unsure of what 'source-map' should do in the instrument command.  My understanding is that it's mainly used to work with pre-instrumented files with existing source maps but I don't know why you'd be running the instrument command if you'd already instrumented the code elsewhere.  Anyway, my plan is to keep working on this until I'm confident it's good, any pointers, gotchas, or ideas for tests on this stuff would be appreciated.  Thanks

lib/commands/instrument.js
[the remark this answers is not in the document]
bcoe: If we do wait for the next major release, might be worth revisiting some of the exclude logic in general; it's a real pain in the butt when combined with source-maps -- I'd argue it might be worth expanding on the test suite to cover all these weird edge-cases, so that it's easier to add new features like this in the future, without bumping into foot guns (i.e., me and @coreyfarrell not being sure how safe the change is).

AndrewFinlay:
Hi all, I'd like to close this PR and break it up into more manageable pieces, any objections?

Issue #898, opened 2018-08-06
NYC doesn't work correctly with local node_modules


hershmire:
## Expected Behavior

NYC to not exclude local node_modules (i.e. `src/node_modules/...` OR `src/somedirectory/node_modules/**`) BUT to exclude root node_modules. We have a lot of Node applications at the company I work for that use the pattern of creating local node_modules as it helps modularize our code at an application/page-level and provides a nice way to require files instead of doing long relative paths to everything. Below is an example file structure many of our apps could see:

```
app/
  |- node_modules/ (root modules and we expect to exclude)
  |- src/
  |  |- node_modules/ (local to the app)
  |  |  |- local-module-1/
  |  |  |  |- test-node/
  |  |  |  |  |- file-test.js
  |  |  |  |- package.json
  |  |  |  |- file.js
  |  |- middleware
  |  |- pages/
  |  |  |- page-1/
  |  |  |  |- test-node/
  |  |  |  |  |- file-test.js
  |  |  |- file.js
  |  |  |  |- node_modules/ (local to page 1 of application)
  |  |  |  |  |- page-local-module-1/
  |  |  |  |  |  |- test-node/
  |  |  |  |  |  |  |- file-test.js
  |  |  |  |  |  |- package.json
  |  |  |  |  |  |- file.js
  |  |  |- page-2/
  ...
```

## Observed Behavior

Since by default, NYC adds the glob `**/node_modules/**` it will exclude all node_modules, which in our case we don't want. To get around this problem, you have a negateExcludes section where we supply `!src/**/node_modules/**` but that doesn't quite work right as it will force to include our test files which live next to the files that are being tested since negateExclude is evaluated after the exclude rules.

Additionally, even with the `all: true` flag set in package.json, NYC doesn't appear to be picking up all files as I thought it would; especially the ones that I purposely left out of having tests point to – i.e. https://github.com/hershmire/nyc-test/blob/master/src/pages/page-1/components/Name.jsx & https://github.com/hershmire/nyc-test/tree/master/src/pages/page-no-tests-1.

It would be nice to not exclude all node_modules by default but just the root ones for starters. OR at least provide a easier way to turn that off. It gets quite complicated when you also have to take into account babel-plugin-istanbul and understand how to set both up properly.

## Bonus Points! Code (or Repository) that Reproduces Issue

I created a quick demo app that shows what I'm running into here:

https://github.com/hershmire/nyc-test

After running `yarn install`, run `yarn coverage`

[...]

PR #912, opened 2018-08-24
Support turning off node_modules default exclude via flag

hershmire:
This PR is part of resolving [issue #898](https://github.com/istanbuljs/nyc/issues/898). The idea is to be able to turn off the default exclusion of `**/node_modules/**` for users that need/want to support local `node_modules`. As mentioned in that issue, the `negateExclude` (i.e. `!**/node_modules/**`) doesn't work for cases when you still want your exclusion list to still exclude matches within your local `node_modules` directory. 

Related PRs for solving the above issue:
- https://github.com/istanbuljs/istanbuljs/pull/213
- https://github.com/istanbuljs/babel-plugin-istanbul/pull/172

**NOTE:** this test-exclude PR (https://github.com/istanbuljs/istanbuljs/pull/213) will need to get merged and published before this is effective.


hershmire:
@bcoe This issue I wrote up back in August might answer some of your questions – https://github.com/istanbuljs/nyc/issues/898.

hershmire:
@bcoe Any update?

JaKXz:
@hershmire I'm sorry for the late reply & if what I say next might be frustrating to you:

If the `node_modules` that are in those sub-projects/pages are actually local modules (that you hand rolled in place) and not `npm install`ed into those sub-projects/pages, why does the folder have to be called `node_modules`? CJS requires can work with relative paths so I'm not sure that adding all this cyclomatic complexity to `nyc`/`istanbul` is the right approach. As @coreyfarrell said in #898 there are more complex use cases with mono-repos that are becoming more common now to support, so unless this plays nice I don't see this being implemented TBPH. 

to anyone: please correct me if I'm wrong.

bcoe:
@JaKXz @hershmire I would love to add better semantics for including `node_modules`, but I think we should do this work in the [test-exclude](https://github.com/istanbuljs/istanbuljs/tree/master/packages/test-exclude) module, rather than in nyc itself 👍 let's try to pick this work back up though.

coreyfarrell:
@hershmire Sorry for the long delay on this, my intent is that this feature will be available in nyc 14.

@bcoe this feature is already added to `test-exclude` via istanbuljs/istanbuljs#213, now we just need to expose it to users via nyc.

@AndrewFinlay we will want to add the `--exclude-node-modules` option to `nyc instrument`.  I didn't add that to this change because I don't want to cause conflicts with #1007 which is nearly ready for merge.

Issue #913, opened 2018-08-28
Nyc processing files not covered by include when `all` is enabled.


AGrzes:
## Link to bug demonstration repository.
https://github.com/AGrzes/nyc-include-issue

## Expected Behavior
When `include` is specified in nyc configuration it instrument only files matching include. 

## Observed Behavior
Nyc processes all files then discards those that not match `include`. In conjunction with `ts-node` it leads to compilation of unnecessary files and unexpected failures.  

For example in demonstration repository
```
$ npm test

> nyc-issue@1.0.0 test /home/agrzes/git/nyc-issue
> nyc mocha test/**/*.{ts,tsx}


/home/agrzes/git/nyc-issue/node_modules/ts-node/src/index.ts:261
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
bad/bad.ts(1,6): error TS1005: ';' expected.
bad/bad.ts(1,9): error TS1005: ';' expected.
bad/bad.ts(1,13): error TS1005: ';' expected.
bad/bad.ts(1,15): error TS1005: ';' expected.
bad/bad.ts(1,23): error TS1005: ';' expected.
bad/bad.ts(1,26): error TS1005: ';' expected.
bad/bad.ts(1,1): error TS2304: Cannot find name 'This'.
bad/bad.ts(1,6): error TS2304: Cannot find name 'is'.
bad/bad.ts(1,9): error TS2304: Cannot find name 'not'.
bad/bad.ts(1,13): error TS2304: Cannot find name 'a'.
bad/bad.ts(1,15): error TS2304: Cannot find name 'correct'.
bad/bad.ts(1,23): error TS2304: Cannot find name 'ts'.
bad/bad.ts(1,26): error TS2552: Cannot find name 'file'. Did you mean 'File'?

    at createTSError (/home/agrzes/git/nyc-issue/node_modules/ts-node/src/index.ts:261:12)
    at getOutput (/home/agrzes/git/nyc-issue/node_modules/ts-node/src/index.ts:367:40)
    at Object.compile (/home/agrzes/git/nyc-issue/node_modules/ts-node/src/index.ts:558:11)
    at Object.m._compile (/home/agrzes/git/nyc-issue/node_modules/ts-node/src/index.ts:439:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:700:10)
[...]

PR #914, opened 2018-09-05
Bug 913

AGrzes:
Proposed fix for #913.
I created simple falling test and did small change to eliminate the issue.
Please let me know if any improvement is needed.

### DIFF
--- a/test/src/nyc-tap.js
+++ b/test/src/nyc-tap.js
@@ -440,6 +440,26 @@ describe('nyc', function () {
     return done()
   })
 
+  it('Do not transpiles  files when not included', function (done) {
+    var notNeedTranspilePath = path.join(fixtures, './do-not-need-transpile.do-not-transpile')
+    fs.writeFileSync(
+      notNeedTranspilePath,
+      '--> pork chop sandwiches <--\nvar a = 99',
+      'utf-8'
+    )
+
+    var nyc = (new NYC(configUtil.buildYargs(fixtures).parse([
+      '--require=./test/fixtures/transpile-hook',
+      '--extension=.do-not-transpile',
+      '--include=needs-transpile.do-not-transpile'
+    ])))
+
+    nyc.reset()
+    nyc.addAllFiles()
+    fs.unlinkSync(notNeedTranspilePath)
+    return done()
+  })
+
   describe('cache', function () {
     it('handles collisions', function (done) {
       var nyc = new NYC(configUtil.buildYargs(fixtures).parse())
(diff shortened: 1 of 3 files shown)


bcoe:
@coreyfarrell @AGrzes this seems reasonable to me. tldr; you're proposing that files that fail the `shouldInstrument` test should not be processed during the `--all` step.

One question for you, I believe sometimes typescript files are laid out in a different directory, and than compiled in another folder, e.g., `src/` which gets compiled into `lib/`; how would this effect include rules that reference `lib/` and not `src/`.

AGrzes:
My reasoning (regarding old code) id the following 
```javascript
  this.walkAllFiles(this.cwd, function (filename) {
    filename = path.resolve(_this.cwd, filename)
    _this.addFile(filename)
    var coverage = coverageFinder()
    var lastCoverage = _this.instrumenter().lastFileCoverage()
    if (lastCoverage) {
      filename = lastCoverage.path
    }
    if (lastCoverage && _this.exclude.shouldInstrument(filename)) {
      coverage[filename] = lastCoverage
    }
})
```
``` javascript
NYC.prototype.addFile = function (filename) {
  var source = this._readTranspiledSource(path.resolve(this.cwd, filename))
  var instrumentedSource = this._maybeInstrumentSource(source, filename, relFile)

  return {
    instrument: !!instrumentedSource,
    relFile: relFile,
    content: instrumentedSource || source
  }
}
```
```javascript
NYC.prototype._maybeInstrumentSource = function (code, filename, relFile) {
  var instrument = this.exclude.shouldInstrument(filename, relFile)
  if (!instrument) {
    return null
  }

  var ext, transform
  for (ext in this.transforms) {
    if (filename.toLowerCase().substr(-ext.length) === ext) {
      transform = this.transforms[ext]
      break
    }
  }

  return transform ? transform(code, {filename: filename, relFile: relFile}) : null
}
```
* the `addFile` is transpiling source (causing the issue reported by me).
* `_maybeInstrumentSource` will  instrument the source based on `shouldInstrument` check.
* The result of instrumentation will be stored in `coverage` variable based on `shouldInstrument` check.
* In essence if `shouldInstrument` returns false then any instrumentation results are discarded so I believe my change does not change end result (in cases where there was no error during transpiling)

Regarding `src` and `lib` I'm not sure I get your question. Could you explain further?

test/src/nyc-tap.js
```diff
@@ -440,6 +440,26 @@ describe('nyc', function () {
     return done()
   })
 
+  it('Do not transpiles  files when not included', function (done) {
```
bcoe: should read:

"does not transpile files when not included"

PR #957, opened 2019-01-03
feat: avoid hardcoded HOME for spawn-wrap working dir

### LINKED ISSUE #951: spawn-wrap causes failure due to unwritable homedir
## Observed Behavior

When using nyc as a system user that is not allowed to write to its home directory (e.g. `HOME=/nonexistent`), a command like `nyc qunit test` or some other use of nyc fails as follows:

```
+ npm run-script test

> nyc -r text -r html qunit test/

/src/node_modules/nyc/node_modules/mkdirp/index.js:90
                    throw err0;
                    ^

Error: EACCES: permission denied, mkdir '/nonexistent'
    at Error (native)
    at Object.fs.mkdirSync (fs.js:923:18)
    at sync (/src/node_modules/nyc/node_modules/mkdirp/index.js:71:13)
    at Function.sync (/src/node_modules/nyc/node_modules/mkdirp/index.js:77:24)
    at setup (/src/node_modules/nyc/node_modules/spawn-wrap/index.js:414:10)
    at wrap (/src/node_modules/nyc/node_modules/spawn-wrap/index.js:60:18)
    at Object.<anonymous> (/src/node_modules/nyc/bin/nyc.js:57:3)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
```

I believe this is because spawn-wrap, unlike nyc, defaults to using a a sub directory of `HOME` as its working directory. Instead of using process.cwd() or TMPDIR etc, like nyc does.

This should probably be fixed upstream in spawn-wrap, but could be worked around in nyc by passing NYC_CWD as the `workingDir` parameter to the spawn-wrap function.

Currently ([nyc@v13.1.0:/bin/nyc.js](https://github.com/istanbuljs/nyc/blob/v13.1.0/bin/nyc.js#L44-L57)), nyc invokes spwan-wrap with two parameters, leaving the third parameter (workingDir) unset. In spawn-wrap ([spawn-wrap:/index.js](https://github.com/tapjs/spawn-wrap/blob/v1.4.2/index.js#L40), this means it defaults to the setup of creating a directory in the os-home directory.

```
Node v11.4.0
npm 6.5.0
```

Krinkle:
By default, spawn-wrap writes temporary files to HOME.
It used to be /tmp, but it changed that to HOME to support
environments that have 'noexec' flags set on their tmpfs mount.
Ref https://github.com/tapjs/spawn-wrap/issues/3.

The problem with this is that nyc now no longer works in environments
without a (writable) home directory (e.g. the 'nobody' user on
Linux).

While it is fine to fallback to HOME, it should write elsewhere
if that is unavailable, and ideally in a way that doesn't require
every sysadmin or end-user to hardcode some environment variable
in their package.json specifically for nyc or spawn-wrap.

A common way to communicate this intent is with the XDG_CACHE_HOME
environment variable.

Fixes #951.

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -42,6 +42,9 @@ if ([
   if (argv.all) nyc.addAllFiles()
 
   var env = {
+    // Support running nyc as a user without HOME (e.g. linux 'nobody'),
+    // https://github.com/istanbuljs/nyc/issues/951
+    SPAWN_WRAP_SHIM_ROOT: process.env.SPAWN_WRAP_SHIM_ROOT || process.env.XDG_CACHE_HOME || require('os').homedir(),
     NYC_CONFIG: JSON.stringify(argv),
     NYC_CWD: process.cwd(),
     NYC_ROOT_ID: nyc.rootId,


JaKXz:
@Krinkle would you mind rebasing this PR? We have since fixed the failing test that's broken here :)

Krinkle:
@JaKXz Thanks. Rebased now.

bin/nyc.js
```diff
@@ -42,6 +42,9 @@ if ([
   if (argv.all) nyc.addAllFiles()
 
   var env = {
+    // Support running nyc as a user without HOME (e.g. linux 'nobody'),
+    // https://github.com/istanbuljs/nyc/issues/951
+    SPAWN_WRAP_SHIM_ROOT: process.env.XDG_CACHE_HOME || require('os').homedir(),
```
coreyfarrell: What if `SPAWN_WRAP_SHIM_ROOT` is already set?  Maybe it would be better to use:
```js
SPAWN_WRAP_SHIM_ROOT: process.env.SPAWN_WRAP_SHIM_ROOT || process.env.XDG_CACHE_HOME || require('os').homedir(),
```

Krinkle: @coreyfarrell Sure, that seems fair.

I omitted this because I thought it'd be unexpected for the user's own spawn-wrap configuration (if that exists) to apply to nyc's use of spawn-wrap.

coreyfarrell: Honoring the existing `SPAWN_WRAP_SHIM_ROOT` value prevents this from being a breaking change.

coreyfarrell:
@Krinkle thanks for this contribution!

PR #961, opened 2019-01-08
build: move windows tests to travis

JaKXz:


JaKXz:
Related: https://github.com/jasongin/nvs/issues/120

.travis.yml
```diff
@@ -1,14 +1,23 @@
...
 node_js:
+  - "node"
+  - 10
   - 8
   - 6
-  - "stable"
+matrix:
+  include:
+    - os: windows
+      node_js: "latest"
```
JaKXz: should this be allowed to fail on Windows? We are currently getting a consistent out of memory error: https://travis-ci.org/istanbuljs/nyc/jobs/476663873

coreyfarrell: I'd say for now we could just drop testing `latest`?  It's not being tested by appveyor.

https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system says that Windows VM's get 8GB ram which I'd think should be enough but Windows is extremely wasteful.  More than anything I'm concerned this could be identifying a memory management issue with latest node.js.  Otherwise `nyc` should use more than travis-ci offers in all versions of node.js.

coreyfarrell: @JaKXz let's just drop `latest` from Windows on this PR.  I can't even get 11.6.0 to install on appveyor so switching everything to travis-ci won't lose anything for us.  Once `latest` is removed you can just merge, no need to wait on me.

coreyfarrell: I noticed that the branch is under istanbuljs/nyc, I'm pushing an additional commit to disable `latest` on Windows.

bcoe:
@JaKXz thank you for seeing this over the finish line.

PR #963, opened 2019-01-08
DO NOT MERGE: Try node.js latest in appveyor.

coreyfarrell:
#961 is testing migration to travis-ci, node.js latest is throwing `ENOMEM`.  The point of this PR is to see if we get the same result in appveyor.


JaKXz:
The appveyor hook seems to be 💔.

PR #964, opened 2019-01-09
Temp dir tests

AndrewFinlay:
Temp dir tests as promised, take a look and let me know if I'm heading in the right direction with these.

Tests aim to cover default temp directory, setting temp dir with the old option temp-directory, and setting temp-dir with the new option temp-dir.  Any directories created while running the tests are removed at the end.


JaKXz:
This is awesome, thank you @AndrewFinlay!

PR #968, opened 2019-01-14
Make update to caching information in readme

brokentone:
Since https://github.com/istanbuljs/nyc/pull/454, caching has been enabled by default but the README does not reflect this. Our team lost some time following the README, so contributing back our findings for the next team!

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,9 +251,7 @@ modules should be required in the subprocess collecting coverage:
 
 ## Caching
 
-You can run `nyc` with the optional `--cache` flag, to prevent it from
-instrumenting the same files multiple times. This can significantly
-improve runtime performance.
+`nyc`'s default behavior is to cache instrumented files to disk to prevent instrumenting source files multiple times, and speed `nyc` execution times. You can disable this behavior by running `nyc` with the `--cache false` flag. You can also change the default cache directory from `./node_modules/.cache/nyc` by setting the `--cacheDir` flag.
 
 ## Configuring `nyc`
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -251,9 +251,7 @@ modules should be required in the subprocess collecting coverage:
 
 ## Caching
 
-You can run `nyc` with the optional `--cache` flag, to prevent it from
-instrumenting the same files multiple times. This can significantly
-improve runtime performance.
+`nyc`'s default behavior is to cache instrumented files to disk to prevent instrumenting source files multiple times, and speed `nyc` execution times. You can disable this behavior by running `nyc` with the `--cache false` flag. You can also change the default cache directory from `./node_modules/.cache/nyc` by setting the `--cacheDir` flag.
```
JaKXz: Isn't it `--cache-dir`?

brokentone: Turns out... that while this is processed with the rest of the config, it's not actually specified in the `yargs` -- which I assume means won't come through. Assumedly it can be specified from `.nycrc` files though.
https://github.com/istanbuljs/nyc/blob/master/lib/config-util.js#L105-L111

1. I can add this to the `yargs` settings.
2. I can remove the reference to it and let someone else work on this later.

JaKXz: It would be great to see it added! Cc @bcoe

brokentone:
@JaKXz @coreyfarrell, etc -- I think I've made the necessary updates, please let me know if you need anything else from me!

JaKXz:
I'll leave this to @bcoe or @coreyfarrell to merge

Issue #833, opened 2018-05-08
Include certain node modules


wmmihaa:
Is there a way to include some of the files in the node_modules folder? I'm using a **.nycrc** file as:
``` json
{
  "require": [],
  "include": [
    "**/*.js",
    "**/node_modules/microservicebus-core/lib/*.js"
  ],
  "exclude": [
    "**/*.test.js"
  ],
  "reporter": [
    "text",
    "lcov"
  ],
  "sourceMap": false,
  "instrument": false,
  "all": true
}
```
But files in the */node_modules/microservicebus-core/lib/* are not covered:

``` bash
  3 passing (218ms)

-------------------|----------|----------|----------|----------|-------------------|
File               |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------------|----------|----------|----------|----------|-------------------|
All files          |    31.95 |    18.89 |       10 |    31.95 |                   |
 SettingsHelper.js |    69.23 |    47.06 |       75 |    69.23 |... 96,101,104,113 |
 Utils.js          |    15.38 |     1.79 |        0 |    15.38 |... 32,233,234,235 |
-------------------|----------|----------|----------|----------|-------------------|
```

pingynator:
Seems like a usecase not covered by nyc.
It seems to be possible to include files from every folder expect node_modules.

Problem seems to be inside nyc: node_moduls/nyc/test-exclude/index.js line 126 and 133 in nyc version 11.8.0.
`exportFunc.defaultExclude = [
  '**/node_modules/**'`

By default nyc excludes everything inside node_modules and files from node_modules can not be included. Sadly explicit includes do not overwrite implicit excludes.

BAD Workaround:
delete line 133 in node_moduls/nyc/node_modules/test-exclude/index.js
"**/node_modules/**"
(NYC Version 11.8.0)


Would like to see better solutions in the next version of nyc. Explicit includes and excludes should override implicit ones.

antoniocascais:
Hi @pingynator. Thanks for the response!

For now we also came up with a workaround (which was forking NYC and remove the logic to exclude the node_modules), but we don't want to have this as a final solution. We will keep digging and try to solve the problem in a proper way...

Thansk for the help! Your comment from 30th of May was **really** helpful in understanding why the coverage was not working properly :+1:

pingynator:
Hey @antoniocascais. I am glad you found a solution.

[...]

PR #977, opened 2019-01-25
fix: Exclude negated not working with '--all' switch

AndrewFinlay:
While I've been tinkering with the instrument command I noticed that the exclude negated functionality provided by the test-exclude package wasn't working for 'nyc --all' and in the future for 'nyc instrument'.  

This PR includes a set of tests and a fix for that behaviour.

Note that it does promote an existing sub-dependency, 'array-uniq' to package.json, and at a higher major version than before.  So if this is a problem let me know and we'll try and find another way to do it. 

Also I've added a couple of files and folders to the cli test fixtures, and added an exclude negation to .gitignore to allow testing exclude negation on a file in node_modules.


index.js
```diff
@@ -242,16 +243,20 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
...
-  glob.sync(pattern, { cwd: dir, nodir: true, ignore: this.exclude.exclude }).forEach(function (filename) {
-    visitor(filename)
+  let filesToWalk = glob.sync(pattern, { cwd: dir, nodir: true, ignore: this.exclude.exclude })
+
+  // package node-glob no longer observes negated excludes, so we need to restore these files ourselves
+  const excludeNegatedPaths = this.exclude.excludeNegated
+  excludeNegatedPaths.forEach(pattern => {
+    filesToWalk = filesToWalk.concat(glob.sync(pattern, { cwd: dir, nodir: true }))
   })
+  filesToWalk = arrayUniq(filesToWalk)
```
JaKXz: To avoid the one time use dependency and the multiple operations on `filesToWalk`, I'd write it something like this. Though yours is a bit clearer in terms of intentions so 🤷‍♂️ 

```js
const excludeNegatedPaths = this.exclude.excludeNegated.map(pattern => glob.sync(pattern, { cwd: dir, nodir: true })

// do the walkAllFiles thing
return glob.sync(pattern, { cwd: dir, nodir: true, ignore: this.exclude.exclude })
  .concat(excludeNegatedPaths)
  .filter((value, index, self) => self.indexOf(value) === index)
  .forEach(visitor)
```

AndrewFinlay: Yeah, I felt a bit bad adding arrayUniq as a dependency as it's so simple too that it would be easier to define it inline than add it to package.json.  I figured promoting it may not be such an issue since it was already a sub dependency, that said I'm certainly not against defining it inline as
```javascript
const arrayUniq = x => [...new Set(x)];
```
I did have a go at removing the `let` in there but I got to a point where I felt I was being a bit pedantic and should probably just leave it be.  Anyway I'll have another look and see if I can make it a bit nicer, and remove the dep, drop the let and return the results.

JaKXz: I don't think we have access to the spread operator for node 6

coreyfarrell: Actually we do have access to array spread and rest arguments.  So `[...new Set(x)]` works in node 6, as does `function (...args) {}`.  Object spread is not available until node.js 8 (so we cannot do things like `{...defaultOpts, ...options}`, for that we still have to use `Object.assign`.

AndrewFinlay:
Wow, that's got to be the worst case of 'it works on my machine' I've seen in a while.  I'll do some digging to see what's going on.

AndrewFinlay:
So I can't get this to fail the `should include 'node_modules' using exclude patterns` test on my local machine.  Running with node 8.15.0, npm 6.4.1 on osx, and no modifications to the source that aren't part of the PR.  

Strangely, I can get the `should include 'node_modules' using exclude patterns` test to pass if a console.log is left in the middle of the walkAllFiles function, see commit 82ed06e.  Now I'm not suggesting this as a solution, but it's an interesting result that's got me wondering if there's some timing issues in here.

Anyway I figured I'd see if this rung any bells before I dive a little deeper into this.

coreyfarrell:
I just pushed eac3110 to my own fork of nyc and it looks like it's going to pass (just waiting on a couple OSX builds).

https://travis-ci.org/coreyfarrell/nyc/builds/496549022

Do you mind if I do a force-push to your nyc branch?  What I'd like to from your branch is:
```sh
git reset eac3110 --hard
git commit --amend --no-edit
git push --force
```

The result will be to delete commits that came after eac3110 and replace eac3110 with another commit that has the same exact content (just force generating a new SHA).

AndrewFinlay:
Can do, give me a minute

JaKXz:
If my suggestion is causing this much trouble I don't think it's worth it for now... It was just a minor nit anyway

AndrewFinlay:
@JaKXz To my eye the code change really seemed to be a minor refactor, aside from the return statement.  At the moment I can't explain the discrepancy, and changing the code back makes me feel I'm leaving a landmine behind.

So my plan is to do a little more digging, I'll keep you up to informed as to where the PR is at.

index.js
coreyfarrell: Deferred for now so we can look at merging what we have.

AndrewFinlay:
So it looks like this might fix #833

test/nyc-bin.js
```diff
@@ -101,6 +101,48 @@ describe('the nyc cli', function () {
...
+
+      proc.on('close', code => {
+        code.should.equal(0)
+        stdout.should.not.match(/excluded\.js/)
+        stdout.should.match(/exclude-negated\.js/)
+        done()
+      })
+    })
+
+    it('should include \'node_modules\' using exclude patterns', function (done) {
```
coreyfarrell: Lets add another test (in a follow-up) to do nearly the same thing as this, just without the `--all` argument and with a test script that actually has `require('./include-exclude/node_modules/cover-me.js');`?  This will prove that `--exclude '!**/node_modules/**'` works again.  I've already done this basic test at the command-line and it did what I expected.

AndrewFinlay: Cool, I'll put something together after lunch

PR #984, opened 2019-02-02
feat: Support nyc report --check-coverage

isaacs:


isaacs:
(force-pushed to fix lint error)

test/nyc-bin.js
```diff
@@ -48,6 +48,38 @@ describe('the nyc cli', function () {
     })
   })
 
+  describe('report and check', function () {
+    it('should show coverage check along with report', function (done) {
```
coreyfarrell: Maybe we should have a second test to verify behavior when `--check-coverage` is not specified?  Granted we should have already had that but just an observation.

coreyfarrell: Actually I can take care of this at a later time.  Unless you're already working on this don't worry about it.

coreyfarrell:
> So does 2 approvals means it's cool for me to land this? I just realized I have write access on this org, but I don't wanna mess up any process.

I just did it.  For this PR a 'squash and merge' with the provided commit message was fine (you followed conventional-commit).  I have no objection if you use 'squash and merge' for approved non-breaking changes as long as you make sure the commit message follows conventional-commit rules.  For breaking changes I'd prefer that we have a chance to decide when those get landed.

PR #982, opened 2019-01-31
Add support to exclude files on coverage report generation

henriquemotaesteves:
Add support to exclude files on coverage report generation (https://github.com/istanbuljs/istanbuljs/pull/275) to `nyc` (https://github.com/istanbuljs/nyc/issues/956).

PR #997, opened 2019-02-22
test: consolidate on tap test runner, stop running build step

bcoe:
In this pull request I've refactored tests to stop running a build step. to facilitate this there's now a helper called `reset-state.js` that resets global state between test runs.

I've also opted to move us back to one test runner (tap) now that this consolidation has happened. Here's how the tests are laid out now:

`test/lib/reset-state.js`, called between test runs to reset state.
`test/nyc-index.js`, tests mainly for methods in `index.js`.
`test/nyc-integration.js`, CLI integration tests (@JaKXz after we land this, I think this would be a great candidate for using snapshots like we did with `c8`).
`test/process-args.js`, tests for the process-args.js helper.

TODO:

- [x] ~coverage for process-args.js is missing for some reason~ I'm not sure how this could have ever worked because we don't instrument process-args, we will have coverage for this file once we switch our coverage approach.

Note:

I've commented out the test:

`it('should include \'node_modules\' using exclude patterns')`

which also seems to be failing on `master`, CC: @AndrewFinlay.

Future Work:

Once we land this, I'd like to switch us to using `c8` shortly afterwards, and eliminate the other wonky build step we have in the codebase.

It would also be awesome to switch to snapshots for `test-integration.js`.


coreyfarrell:
> I've commented out the test:
>
> `it('should include \'node_modules\' using exclude patterns')`
>
> which also seems to be failing on `master`, CC: @AndrewFinlay.

I just posted https://github.com/istanbuljs/nyc/commits/ci-test with a revert of `feat: add support to exclude files on coverage report generation (#982)`.  @AndrewFinlay was correct that his PR started failing once #982 was merged because travis-ci was testing the result of his branch merged with master.

PR #1000, opened 2019-02-22
chore: Full self coverage

coreyfarrell:


index.js
```diff
@@ -24,17 +24,11 @@ const api = require('istanbul-api')
...
-} catch (e) {
-  /* istanbul ignore next */
-  ProcessInfo = require('./lib/process.js')
-}
+const ProcessInfo = require('./lib/process.js')
 
 /* istanbul ignore next */
-if (/index\.covered\.js$/.test(__filename)) {
-  require('./lib/self-coverage-helper')
+if (/self-coverage/.test(__dirname)) {
```
coreyfarrell: This conditional is the only 'self-test' code which remains in the packaged NYC.

package.json
```diff
@@ -5,11 +5,12 @@
   "main": "index.js",
   "scripts": {
     "bundle": "bundle-dependencies update",
+    "pretest": "npm run clean && npm run instrument",
```
coreyfarrell: Even if we don't go with the rest of this change it would be nice to split out the `pretest` script.

build-self-coverage.js
```diff
@@ -1,21 +1,42 @@
...
+    nodir: true
+  })
+
+  result.forEach(file => {
+    instrumentFile(file)
+  })
+}
+
+function instrumentAll () {
+  const globPatterns = ['package.json', 'index.js', 'bin/*.js', 'lib/**/*.js']
```
JaKXz: should we be instrumenting `package.json`?

package.json
```diff
@@ -18,22 +19,19 @@
...
     "exclude": [
       "node_modules",
       "bin",
       "coverage",
+      "self-coverage",
       "test/fixtures/coverage.js",
       "test/build/*",
       "test/src/*",
       "test/nyc.js",
       "test/process-args.js",
```
JaKXz: what's weird is that on master right now, if you just have `test` in the `exclude` list the test fails... i have no idea why but I would love to see that change work

build-self-coverage.js
coreyfarrell: We're not, I added a comment.

package.json
coreyfarrell: I think this is because we need to allow most of `test/fixtures/**`.

test/src/nyc-tap.js
```diff
@@ -428,7 +425,7 @@ describe('nyc', function () {
       )
 
       var nyc = (new NYC(configUtil.buildYargs(fixtures).parse([
-        '--require=./test/fixtures/transpile-hook',
+        '--require=' + transpileHook,
```
JaKXz: nit: could we use template strings for these?

JaKXz:
From @bcoe's [comment](https://github.com/istanbuljs/nyc/pull/997#issue-255314919) should this be replaced by just using c8 instead of a hand-rolled thing like this?

coreyfarrell:
> From @bcoe's [comment](https://github.com/istanbuljs/nyc/pull/997#issue-255314919) should this be replaced by just using c8 instead of a hand-rolled thing like `build-self-coverage`?

I don't object if this is replaced next week or next month, but I think it's important that NYC testing shows coverage for all sources now. 

I don't think we should switch to c8 until it's shown to be ready:
* [ ] We should compare results of NYC to c8, make sure nothing is out of whack.
* [ ] The switch to c8 needs to be tolerant of node.js versions supported by NYC.  I'd prefer that `nyc test` show coverage results, but running `nyc test` on node.js 8 needs to succeed.  I'm OK if coverage only appears in node.js 10.12.0+ but I'm not sure how this auto-detection of c8 support would work (or if c8 can handle this for us).

PR #1003, opened 2019-02-27
wip: add `cwd` option to instrument command

AndrewFinlay:
This adds support for the `instrument --cwd` command, and some supporting tests.  There are few changes in here so I've got some more detailed notes for the commit, and a list of assumptions.  BTW, I'm not going to widen the scope of this PR, this is it!

- Notes:
  Add instrument command support for the `cwd` option
  Refactor nyc-integration instrument tests so tests with output folder specified aren't in the no output folder suite
  Refactor `instrument.js` and `instrumentAllFiles` command to use modern javascript
  Remove call to deprecated `process.exit()` in instrument command
  Removed unused transform metadata param `relFile`
  Use file extension as an index to find the correct transform, rather than trying to match a transform name to the rightmost characters of a filename
  
- Assumptions:
  If you're instrumenting a single file `cwd/a/b/d.js`, and specifying an output directory `cwd/e`, the result will be `cwd/e/d.js`
  That we want to use file extensions for find transforms, the existing implementation would use the '.js' instrumenter for both `.js` and `.mjs` files, these changes assume that was wrong.
  I've set `metadata.filename` to be a relative path here, as that's what the existing tests expect
  When instrumentAllFiles fails to find the correct transform, it returns the uninstrumented source,
  It looks like the `relFile` property isn't used anymore, so I'm assuming it can be safely removed


test/nyc-integration.js
```diff
@@ -525,6 +527,42 @@ describe('the nyc cli', function () {
...
+        const proc = spawn(process.execPath, args, {
+          cwd: fixturesCLI,
+          env: env
+        })
+
+        proc.on('exit', function (code) {
+          code.should.equal(0)
+          const target = path.resolve(subdir, 'output-dir', 'index.js')
+          fs.readFileSync(target, 'utf8')
+            .should.match(/console.log\('Hello, World!'\)/)
```
JaKXz: I think we should start using tap's snapshot testing feature here, with new tests being written. I think you'll have to replace `it` with something like
```js
const spawn = require('await-spawn')

test('title', t => { 
  return spawn(...).then(() => {

  });
```

JaKXz: I'm planning to do a full overhaul of this file soon... but I'm so ___ing swamped right now.

JaKXz: We're ignoring this for now, it can be taken care of after the fact so we can get these merged

AndrewFinlay: @JaKXz Cool, I figured I'd write code to what's there.  When the snapshot tests start then I'll write to that.

BTW I had a chat with @coreyfarrell about this PR, and it's going to get broken up into an ES6 PR and a`cwd` PR.  So I wouldn't put too much time into reviewing it right now, thanks.

AndrewFinlay:
This is being split up into a es6+ PR and a instrument --cwd PR, so I'm closing this one.

PR #1005, opened 2019-02-27
feat: add `delete` option to instrument command

AndrewFinlay:
This change adds a `--clean` switch to the nyc instrument command, it must be set true and an output directory must be specified.

I've set the default value for this to false so it can be a non-breaking change, but this should change in the next major release.  I've included tests for this function, introduced lint support for `afterEach` in the tests, and required the `makeDir` module.


coreyfarrell:
One concern I have with this PR is that the `--clean` flag has meaning in the global CLI context, it controls deletion of `.nyc_output/`.  I'm not sure how I feel about having a sub command use the same flag name as the main command but for a different purpose.

CC @JaKXz @bcoe

AndrewFinlay:
After a chat with @coreyfarrell I've changed the name and added some level of defence against `rimraf` foot guns.

PR #1006, opened 2019-02-28
feat: add `es-modules` option to instrument command

AndrewFinlay:
This is pretty straightforward, allow the es-modules option to flow through to the instrumenter when running the standalone instrument command.

Tests adapted from the existing tests for the regular es-modules command.

### DIFF
--- a/lib/commands/instrument.js
+++ b/lib/commands/instrument.js
@@ -51,6 +51,11 @@ exports.builder = function (yargs) {
       type: 'boolean',
       description: 'should nyc exit when an instrumentation failure occurs?'
     })
+    .option('es-modules', {
+      default: true,
+      type: 'boolean',
+      description: 'tell the instrumenter to treat files as ES Modules'
+    })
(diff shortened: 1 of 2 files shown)


coreyfarrell:
This might be a breaking change as currently `nyc instrument file.js` effectively has `--es-modules=false`.  Not sure this is an issue, we might already be looking at pushing forward with NYC@14 anyways due to some of the include/exclude processing changes.

coreyfarrell:
@bcoe @JaKXz I think setting the default true is correct as it matchs other parts of NYC, should this be flagged as:
```
BREAKING CHANGE: `nyc instrument` now enables the `--es-module` option by default.
```

BTW one of the projects I work with uses gerrit for code review, one of the really nice things is that the commit message is part of the review (not suggesting gerrit is viable, just wish github had a similar process as it would work really well for conventional-commits).  So anyways please let me know what you think of the breaking change text I stated above (or if you think this should not be marked breaking).  This can matter for sources such containing `function package() {}` or other keywords forbidden by `'use strict'`, the new default will be to fail instrumentation on sources that violate certain `'use strict'` rules.

PR #1007, opened 2019-03-01
feat: add `include` and `exclude` options to instrument command

### LINKED ISSUE #596: Globs are not working for $ istanbul instrument -x "**glob**"

### LINKED ISSUE #890: Instrumenter should ignore excluded files and folders

AndrewFinlay:
This allows you to better select which files are instrumented when running the instrument command.  It's a pretty minimal implementation to make it as easy as possible to integrate with other proposed changes.  Test support is included with this change.

Fixes #596, fixes #890


coreyfarrell:
Do we want to add the `--cwd` option as well since this is passed to the `test-exclude` constructor?  This could be useful if `process.cwd()` is not the correct location.

coreyfarrell:
Hold off a minute, I may have noticed something.

index.js
```diff
@@ -202,6 +202,11 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
...
     var transform
     var inFile = path.resolve(inputDir, filename)
+
+    if (!_this.exclude.shouldInstrument(filename)) {
+      return
+    }
+
     var code = fs.readFileSync(inFile, 'utf-8')
 
     for (ext in _this.transforms) {
```
coreyfarrell: This results in no output for files that are not instrumented.  But if I have:
```
src/index.js
src/no-coverage.js
```
With `src/no-coverage.js` excluded, I think I still need both `dest/index.js` and `dest/no-coverage.js` to be created, just `dest/no-coverage.js` should be unmodified.  Same if I run `nyc instrument src/no-coverage.js` - that should print the original source.

So I'm thinking:
```
if (_this.exclude.shouldInstrument) {
  for (ext in ...) { ... }
}
```

I suspect this will change some of the tests, you will have to look for `cov_` or not in the output files for example instead of checking if the file exists/doesn't.

bcoe: If I understand correctly, you're thinking your point is that `dest/` should have a fully functional codebase, just some files will be instrumented and some won't. I agree, it would be a bit shocking to be missing files, it makes it impossible to import the library.

I believe we check for the `cov__` global in a few other tests, I'd look for an existing test using this approach.

coreyfarrell: Yes, though this idea might require bypassing `walkAllFiles` entirely, just using:
```js
glob.sync('**/*', { cwd: input, nodir: true }).forEach(visitor)
```

Specifically `walkAllFiles` uses the extension and exclude filters which we don't really want if we're attempting to copy everything from src to dest.  This could be problematic though when using `npx nyc instrument . dest` as that would include `node_modules`.  Maybe `nyc instrument` needs a `--full-copy` option?  If enabled we directly use `glob.sync` to capture all files, if not then we use `testExclude.globSync`.  I'm leaning towards leaving that option disabled by default.

Possibly future enhancement: when `nyc instrument` copies files which have `chmod +x` enabled should we do the same for the copied file?

bcoe: Let's create an issue for keeping the perms the same; I think it would potentially even be worth splitting this into its own library -- it sounds like we want something like `cpr` that:

* allows us to apply our own `include`/`exclude` rules.
* maintains file permissions.

Definitely beyond the scope of this pull however; although I think maintaining the existing behavior of copying over the whole directory is smart.

AndrewFinlay: I've pushed a new commit that addresses some of these concerns, namely copying all existing files in `input` to output before instrumentation and replacing output files in place.  To do this I've brought in `fs-extra` as it seems like a popular and maintained module, but I'm not particularly attached to it so if it needs to change let me know of your preferred replacement.

index.js
```diff
@@ -200,6 +200,10 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
   try {
     const stats = fs.lstatSync(input)
     if (stats.isDirectory()) {
+      if (output) {
+        fs.copySync(input, output)
```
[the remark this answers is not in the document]
coreyfarrell: A potential issue with this approach:
```sh
nyc instrument --delete . self-coverage
```

This would result in copying the entire project into `./self-coverage`, which would be correct.  The issue is that `this.walkAllFiles` happens after these files are created.  I think we need to stop using `this.walkAllFiles`, instead perform:
```js
const files = this.exclude.globSync(input);
if (output) {
  fs.copySync(input, output);
}
files.forEach(relFile => {
  visitor(relFile);
})
```

This way `testExclude.globSync` is run before we create any files.  I know this shouldn't matter as someone following the above example should add `self-coverage` to the exclude list, but I think it will turn into an issue.

index.js
```diff
@@ -6,7 +6,7 @@ const arrify = require('arrify')
 const cachingTransform = require('caching-transform')
 const util = require('util')
 const findCacheDir = require('find-cache-dir')
-const fs = require('fs')
+const fs = require('fs-extra')
```
coreyfarrell: Pulling in `fs-extra` just for `copySync` makes me sad.  That said I checked and using `cpy` increases the production install size by like 5MB (possibly due to using globby).  I'm not aware of a simple globless module that just does recursive copy.

I'm concerned about replacing the built-in `fs` with the `fs-extra` module.  I'd prefer that this instead:
```js
const fs = require('fs');
const { copySync } = require('fs-extra');
```

lib/commands/instrument.js
```diff
@@ -48,6 +49,16 @@ exports.builder = function (yargs) {
...
       description: 'should nyc exit when an instrumentation failure occurs?'
     })
+    .option('include', {
+      alias: 'n',
+      default: [],
+      describe: 'a list of specific files and directories that should be instrumented, glob patterns are supported'
+    })
+    .option('exclude', {
+      alias: 'x',
+      default: testExclude.defaultExclude,
```
coreyfarrell: Does `nyc instrument input output` honor an `exclude` setting in package.json or .nycrc?

test/nyc-integration.js
```diff
@@ -628,55 +622,163 @@ describe('the nyc cli', function () {
...
         proc.on('close', function (code) {
           code.should.equal(0)
-          var files = fs.readdirSync(path.resolve(fixturesCLI, './output'))
+          const files = fs.readdirSync(path.resolve(fixturesCLI, './output'))
           files.should.include('index.js')
           done()
         })
       })
 
+      it('prevents writing output to a sub-dir of the input dir', function (done) {
```
coreyfarrell: This is potentially a breaking change.  Can we find a way around this?  Maybe something like:
```js
glob.sync(`${input}/*`, {dot:true}).forEach(item => {
  fs.copySync(path.join(input, item), path.join(output, item));
});
```
This might require dealing with directories and files separately, so:
```js
const files = glob.sync(`${input}/*`, {dot: true, nodir: true});
const dirs = glob.sync(`${input}/*/`, {dot: true});
```

index.js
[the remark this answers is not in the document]
AndrewFinlay: Good point, `fs-extra.copySync` currently prevents this from happening as it won't let you copy `.` into `./dest`.  It seems a more custom solution without `fs-extra.copySync` may be a better approach.

index.js
AndrewFinlay: It seemed like the best fit out of the modules that I looked at, not too big, good support and a lot of users.  It also had a copySync method that fits well with `nyc`.  `fs-extra.copySync` does have a big drawback in that it can't copy `./` to `./dest`, that the current `nyc instrument` implementation allows.  So this *might* disappear to be replaced with a custom solution.

lib/commands/instrument.js
AndrewFinlay: Yes, with this change include/exclude settings in `.nycrc` or the `package.json` `nyc` stanza will be honored by `nyc instrument`.  The settings are recovered from the config file/nyc stanza and passed into the `nyc` constructor when creating an instance to run the instrumentation.  From here it will be applied by the filters in 'test-exclude'.

test/nyc-integration.js
AndrewFinlay: I'm looking into it now, I should have a solution soon that *maybe* drops the `fs-extra` dependency.

lib/commands/instrument.js
AndrewFinlay: So would this item need specific tests?

coreyfarrell: I'd be OK with doing it in a follow-up though as long as you've verified manually that it works.

test/nyc-integration.js
coreyfarrell: Please make sure that whatever solution you choose maintains `chmod` value for the copy.

index.js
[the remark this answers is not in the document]
coreyfarrell: One other though, should the copy skip the `.git/` folder?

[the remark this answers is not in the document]
AndrewFinlay: Most likely, I can't think of a use case where you'd need to copy the `.git/` folder from `.` to `./dest` while instrumenting files.

lib/commands/instrument.js
AndrewFinlay: I verified it before posting my response.  To be honest I thought that allowing `include` and `exclude` to be configured via cli, config file or config stanza was the point of this work.

package.json
```diff
@@ -95,7 +97,6 @@
     "any-path": "^1.3.0",
     "chai": "^4.2.0",
     "coveralls": "^3.0.3",
-    "glob": "^7.1.3",
```
JaKXz: this looks like a merge conflict?

JaKXz:
@AndrewFinlay can we get these changes rebased and merged? :) cc @coreyfarrell @bcoe

coreyfarrell:
@JaKXz or @bcoe can I get a second review on this?  Thanks.

index.js
```diff
@@ -191,8 +195,27 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
   try {
     const stats = fs.lstatSync(input)
     if (stats.isDirectory()) {
+      const globOptions = { dot: true, mark: true, ignore: ['**/.git', '**/.git/**/*'] }
+      const outputPaths = (output) ? glob.sync(`${path.resolve(input)}/**/*`, globOptions) : []
```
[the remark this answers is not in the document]
AndrewFinlay: Taking a look right now, I think you're right though so I'll see if I can make this more succinct.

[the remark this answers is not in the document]
AndrewFinlay: Hold up on looking at the last push, just remembered why I had to write it so strangely in the first place.  I'll verify that what's there works in the edge case I was working around.

index.js
```diff
@@ -195,26 +195,13 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
...
       this.walkAllFiles(input, visitor)
 
-      if (files.length || dirs.length) {
-        dirs.map(file => path.resolve(output, path.relative(input, file)))
-          .forEach(dir => mkdirp.sync(dir))
-
-        files.map(file => path.relative(input, file))
-          .forEach(file => { copySync(path.join(input, file), path.join(output, file), { overwrite: false }) })
+      if (output) {
+        const globOptions = { dot: true, ignore: ['**/.git', `**/${path.basename(output)}`] }
```
coreyfarrell: I suspect this is a problem for `nyc instrument . output` if `./lib/output/data.json` exists (that output folder should be copied).  Instead of ignoring `**/${path.basename(output)}` couldn't we just ignore `output`?

test/nyc-integration.js
```diff
@@ -693,6 +693,23 @@ describe('the nyc cli', function () {
         })
       })
 
+      it('can instrument the root directory', function (done) {
```
coreyfarrell: I know it's just a test but can we say call it the project directory instead of the root directory?

JaKXz:
@coreyfarrell I'm going to defer my review to @bcoe, I'm not quite familiar enough atm and don't have the cycles this week

index.js
AndrewFinlay: This only produces a list of top level files using ``glob.sync(`${path.resolve(input)}/*`, ...)``, so the inclusion of ``**/${path.basename(output)}`` won't exclude any further sub-dirs.  The recursive copy of each top level sub-dir is then handled by `fs-extra.copySync`.  I agree it looks a little silly and should really be just `output`, but I couldn't get it to work with `glob.sync`.  I should point out that this also means we're only ignoring `.git` directories at the top level in this implementation.

Anyway I'll see if I can come up with a nicer solution here, in the meantime I'll drop in the change below and rebase

coreyfarrell: What if someone does `nyc instrument . subdir/output`?  In that case wouldn't we need to ignore `subdir`, not `output`?  I ask due to the use of `path.basename`.

AndrewFinlay:
This PR has been split up so that it only adds support for `nyc instrument --include --exclude`. `nyc instrument` copy all functionality will be in a separate PR.

coreyfarrell:
@AndrewFinlay Thanks for working with me to figure out scope of this change and how to deal with it.

PR #1008, opened 2019-03-01
docs: update issue template

JaKXz:
When tabrindle/envinfo#90 + tabrindle/envinfo#93 is released (and fixed) this will be 🔥

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,21 +1,24 @@
 <!--Please use the template provided below when reporting bugs:-->
-## Link to bug demonstration repository.
+## Link to bug demonstration repository
 <!--
 A demo repository will show the combination of module versions being used,
 command-lines and configurations.  NYC is very complex, it is difficult or
 even impossible to triage most bugs without this information.
 
 If you cannot share a link to your repository (maybe it's private) then you must
-create a basic `nyc-bug-demo` repository and link to that please.
+create a basic `nyc-bug-demo` repository and link to that please [mandatory].
 -->
 
 ## Expected Behavior
 
 ## Observed Behavior
 
+### Troubleshooting steps
+- [ ] still occurrring when I put `cache: false` in my nyc config
+
 ## Environment Information
 <!--
-run the following script: npx envinfo --preset nyc
+[mandatory] run the following script: npx envinfo --preset nyc
 
 OR:
 npm i -g envinfo || yarn global add envinfo
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -13,13 +13,11 @@ create a basic `nyc-bug-demo` repository and link to that please.
 
 ## Observed Behavior
 
-## Forensic Information
+## Environment Information
 
-**Operating System:** the operating system you observed the issue on.
-**Environment Information:** information about your project's environment, see instructions below:
+1. run the following script: `npx envinfo --preset nyc`
 
-1. run the following script:
+```
+# paste the output here
 
-sh -c 'node --version; npm --version; npm ls' > output.txt
-
-2. share a gist with the contents of output.txt.
+```
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -18,11 +18,7 @@ create a basic `nyc-bug-demo` repository and link to that please [mandatory].
 
 ## Environment Information
 <!--
-[mandatory] run the following script: npx envinfo --preset nyc
-
-OR:
-npm i -g envinfo || yarn global add envinfo
-envinfo --preset nyc
+[mandatory] run the following script: npx envinfo@latest --preset nyc
 -->
 ```
 # paste the output here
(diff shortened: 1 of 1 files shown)


coreyfarrell:
I think this is a great idea once the pattern matching like `*istanbul*` works correctly.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,30 @@
...
-**Operating System:** the operating system you observed the issue on.
-**Environment Information:** information about your project's environment, see instructions below:
-
-1. run the following script:
+## Environment Information
+<!--
+[mandatory] run the following script: npx envinfo --preset nyc
 
-sh -c 'node --version; npm --version; npm ls' > output.txt
+OR:
```
bcoe: I'd honestly just point out the `npx` command. this is really slick.

coreyfarrell:
I just tried installing `envinfo` again, got the following output:
```
  System:
    OS: Linux 4.20 Fedora 28 (Twenty Eight) 28 (Twenty Eight)
    CPU: (8) x64 Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
    Memory: 5.95 GB / 31.26 GB
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
    Yarn: 1.13.0 - ~/.nvm/versions/node/v10.15.3/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.3/bin/npm
  npmPackages:
    *babel*: Not Found
    *istanbul*: Not Found
    nyc: ^13.3.0 => 13.3.0 
    source-map-support: Not Found
    ts-node: Not Found
    typescript: Not Found
```
If it can't find `*istanbul*` or `*babel*` then maybe we need to list some individual packages?  In particular even if the glob based package matching worked I'm not sure it's a good idea to list all babel packages (usually 40+ babel packages).  Maybe just `@babel/core` and `@babel/register`?

So I'm not sure this is ready.  I do think we should list out the individual istanbul package versions, if envinfo does the glob or if another PR needs to be sent to envinfo to list out each important istanbul package.

JaKXz:
@coreyfarrell we are waiting on a fix from @tabrindle [or a willing contributor 😄]. You are welcome to make a PR to envinfo to change the nyc preset but I think we should have all that information tbph, because e.g. some people might have babel 6 and that's a good red flag in the troubleshooting steps

.github/ISSUE_TEMPLATE.md
JaKXz: Just wanted to accommodate those who might not have npx yet... do you feel strongly about this though? If you do I will make the change.

tabrindle:
Up to y'all on how you want to do this until I can add some more functionality to envinfo - here's one solution. https://github.com/tabrindle/envinfo/pull/93

BTW, this is how vue-cli handles it for now.

coreyfarrell:
tabrindle/envinfo#93 is definitely a big improvement.

@JaKXz do you think we need to tell people to use `envinfo --preset nyc --fullTree`?  Without fullTree this will not show any module that is not specified by the local package.json (istanbul-lib-instrument for example).  This might not be a big deal for older versions of nyc that bundle everything but for nyc 14+ won't bundle so it'll be possible for istanbul-lib-instrument to be updated without a new nyc release.

.github/ISSUE_TEMPLATE.md
coreyfarrell: npx is included in node.js 8+.  nyc@14 will support node.js 6 so I lean towards leaving the global install method listed for now.  I think we should remove the instructions at some point in the future, maybe once we have a second semver-major that does not support node.js 6?

Side note: I actually have `alias npx='npx --no-install'` in my bashrc, so I have to use `npm i -g envinfo` even through I'm running node.js 10.  That said I know that I explicitly disabled the auto-install function of npx so I also know how to deal with the consequences of that.

One potential concern for the global install, will `yarn global add envinfo` update an existing envinfo package?  I know `npm i -g envinfo` does update.

coreyfarrell:
I just realized that the preset could specify `options: {fullTree: true}` and this would work.  Do we want that?

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,30 @@
...
+### Troubleshooting steps
+- [ ] still occurrring when I put `cache: false` in my nyc config
 
-**Operating System:** the operating system you observed the issue on.
-**Environment Information:** information about your project's environment, see instructions below:
-
-1. run the following script:
+## Environment Information
+<!--
+[mandatory] run the following script: npx envinfo --preset nyc
```
tabrindle: because envinfo is included in a LOT of projects now, you may want to specify a version or `npx envinfo@latest`, or you may get an old version before your preset was added.

.github/ISSUE_TEMPLATE.md
JaKXz: I removed it, just to make it more consistent, using `npx envinfo@latest` as @bcoe and @tabrindle suggested :)

tabrindle:
Published in envinfo@7.2.0

`npx envinfo@7.2.0 --preset nyc`

PR #1009, opened 2019-03-04
docs: Update description of how files are selected for instrumentation

AndrewFinlay:
So I've put together a draft to better explain how files are included for coverage in nyc.

Take a look and tell me what you think, it probably needs a few changes here or there but I think it will take out a little of the mystery in configuring nyc.


README.md
```diff
@@ -195,52 +195,78 @@ Install custom reporters as a development dependency and you can use the `--repo
...
 
-You can tell nyc to exclude specific files and directories by adding
-an `nyc.exclude` array to your `package.json`. Each element of
-the array is a glob pattern indicating which paths should be omitted.
+By default, nyc only collects coverage for source files that are visited
+during a test.  It does this by watching for files that are `require()`'d 
+during the test.  Only source files that are visited during a test will appear 
+in the coverage report and contribute to coverage statistics.  
+
+Nyc will cover all files if the `--all` flag is set.  In this case all files will
```
JaKXz: nyc will instrument all files...

README.md
```diff
@@ -195,52 +195,78 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+Nyc will only cover files that are located under `cwd`, and then only `*.js` files
+or files with extensions listed in in the `nyc.extension` array.
+
+You can reduce the set of covered files by adding `include` and `exclude` 
+filter arrays to your config.  These allow you to shape the set of covered files 
+by specifying glob patterns that can filter files from  the covered set.  The `exclude` 
+array may also use exclude negated glob patterns, these are specified with 
+a `!` prefix, and can restore sub-paths of excluded paths.
 
 Globs are matched using [minimatch](https://www.npmjs.com/package/minimatch).
```
JaKXz: I hope we've switched to micromatch by now?

README.md
```diff
@@ -195,52 +195,78 @@ Install custom reporters as a development dependency and you can use the `--repo
...
-```
-> Note: Since version 9.0 files under `node_modules/` are excluded by default.
-  add the exclude rule `!**/node_modules/` to stop this.
 
-> Note: exclude defaults to `['coverage/**', 'test/**', 'test{,-*}.js', '**/*.test.js', '**/__tests__/**', '**/node_modules/**']`,
+##### Using include and exclude arrays
+
+If there are paths specified in the `include` array, then the set of 
+covered files will be limited to eligible files found in those paths.
+If the `include` array is left empty then all eligible files will be
```
JaKXz: This might read like setting `include: []` is equivalent to `include: ['**']`, which I don't think is intended. I would change it to: If the `include` array is left undefined...

README.md
```diff
@@ -195,52 +195,78 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+already been excluded in the `exclude` array.  Exclude options can be
+specified on the command line with the `-x` switch.
+
+> Note: Exclude defaults to `['coverage/**', 'test/**', 'test{,-*}.js', '**/*.test.js', '**/__tests__/**', '**/node_modules/**']`,
 which would exclude `test`/`__tests__` directories as well as `test.js`, `*.test.js`,
 and `test-*.js` files. Specifying your own exclude property overrides these defaults.
 
-## Including files
+> Note: Since version 9.0, files under `node_modules/` are always excluded by nyc.
+  To reverse this you must add the negated exclude rule `!**/node_modules/`.
```
JaKXz: you also have to add back all the other defaults if you want them

README.md
```diff
@@ -195,52 +195,78 @@ Install custom reporters as a development dependency and you can use the `--repo
...
   }
 }
 ```
 
-> `nyc` uses minimatch for glob expansions, you can read its documentation [here](https://www.npmjs.com/package/minimatch).
-
-> Note: include defaults to `['**']`
-
-> ### Use the `--all` flag to include files that have not been required in your tests.
+> Note: Be wary of automatic OS glob expansion when specifying include/exclude globs with the CLI.
```
JaKXz: I think the `>` blockquoted Note: this pattern is actually less clear as I've learned from originally writing this. I think it'd be best to just let it be normal body text.

Maybe bold the **Note:** to call attention to it

README.md
```diff
@@ -195,52 +195,78 @@ Install custom reporters as a development dependency and you can use the `--repo
...
-      "build"
-    ]
-  }
-}
-```
-> Note: Since version 9.0 files under `node_modules/` are excluded by default.
-  add the exclude rule `!**/node_modules/` to stop this.
 
-> Note: exclude defaults to `['coverage/**', 'test/**', 'test{,-*}.js', '**/*.test.js', '**/__tests__/**', '**/node_modules/**']`,
+##### Using include and exclude arrays
```
JaKXz: Shouldn't this just be an h3 to be consistent?

README.md
AndrewFinlay: I toyed with instrument there, but thought it may be a little too specific.  You're right though, only tests can cover source.

README.md
AndrewFinlay: Both `test-exclude` and `glob` use `minimatch` to do globbing

README.md
AndrewFinlay: Cool, I'll put that in.

README.md
AndrewFinlay: The note above mentions that exclude will override the defaults, is that enough?

README.md
AndrewFinlay: Cool, I'll try that out tomorrow

README.md
AndrewFinlay: I wanted it to seem like a very minor heading in the scope of the whole section.  But it can change to h3

AndrewFinlay:
@JaKXz thanks for the review, 

I'll add the [skip ci] in future, simply didn't know it existed.

The line endings are simply an artefact of how I work with my editor.  I'll fix it up tomorrow

README.md
coreyfarrell: We switched from micromatch to minimatch on `test-exclude@5` and `nyc@13`.  We were always using `glob` which as @AndrewFinlay points out uses `minimatch`.  I'm not strictly against switching back to `micromatch` but would object to doing this without switching to a `micromatch` based globber.  Obviously this could only be done as part of a semver-major release due to potential risk of changes to edge cases.

README.md
coreyfarrell: @AndrewFinlay See istanbuljs/istanbuljs#213 with regard to this.  Once this is merged we'll want to make NYC support the new option and document that as the preferred way since correctly processing negated excludes can be costly for the `--all` option.

README.md
JaKXz: Hmm I thought minimatch was deprecated at some point in favour of micromatch? Weren't there some security issues or something?

coreyfarrell: From `micromatch` readme:
> not subject to DoS with brace patterns, like minimatch and multimatch

This appears to be an outdated statement, `npm info minimatch@2` says:
> DEPRECATED ⚠️  - Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue

We use 3.0.4 (as does glob).

README.md
AndrewFinlay: @JaKXz I've modified the preceding section to 'hopefully' handle this issue.
@coreyfarrell Sounds like this will have to be expanded on when that drops.  I figure this is a little broken at the moment anyway so it could be shelved until it's fixed.

JaKXz:
This is odd... the single newlines are being rendered as `<br/>`s... not sure why that's happening.

coreyfarrell:
> This is odd... the single newlines are being rendered as <br/>s... not sure why that's happening.

@JaKXz where are you seeing this?

JaKXz:
@coreyfarrell in the rich diff

README.md
```diff
@@ -195,52 +195,76 @@ Install custom reporters as a development dependency and you can use the `--repo
...
-```json
-{
-  "nyc": {
-    "exclude": [
-      "**/*.spec.js",
-      "build"
-    ]
-  }
-}
+You can reduce the set of covered files by adding `include` and `exclude` filter arrays to your config.
```
coreyfarrell: I think the trailing spaces here are causing markdown to add `<br>` after this line and a few others.

README.md
```diff
@@ -195,52 +195,75 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+[
+  'coverage/**',
+  'test/**',
+  'test{,-*}.js',
+  '**/*.test.js',
+  '**/__tests__/**',
+  '**/node_modules/**'
+]
+```
+These settings exclude `test`/`__tests__` directories as well as `test.js`, `*.test.js`, and `test-*.js` files.
```
coreyfarrell: Lets say "`test` and `__tests__` directories" to avoid using the slash between two folders.

Also the default list is a bit outdated.  From [test-exclude](https://github.com/istanbuljs/istanbuljs/blob/master/packages/test-exclude/index.js):
```js
exportFunc.defaultExclude = [
    'coverage/**',
    'packages/*/test/**',
    'test/**',
    'test{,-*}.js',
    '**/*{.,-}test.js',
    '**/__tests__/**',
    '**/node_modules/**',
    '**/babel.config.js'
];
```

README.md
```diff
@@ -195,52 +195,75 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+Specifying your own exclude property completely replaces these defaults.
 
-## Including files
+**Note:** Since version 9.0, files under `node_modules/` are always excluded by nyc.
```
coreyfarrell: Do we need to specify `Since version 9.0`?  That's over 2 years old, feel like we can safely remove the version qualifier.  @JaKXz do you agree?

README.md
```diff
@@ -195,52 +195,75 @@ Install custom reporters as a development dependency and you can use the `--repo
...
 
-You can tell nyc to exclude specific files and directories by adding
-an `nyc.exclude` array to your `package.json`. Each element of
-the array is a glob pattern indicating which paths should be omitted.
+By default, nyc only collects coverage for source files that are visited during a test.
+It does this by watching for files that are `require()`'d during the test.
+Only source files that are visited during a test will appear in the coverage report and contribute to coverage statistics.
 
-Globs are matched using [minimatch](https://www.npmjs.com/package/minimatch).
+Nyc will instrument all files if the `--all` flag is set.
```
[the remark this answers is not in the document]
AndrewFinlay: I was aiming for consistency, this slipped through

README.md
```diff
@@ -195,52 +195,75 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+
+Globs are matched using [minimatch](https://www.npmjs.com/package/minimatch).
 
-> Note: exclude defaults to `['coverage/**', 'test/**', 'test{,-*}.js', '**/*.test.js', '**/__tests__/**', '**/node_modules/**']`,
-which would exclude `test`/`__tests__` directories as well as `test.js`, `*.test.js`,
-and `test-*.js` files. Specifying your own exclude property overrides these defaults.
+We use the following process to remove files from consideration,
+ * First, limit the set of covered files to those files in paths listed in the `include` array.
+ * Then, remove any files that are found in the `exclude` array.
+ * Finally, restore any exclude negated files that have been excluded in the second step
```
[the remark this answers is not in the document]
AndrewFinlay: The exclude negates can only restore files that have been removed by the `exclude`s.  It doesn't have any effect on files removed by the `include`s.

README.md
JaKXz: I feel like we flipped back and forth on this over the last few versions but I'm not definitively sure about that...

README.md
AndrewFinlay: Oh, I didn't check that, that was slight rearrange of what's already there.  I'll take a look.

README.md
AndrewFinlay: Again this was a copy paste from the existing.  I'll change it when we hear from @JaKXz.

README.md
```diff
@@ -195,52 +195,75 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+It does this by watching for files that are `require()`'d during the test.
+Only source files that are visited during a test will appear in the coverage report and contribute to coverage statistics.
 
-Globs are matched using [minimatch](https://www.npmjs.com/package/minimatch).
+nyc will instrument all files if the `--all` flag is set.
+In this case all files will appear in the coverage report and contribute to coverage statistics.
 
-For example, the following config will exclude any files with the extension `.spec.js`,
-and anything in the `build` directory:
+nyc will only cover files that are located under `cwd`, and then only `*.js` files or files with extensions listed in in the `extension` array.
```
JaKXz: this should be instrument, or "collect coverage for files..."

README.md
```diff
@@ -195,52 +195,75 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+These settings exclude `test`/`__tests__` directories as well as `test.js`, `*.test.js`, and `test-*.js` files.
+Specifying your own exclude property completely replaces these defaults.
 
-## Including files
+**Note:** Since version 9.0, files under `node_modules/` are always excluded by nyc.
+To reverse this you must add the negated exclude rule `!**/node_modules/`.
 
-As an alternative to providing a list of files to `exclude`, you can provide
-an `include` key with a list of globs to specify specific files that should be covered:
+For example, the following config will only collect coverage for files in the `src` directory, and exclude any files with the extension `.spec.js`.
```
JaKXz: "collect coverage for every file in the src directory [regardless of whether or not it is require'd in a test]..." 

that might be a bit verbose but I think is clearer and more correct.

README.md
JaKXz: yep - I did not see them hence my confusion! Good eye!

README.md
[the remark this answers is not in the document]
coreyfarrell: I see from the code that it does the right thing, I'm still going to add a test to make CI verify this behavior doesn't change in the future.

README.md
JaKXz: Anyway my answer is I agree with @coreyfarrell, I just wonder if we flipped between them in a version after 9.0.

Anyway with the correct list of default excludes is this even necessary?

coreyfarrell: This comment is needed because `test-exclude` always adds `**/node_modules/**` back into exclude list.  So `exclude: ['something.js']` would allow `test/**` to be instrumented but still would not allow anything matching `**/node_modules/**` to be instrumented.  istanbuljs/istanbuljs#213 is meant to allow this to be controlled - we'll need to add an option to nyc once a new test-exclude is released.  Once that happens this section will get a small update.

For the defaults list I think we could link it if we moved the default excludes to a separate source file in `test-exclude`, just link the file.  The issue with linking lines is that the line numbers can change (especially since the defaults list is at the end of the file).

AndrewFinlay: This statement is important as it's the only default exclude that isn't overridden by specifying your own `exclude`s

JaKXz: > isn't overridden by specifying your own `exclude`s

except a negated exclude right?

coreyfarrell: How about:
> **Note**: The exclude list always implicitly contains `**/node_modules/**` even if not specified in the overridden list.

AndrewFinlay: @JaKXz Yeah, it can be changed with a negated exclude
@coreyfarrell sounds good, but I'll leave in the following line about exclude negation and `node_modules`

AndrewFinlay: @JaKXz Regarding the list of default excludes, it looks like the GH maintainers don't like the idea of linking source to md https://github.com/github/markup/issues/346.  So we can change it to the correct list or a link to the where the excludes are defined in `test-exclude`

coreyfarrell: @AndrewFinlay yes we'll update this section of the doc when we add the option to nyc.

PR #1010, opened 2019-03-06
feat: Fix excludeAfterRemap functionality.

### LINKED ISSUE #956: How to ignore coverage data (instrumented code) from dependencies?
## Link to bug demonstration repository.

https://github.com/henriquemotaesteves/nyc-bug-demo

## Expected Behavior

How to ignore coverage data (instrumented code) from dependencies?

## Observed Behavior

If a dependency is packed and published with coverage data (instrumented code), then these data are included in the coverage report.

## Forensic Information

https://gist.github.com/henriquemotaesteves/2bcb39a27e1beda6a8a22c47805c8f95

[1] - https://github.com/webpack-contrib/istanbul-instrumenter-loader
[2] - https://github.com/dolanmiu/docx

coreyfarrell:
I'm not sure if this qualifies as a feature or a fix, filtering after remapping has never worked.  Original implementation of filtering remapped sources at dd40dc51, addition of excludeAfterRemap option at cdfdff36.  Unfortunately the filtering step was done before coverage was remapped using source-maps, so it didn't do anything.

This is a potentially breaking change - bundled / compiled code was never filtered based on the original source filename, after this it will be.  As an auxiliary change `test-exclude` options are also added to the `nyc report` and `nyc check-coverage` as those commands should process exclude-after-remap.

Fixes #956

### DIFF
--- a/lib/commands/check-coverage.js
+++ b/lib/commands/check-coverage.js
@@ -1,3 +1,4 @@
+const testExclude = require('test-exclude')
 const NYC = require('../../index.js')
 
 exports.command = 'check-coverage'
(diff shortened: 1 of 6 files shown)


bcoe:
@coreyfarrell see my comments [here](https://github.com/istanbuljs/istanbuljs/issues/318); do we just bite the bullet and move away from the `--exclude-after-remap` command; perhaps it would give us more flexibility for a variety of configurations if we broke this out into its own configuration setting?

lib/commands/check-coverage.js
```diff
@@ -6,6 +7,24 @@ exports.describe = 'check whether coverage is within thresholds provided'
...
+    })
+    .option('exclude-after-remap', {
+      default: true,
+      type: 'boolean',
+      description: 'should exclude logic be performed after the source-map remaps filenames?',
+      global: false
+    })
+    .option('include', {
+      alias: 'n',
+      default: [],
```
JaKXz: @coreyfarrell sorry for the post-review - shouldn't this be `['**']`?

coreyfarrell: Nope.  An empty list is functionally equal to `['**']` but `[]` actually bypasses the check so it's more efficient.

https://github.com/istanbuljs/istanbuljs/blob/master/packages/test-exclude/index.js#L50-L54 is where the default is processed, `[]` is replaced with `false` which short-circuits the include check.

PR #1012, opened 2019-03-07
Fix unable lookup source

### LINKED ISSUE #1014: nyc instrument should resolve source filenames.
Create ./somefile.js then run `nyc instrument somefile.js`.  The instrumented code produces coverage data for the filename `'somefile.js'` when it really should be producing coverage for `path.resolve('somefile.js')`.  This causes oddness in reporting when combining pre-instrumented code with code that nyc instruments on the fly (during execution).

CC @AndrewFinlay

golonikum:
Fixes #1014

### DIFF
--- a/index.js
+++ b/index.js
@@ -212,7 +212,7 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
     }
 
     if (transform) {
-      code = transform(code, {filename: path.resolve(inputDir, filename), relFile: inFile})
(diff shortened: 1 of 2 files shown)


coreyfarrell:
@golonikum thanks for the contribution.  I've updated the tests to work with the corrected results and I made a small tweak to your code to use a variable which had already resolved the filename.

@bcoe @JaKXz I didn't remove it but I think the `relFile` passed to the transform function is unused.  Possibly a remnant from days past, maybe from before `caching-transform` was split into a separate module.

PR #1015, opened 2019-03-07
chore: adds prettier-standard with pre-commit hook

JaKXz:
To be consistent with the istanbuljs monorepo, and to make further contributions more convenient.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,9 +99,12 @@
     "bundle-dependencies": "^1.0.2",
     "chai": "^4.2.0",
     "coveralls": "^3.0.2",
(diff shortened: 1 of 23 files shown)


package.json
```diff
@@ -143,5 +146,13 @@
     "uuid",
     "yargs",
     "yargs-parser"
-  ]
+  ],
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
```
coreyfarrell: I'm really not a fan of projects installing git-hooks :-1:

PR #1023, opened 2019-03-12
chore: update instrument code to use es2015+

AndrewFinlay:
This is an attempt to use some more modern language features in the `instrument.js` and the `instrumentAllFiles` command.  It's a cut back version of what I was proposing in #1003 as per conversation with @coreyfarrell.

As there's quite a bit here I'll drop in some notes and assumptions
* Notes:
1. Refactor `nyc-integration` instrument tests so tests with output folder specified aren't in the no output folder suite
2. Refactor `instrument.js` and `instrumentAllFiles` command to use modern javascript
3. Remove call to deprecated `process.exit()` in instrument command
4. Removed unused transform metadata param `relFile`, _this is an issue in other parts of the code_.
5. Use file extension as an index to find the correct transform, rather than trying to match a transform name to the rightmost characters of a filename, _this may be an issue in other parts of the code_.

* Assumptions:
1. That we want to use file extensions to find transforms, the existing implementation would use the `.js` instrumenter for both `.js` and `.mjs` files, these changes assume that was wrong.
2. When `instrumentAllFiles` fails to find the correct transform, it returns the uninstrumented source,
3. It looks like the `relFile` property isn't used anymore, so I'm assuming it can be safely removed

### DIFF
--- a/index.js
+++ b/index.js
@@ -216,13 +216,12 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
     if (stats.isDirectory()) {
       this.walkAllFiles(input, visitor)
     } else {
-      const inputPath = path.resolve('.', input)
-      visitor(inputPath, path.basename(input))
+      visitor(path.resolve(input), path.basename(input))
     }
   } catch (err) {
     return cb(err)
   }
-  return cb()
+  cb()
 }
 
 NYC.prototype.walkAllFiles = function (dir, visitor) {
--- a/index.js
+++ b/index.js
@@ -191,48 +191,40 @@ NYC.prototype.addAllFiles = function () {
 }
 
 NYC.prototype.instrumentAllFiles = function (input, output, cb) {
-  var _this = this
-  var inputDir = '.' + path.sep
-  var visitor = function (filename) {
-    var ext
-    var transform
-    var inFile = path.resolve(inputDir, filename)
-    var code = fs.readFileSync(inFile, 'utf-8')
+  const visitor = (filename, singleFile = false) => {
+    const inputPath = path.resolve(this.cwd, input, filename)
+    const inCode = fs.readFileSync(inputPath, 'utf-8')
 
-    for (ext in _this.transforms) {
-      if (filename.toLowerCase().substr(-ext.length) === ext) {
-        transform = _this.transforms[ext]
-        break
-      }
-    }
+    const extname = path.extname(filename).toLowerCase()
+    const transform = this.transforms[extname] || false
+    const outCode = (transform)
+      ? transform(inCode, { filename: `./${path.relative(this.cwd, inputPath)}` })
+      : inCode
 
-    if (transform) {
-      code = transform(code, { filename: inFile, relFile: inFile })
-    }
-
-    if (!output) {
-      console.log(code)
+    if (output) {
+      const subpath = singleFile || path.relative(input, path.resolve(input, filename))
+      const outputPath = path.resolve(this.cwd, output, subpath)
+      mkdirp.sync(path.dirname(outputPath))
+      fs.writeFileSync(outputPath, outCode, 'utf-8')
     } else {
-      var outFile = path.resolve(output, filename)
-      mkdirp.sync(path.dirname(outFile))
-      fs.writeFileSync(outFile, code, 'utf-8')
+      console.log(outCode)
     }
   }
 
   this._loadAdditionalModules()
 
   try {
-    var stats = fs.lstatSync(input)
+    const inputPath = path.resolve(this.cwd, input)
+    const stats = fs.lstatSync(inputPath)
     if (stats.isDirectory()) {
-      inputDir = input
-      this.walkAllFiles(input, visitor)
+      this.walkAllFiles(inputPath, visitor)
     } else {
-      visitor(input)
+      visitor(inputPath, path.basename(inputPath))
     }
   } catch (err) {
     return cb(err)
   }
-  cb()
+  return cb()
 }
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -191,48 +191,38 @@ NYC.prototype.addAllFiles = function () {
...
 
   try {
-    var stats = fs.lstatSync(input)
+    const stats = fs.lstatSync(input)
     if (stats.isDirectory()) {
-      inputDir = input
       this.walkAllFiles(input, visitor)
     } else {
-      visitor(input)
+      const inputPath = path.resolve('.', input)
```
coreyfarrell: The `'.'` argument is redundant / always implied by `path.resolve`.  Though if this will be replaced by the `--cwd` option then don't worry about change it.

index.js
```diff
@@ -191,48 +191,38 @@ NYC.prototype.addAllFiles = function () {
...
+    const extname = path.extname(filename).toLowerCase()
+    const transform = this.transforms[extname] || (code => code)
+    const outCode = transform(inCode, { filename: inFile })
 
-    if (!output) {
-      console.log(code)
-    } else {
-      var outFile = path.resolve(output, filename)
+    if (output) {
+      const subPath = singleFile || path.relative(input, path.resolve(input, filename))
```
coreyfarrell: Could you explain what is going on with this change?

index.js
```diff
@@ -191,48 +191,38 @@ NYC.prototype.addAllFiles = function () {
...
     } else {
-      visitor(input)
+      const inputPath = path.resolve('.', input)
+      visitor(inputPath, path.basename(input))
     }
   } catch (err) {
     return cb(err)
   }
-  cb()
+  return cb()
```
coreyfarrell: I'd prefer we not do this change as `instrumentAllFiles` does not have a return.  My preference would be for the `return cb(err)` in the catch block be split into two statements:
```js
cb(err)
return
```
This way the code self documents lack of return value.

lib/commands/instrument.js
```diff
@@ -62,12 +62,12 @@ exports.builder = function (yargs) {
 }
 
 exports.handler = function (argv) {
-  // if instrument is set to false,
-  // enable a noop instrumenter.
-  if (!argv.instrument) argv.instrumenter = './lib/instrumenters/noop'
-  else argv.instrumenter = './lib/instrumenters/istanbul'
+  // If instrument is set to false enable a noop instrumenter.
+  argv.instrumenter = (!argv.instrument)
```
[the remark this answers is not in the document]
AndrewFinlay: Missed that, thanks.

test/nyc-integration.js
```diff
@@ -581,6 +581,12 @@ describe('the nyc cli', function () {
           done()
         })
       })
+    })
+
+    describe('output folder specified', function () {
+      afterEach(function () {
+        rimraf.sync(path.resolve(fixturesCLI, 'output'))
```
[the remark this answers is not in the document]
AndrewFinlay: What do you think of adding `rimraf.sync(path.resolve(fixturesCLI, 'subdir', 'output-dir'))` to the `afterEach`?

index.js
AndrewFinlay: I'll change it for correctness

index.js
AndrewFinlay: I'm assuming that if an output dir is set, and we instrument a single file, then the new file should be written to `output/path.basename(inputFilename)`.  If singleFile is set to false then we're instrumenting a set of files, so the `path.relative(input, path.resolve(input, filename)` works to maintain the existing directory structure in the resulting instrumented directory.  It aims to replace the `var inputDir` variable that's currently reassigned in the case of instrumenting a directory.

So all that being said I'm not really happy with this solution, so I can have a go at reworking it and would be open to suggestions on fixing it.

index.js
AndrewFinlay: Cool, I'll change that back, I wasn't too fond of that change.

test/nyc-integration.js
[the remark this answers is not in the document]
AndrewFinlay: On redirecting output from `fixtures/cli/subdir/output-dir/` to `fixtures/cli/output`, there are no more test uses of `output-dir`.  Did you want the `output-dir` cleanup code removed as well?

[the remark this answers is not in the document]
coreyfarrell: Sorry I may have been wrong in my initial comment.  I thought that the `output-dir` outputs all failed, looks like they do succeed.  Also `path.resolve(fixturesCLI, 'subdir', 'output-dir')` is covered by a .gitignore file so maybe it's better if we use that across the board?  Sorry for mixing this up.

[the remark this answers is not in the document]
AndrewFinlay: No worries, will roll this back to what it was.

index.js
AndrewFinlay: I'll do a little more digging on this, but I'm currently thinking there is no behaviour change here.

AndrewFinlay: Ah, ok that is breaking, it does preserve full path there.  Not sure what's desired in the long run but I'll get it back to behaving as is for now.

AndrewFinlay: Does this need a test?

coreyfarrell: Please drop the subPath calculation for now (as well as the singleFile argument).  I would like to make sure we have tests dealing with this but I think logic change should not be part of this ES2015 refactor.  For now lets just do:
```js
const outFile = path.resolve(output, filename)
```

AndrewFinlay: So I've taken that back to be much closer to the original implementation

PR #1024, opened 2019-03-12
feat: add `cwd` option to es6+ instrument command

AndrewFinlay:
This adds support and tests for the '--cwd' option for the instrument command, it's intended to sit on top of the recent es6+ changes.


lib/commands/instrument.js
```diff
@@ -28,6 +29,10 @@ exports.builder = function (yargs) {
       type: 'boolean',
       description: "should nyc's instrumenter produce source maps?"
     })
+    .option('cwd', {
```
coreyfarrell: `cwd` is actually already defined in `lib/config-util.js` in a way that allows it to be global.  I think we should not add the `.option` to this file but we should update the `describe` string in `lib/config-util.js`.  Note the current definition of `lib/config-util.js` often does not result in `cwd: process.cwd()`.  It uses `find-up` to search for the `package.json` which owns `process.cwd()`, then uses the dirname of that `package.json`.

Existing: `working directory used when resolving paths`
Maybe change to: `base directory of this project`

The important thing is that `--cwd` effects the interpretation of `include` and `exclude`.  I think it should not effect the interpretation of the input/output positional arguments.  Consider this, when you run `nyc --cwd=.. ./somescript.js` the `./somescript.js` is found from $PWD, not from the value of `--cwd`.

lib/commands/instrument.js
```diff
@@ -73,6 +78,7 @@ exports.handler = function (argv) {
     produceSourceMap: argv.produceSourceMap,
     extension: argv.extension,
     require: argv.require,
+    cwd: argv.cwd,
```
coreyfarrell: This we do need to properly configure the `test-exclude` instance used to determine which files should be instrumented.  This makes a difference if `package.json` is at `..` (you are in a subdirectory of the project).

index.js
```diff
@@ -193,15 +193,15 @@ NYC.prototype.addAllFiles = function () {
 NYC.prototype.instrumentAllFiles = function (input, output, cb) {
   let inputDir = '.' + path.sep
   const visitor = filename => {
-    const inFile = path.resolve(inputDir, filename)
+    const inFile = path.resolve(this.cwd, inputDir, filename)
```
coreyfarrell: My gut reaction is that we these changes to `index.js` should not happen.

PR #1026, opened 2019-03-13
feat: Pass `extension` to test-exclude, use TestExclude.globSync.

coreyfarrell:
BREAKING CHANGE: The logic involving include/exclude processing has
changed.  Results should be verified to ensure all desired sources have
coverage data.


coreyfarrell:
The testing changes are less significant than it looks.  The tests were actually checking for results that were impossible in practice.  Although `shouldInstrument('file-without-extension')` could previously return `true`, the file would never have been instrumented because it has no extension for the require hook.

PR #1028, opened 2019-03-14
feat: add `--build-process-tree` feature

isaacs:
This allows a user to build the processinfo temp files, without actually
dumping the process tree to stdout.

The goal is to use this in node-tap to see which processes covered which
files, so we know which tests to re-run on changes to the source files.

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,6 +42,7 @@ function NYC (config) {
   this._reportDir = config.reportDir || 'coverage'
   this._sourceMap = typeof config.sourceMap === 'boolean' ? config.sourceMap : true
   this._showProcessTree = config.showProcessTree || false
+  this._buildProcessTree = this._showProcessTree || config.buildProcessTree
   this._eagerInstantiation = config.eager || false
   this.cwd = config.cwd || process.cwd()
   this.reporter = arrify(config.reporter || 'text')
@@ -325,7 +326,7 @@ NYC.prototype.createTempDirectory = function () {
   mkdirp.sync(this.tempDirectory())
   if (this.cache) mkdirp.sync(this.cacheDirectory)
 
-  if (this._showProcessTree) {
+  if (this._buildProcessTree) {
     mkdirp.sync(this.processInfoDirectory())
   }
 }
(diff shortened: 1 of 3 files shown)


coreyfarrell:
@isaacs this produces a parse error in `test/nyc-integration.js`.  Would you mind moving the `describe('--build-process-tree', function () {` out of the `describe('--show-process-tree', function () {`, then add the closing `'})` after your new test (currently you're not closing the added describe function).

isaacs:
@coreyfarrell Oh, oops, some rebase scar tissue got in there, how embarrassing.  Fixed and force-pushed.

lib/commands/report.js
```diff
@@ -47,6 +47,12 @@ exports.builder = function (yargs) {
       default: false,
       type: 'boolean'
     })
+    .option('build-process-tree', {
```
coreyfarrell: Sorry I didn't catch this earlier (not sure if it was here), but this doesn't make sense.  `nyc report` doesn't execute any processes so it can't build a process tree (it can only show it).  IMO it only makes sense to add this option to `lib/config-utils.js` for the 'main' command.

isaacs: Good point!  Either the processinfo is there or it isn't, but `nyc report` can't put it there.  Removed and pushed.

index.js
```diff
@@ -42,6 +42,7 @@ function NYC (config) {
   this._reportDir = config.reportDir || 'coverage'
   this._sourceMap = typeof config.sourceMap === 'boolean' ? config.sourceMap : true
   this._showProcessTree = config.showProcessTree || false
+  this._buildProcessTree = this._showProcessTree || config.buildProcessTree
```
JaKXz: This is the only thing that confuses me... Is there a difference between `build` and `show` here? Is it just a preference for the name `--build-process-tree`?

isaacs: In order to show the process tree, you must also build it.  So `--show-process-tree` implies `--build-process-tree` as well.  But the reverse is not true, ie, you could build it without showing it.

coreyfarrell: @JaKXz any objections if I merge this?

JaKXz:
@coreyfarrell :shipit: 

Thanks @isaacs I was just curious!

PR #1031, opened 2019-03-17
feat: Rename `plugins` option to `parser-plugins`.

### LINKED ISSUE #986: Rename `plugins` option to `parser-plugins`.
At https://github.com/istanbuljs/nyc/issues/947#issuecomment-461350050 @novemberborn pointed out that the `plugins` option is easily confused with normal babel plugins where nyc expects a list of babel parser plugins.  This issue is opened so I don't forget that we should rename this option.

coreyfarrell:
The original option name created confusion about what was expected.
This helps identify that babel parser plugins are expected, not babel
transform plugins.

BREAKING CHANGE: The `plugins` option has been renamed to
`parser-plugins`.

Fixes #986

### DIFF
--- a/index.js
+++ b/index.js
@@ -137,7 +137,7 @@ NYC.prototype._createInstrumenter = function () {
     compact: this.config.compact,
     preserveComments: this.config.preserveComments,
     esModules: this.config.esModules,
-    plugins: this.config.plugins
+    plugins: this.config.parserPlugins
   })
 }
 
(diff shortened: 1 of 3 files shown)


JaKXz:
@coreyfarrell do we have an nyc@14 branch to merge breaking changes to and rebase eventually?

coreyfarrell:
@JaKXz master already has one breaking change merged and #1026 is being labeled as breaking to be on the safe side.  The plan is that the next release will be 14.0.0.

PR #1032, opened 2019-03-18
docs: project root directory and `--cwd` doc

AndrewFinlay:
So I got a little off track while trying to add `instrument --cwd`, so I figured I'd spend some time understanding what it does, opposed to what I think it should do, and put some documentation together in the process.  I'm still not 100% sure this is correct, it's my reading of the intention of the code.  A few notes, I haven't mentioned `env.NYC_CWD` as I figured that it's not really supposed to be directly accessed by users.  I also didn't mention that `cwd` is fed in while creating a new instrumenter, as it looks to be ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ If nyc fails to find a directory containing a `package.json` file, it will use c
 You can change the project root directory with the `--cwd` option.
 
 **Note:** nyc stanza configuration is taken from the `package.json` file in the *default* project root directory.
+Default configuration files must also be in the *default* project root directory, e.g. `.nycrc`, `.nycrc.json`, `nyc.config.js`.
 
 nyc uses the project root directory when:
  * looking for source files to cover
--- a/README.md
+++ b/README.md
@@ -195,6 +195,26 @@ Install custom reporters as a development dependency and you can use the `--repo
 nyc report --reporter=<custom-reporter-name>
 ```
 
+## Setting the project root directory
+
+nyc runs a lot of file system operations relative to the project root directory.
+During startup nyc will look for the *default* project root directory.
+The *default* project root directory is the first directory found that contains a `package.json` file when searching from the current working directory up.
+If nyc fails to find a directory containing a `package.json` file, it will use current working directory.
+You can change the project root directory with the `--cwd` option.
+
+**Note:** nyc stanza configuration is taken from the `package.json` file in the *default* project root directory.
+
+nyc uses the project root directory when:
+ * looking for source files to cover
+ * creating globs for include and exclude rules during file selection
+ * loading custom require hooks from the `require` array
+
+nyc may create artefact directories within the project root, such as:
+  * the report directory, `<project-root>/coverage`
+  * the cache directory, `<project-root>/node_modules/.cache/nyc`
+  * the temp directory, `<project-root>/.nyc_output`
+
 ## Selecting files for coverage
 
 By default, nyc only collects coverage for source files that are visited during a test.
@@ -216,7 +236,6 @@ We use the following process to remove files from consideration:
  1. Limit the set of covered files to those files in paths listed in the `include` array.
  2. Remove any files that are found in the `exclude` array.
  3. Restore any exclude negated files that have been excluded in the second step.
-
 
 ### Using include and exclude arrays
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,6 +195,26 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+You can change the project root directory with the `--cwd` option.
+
+**Note:** nyc stanza configuration is taken from the `package.json` file in the *default* project root directory.
+
+nyc uses the project root directory when:
+ * looking for source files to cover
+ * creating globs for include and exclude rules during file selection
+ * loading custom require hooks from the `require` array
+
+nyc may create artefact directories within the project root, such as:
```
coreyfarrell: The Yankee in me wants to resist use of the British spelling of `artifact` :grinning: (joking, I don't really care that much)

README.md
```diff
@@ -195,6 +195,26 @@ Install custom reporters as a development dependency and you can use the `--repo
...
 
+## Setting the project root directory
+
+nyc runs a lot of file system operations relative to the project root directory.
+During startup nyc will look for the *default* project root directory.
+The *default* project root directory is the first directory found that contains a `package.json` file when searching from the current working directory up.
+If nyc fails to find a directory containing a `package.json` file, it will use current working directory.
+You can change the project root directory with the `--cwd` option.
+
+**Note:** nyc stanza configuration is taken from the `package.json` file in the *default* project root directory.
```
coreyfarrell: Configuration can also be taken from `.nycrc`, `.nycrc.json` or `nyc.config.js` found in the default project root directory.

AndrewFinlay: Ok, they deserve a mention here too.  I'll make the change.

README.md
AndrewFinlay: We've had discussions at work about this one, from my understanding, an artifact is something left behind by an ancient civilisation, while an artefact is the product of a system.  So it's not an issue of Americanisation, but a newer, less popular but more accurate word.  All that being said, Wikipedia states that it's a difference in Americanisation.

README.md
coreyfarrell: For now lets remove mention of loading the configuration from this section.  Configuration loading / precedence can be documented in a section about configuration (in a separate PR).

README.md
coreyfarrell: I've never seen the word `artefact`, initially I was going to point to it as a misspelling but then decided to do a search and found that it was valid in other places.  AFAIK in American English everything left behind is an `artifact` (doesn't matter who/what left it behind).

README.md
```diff
@@ -195,6 +195,24 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+You can change the project root directory with the `--cwd` option.
+
+nyc uses the project root directory when:
+ * looking for source files to cover
+ * creating globs for include and exclude rules during file selection
+ * loading custom require hooks from the `require` array
+
+nyc may create artefact directories within the project root, such as:
+  * the report directory, `<project-root>/coverage`
+  * the cache directory, `<project-root>/node_modules/.cache/nyc`
```
bcoe: in the tool I've been working on called `c8`, we consolidated on the `coverage` folder, rather than having a `.nyc_output` folder, we might want to consider this in future nyc versions.

README.md
```diff
@@ -195,6 +195,24 @@ Install custom reporters as a development dependency and you can use the `--repo
...
+
+nyc runs a lot of file system operations relative to the project root directory.
+During startup nyc will look for the *default* project root directory.
+The *default* project root directory is the first directory found that contains a `package.json` file when searching from the current working directory up.
+If nyc fails to find a directory containing a `package.json` file, it will use current working directory.
+You can change the project root directory with the `--cwd` option.
+
+nyc uses the project root directory when:
+ * looking for source files to cover
+ * creating globs for include and exclude rules during file selection
```
bcoe: I think the root directory potentially matters for sourcemap file lookup too? would need to check that code to be sure.

README.md
coreyfarrell: Do we want to do this for 14.0.0?  This would be a bit of a breaking change but I think it's a good idea.  Do you think `<project-root>/coverage/nyc-tmp`? in place of the `.nyc_output` folder?

README.md
coreyfarrell: I'm not seeing any way that the root directory is being passed to source-maps.

PR #1033, opened 2019-03-18
chore: A few code cleanups

AndrewFinlay:
A few code cleanups that I bumped into over the last few days.  I've tried to keep these changes as basic as possible, that being said please take a closer look at the `addFile` and `createInstrumenter` changes, in both cases I'm assuming that what I'm removing is dead code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -111,17 +111,10 @@ NYC.prototype._disableCachingTransform = function () {
 }
 
 NYC.prototype._loadAdditionalModules = function () {
-  var _this = this
-  this.require.forEach(function (r) {
-    // first attempt to require the module relative to
-    // the directory being instrumented.
-    var p = resolveFrom.silent(_this.cwd, r)
-    if (p) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -328,13 +315,9 @@ NYC.prototype.reset = function () {
...
 
 NYC.prototype._wrapExit = function () {
-  var _this = this
-
   // we always want to write coverage
   // regardless of how the process exits.
-  onExit(function () {
-    _this.writeCoverageFile()
-  }, { alwaysLast: true })
+  onExit(() => { this.writeCoverageFile() }, { alwaysLast: true })
```
coreyfarrell: Could you add the line-feeds back in?  I like that you switched to arrow function but prefer that single-line only be used for arrow expressions like `() => value` or empty functions like `() => {}`.

AndrewFinlay: I was maybe a little over eager there, change made

PR #1036, opened 2019-03-20
chore: code tidy up and new test

AndrewFinlay:
Cleanup instrumentation code again with a focus on absolute and relative paths

The main aim of this has been to clarify whether we're working with relative or absolute file paths, and removing unnecessary transformations. Although I've made a few other 'small' changes here and there.

Key changes:
* Created a new private method `NYC._transform`, common to `_maybeInstrumentSource` and `instrumentAllFiles`.
* Renamed the param in `walkAllFiles` forEach handler to `relFile` to explicitly state the file representation being used.
* Let the `addAllFiles` visitor function rely on `testExclude` to determine which files to instrument
* Test that the instrument command returns unmodified source if there is no transform found for a file extension.


index.js
```diff
@@ -154,40 +153,34 @@ NYC.prototype._readTranspiledSource = function (filePath) {
 }
 
 NYC.prototype.addAllFiles = function () {
-  var _this = this
+  const visitor = relFile => {
```
coreyfarrell: Since this is a one time use function can we keep it in the callback / as an argument to `this.walkAllFiles` but do switch it to arrow function?

index.js
```diff
@@ -212,26 +205,22 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
...
+  const transform = this.transforms[extname] || (() => null)
 
-  var ext, transform
-  for (ext in this.transforms) {
-    if (filename.toLowerCase().substr(-ext.length) === ext) {
-      transform = this.transforms[ext]
-      break
-    }
-  }
+  return transform(code, { filename: filename })
```
coreyfarrell: Please use object short-hand.. `{ filename }`.

index.js
```diff
@@ -212,26 +205,22 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
...
+}
+
+NYC.prototype._maybeInstrumentSource = function (code, filename) {
+  const instrument = this.exclude.shouldInstrument(filename)
```
coreyfarrell: I think it would be simpler / more readable to just do:
```js
if (!this.exclude.shouldInstrument(filename)) {
  return null;
}
```

That way we avoid the ternary statement.

PR #1038, opened 2019-03-21
docs: a bunch of updates to recent docs

AndrewFinlay:
These are things that I didn't like in my original docs, the final straw being finding 'in in' in one of the lines.  Hopefully this makes things better!

Note I also added a change to the list of default `.nycrc` files to include `nyc.config.js`

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,11 +200,11 @@ nyc report --reporter=<custom-reporter-name>
 nyc runs a lot of file system operations relative to the project root directory.
 During startup nyc will look for the *default* project root directory.
 The *default* project root directory is the first directory found that contains a `package.json` file when searching from the current working directory up.
-If nyc fails to find a directory containing a `package.json` file, it will use current working directory.
+If nyc fails to find a directory containing a `package.json` file, it will use current working directory as the *default* project root directory.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -298,7 +299,7 @@ modules should be required in the subprocess collecting coverage:
 
 ## Configuring `nyc`
 
-Any configuration options that can be set via the command line can also be specified in the `nyc` stanza of your package.json, or within a `.nycrc` (or `.nycrc.json`) file:
+Any configuration options that can be set via the command line can also be specified in the `nyc` stanza of your package.json, or within a `.nycrc`, `.nycrc.json`, or `nyc.config.js` file:
 
 **package.json:**
```
JaKXz: could remove this line since we are saying this can go in multiple places...

and come to think of it, it would be really nice if this was actually just a table with the option names, valid input types and values, and a definition... That's a lot of work though so I will happily merge this PR without that!

JaKXz:
This is awesome @AndrewFinlay [my comment is a nitpick that can be handled in a future PR] - I'll just let another set of eyes take a look before I merge. :)

PR #1039, opened 2019-03-21
docs: `nyc instrument` and `--exclude-node-modules`

AndrewFinlay:
Lists basic usage, with a key limitation that the input dir must equal or be a sub-dir of cwd.

I'll probably make a few changes tomorrow, so suggestions are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,8 +209,8 @@ nyc instrument <input> [output]
 The `[output]` directory is optional and can be located anywhere, if it is not set the instrumented code will be sent to `stdout`.
 For example, `nyc instrument . ./output` will produce instrumented versions of any source files it finds in `.` and store them in `./output`.
 
-
-Run `nyc instrument --help` to display a list of available command options.
+Any existing output can be removed by specifying the `--delete` option.
+Run `nyc instrument --help` to display a full list of available command options.
 
 **Note:** `nyc instrument` will not copy the contents of a `.git` folder to the output directory.
 
--- a/README.md
+++ b/README.md
@@ -207,7 +207,8 @@ nyc instrument <input> [output]
 
 `<input>` can be any file or directory within the project root directory.
 The `[output]` directory is optional and can be located anywhere, if it is not set the instrumented code will be sent to `stdout`.
-For example, `nyc instrument . ./output` will copy all files from `.` to `./output` and produce instrumented versions of any source files it finds along the way.
+For example, `nyc instrument . ./output` will produce instrumented versions of any source files it finds in `.` and store them in `./output`.
+
 
 Run `nyc instrument --help` to display a list of available command options.
 
--- a/README.md
+++ b/README.md
@@ -195,6 +195,18 @@ Install custom reporters as a development dependency and you can use the `--repo
 nyc report --reporter=<custom-reporter-name>
 ```
 
+## Producing instrumented source
+
+nyc can produce a set of instrumented source files, suitable for client side deployment, by running:
+
+```bash
+nyc instrument <input> [output]
+```
+
+Where `input` can be any file or directory within the project root directory.
+The `output` directory is optional and can be located anywhere, if it is not set the instrumented code with be sent to `stdout`.
+
+
 ## Setting the project root directory
 
 nyc runs a lot of file system operations relative to the project root directory.
--- a/README.md
+++ b/README.md
@@ -197,7 +197,8 @@ nyc report --reporter=<custom-reporter-name>
 
 ## Producing instrumented source
 
-The nyc `instrument` command can produce a set of instrumented source files, suitable for client side deployment in end to end testing.
+The `nyc instrument` command can produce a set of instrumented source files.
+These files are suitable for client side deployment in end to end testing.
 You can create an instrumented version of your source code by running:
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -274,8 +280,8 @@ The `exclude` option has the following defaults settings:
 These settings exclude `test` and `__tests__` directories as well as `test.js`, `*.test.js`, and `test-*.js` files.
 Specifying your own exclude property completely replaces these defaults.
 
-**Note:** The exclude list always implicitly contains `**/node_modules/**`, even if not explicitly specified in an overriding `exclude` array.
-To reverse this you must add the negated exclude rule `!**/node_modules/`.
+**Note:** We always add `**/node_modules/**` to the exclude list, even if not specified in the config.
+You can override this by setting the `!**/node_modules/` negated exclude rule.
```
coreyfarrell: I've reopened #912.  By the time we release 14.0.0 the correct way to handle this will be to use `--exclude-node-modules=false`.  I'd like for this documentation change to reflect #912 being merged.  So for example you might have nyc.config.js as:
```js
module.exports = {
  all: true
  include: [
    'lib/**',
    'node_modules/@my-org/**'
  ],
  exclude: [
    'node_modules/@my-org/something/unwanted.js',
    '**/test/**'
  ],
  excludeNodeModules: false
}
```
This configuration would instrument everything in `lib/` and `node_modules/@my-org/`, ignore all other node_modules (due to not being included rather than being excluded.  Negated excludes are processed last so `!**/node_modules/**` makes it impossible to exclude any specific files matching that negated exclude.  So `exclude: ['!**/node_modules/@my-org/**', 'node_modules/@my-org/something/unwanted.js']` would not exclude unwanted.js.

README.md
```diff
@@ -197,14 +197,20 @@ nyc report --reporter=<custom-reporter-name>
 
 ## Producing instrumented source
 
-nyc can produce a set of instrumented source files, suitable for client side deployment, by running:
+The nyc `instrument` command can produce a set of instrumented source files, suitable for client side deployment in end to end testing.
```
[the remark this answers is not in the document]
AndrewFinlay: This feels correct to me.  I had a quick look online to find some support for this and it looks like this is an `appositive` being offset from the rest of the sentence.  So I think it's valid, but I broke it into two sentences anyway.

README.md
```diff
@@ -304,6 +301,35 @@ It will also exclude any files with the extension `.spec.js`.
...
+    "all": true,
+    "include": [
+      "lib/**",
+      "node_modules/@my-org/**"
+    ],
+    "exclude": [
+      "node_modules/@my-org/something/unwanted.js",
+      "**/test/**"
+    ],
+    excludeNodeModules: false
```
coreyfarrell: Missing quotes for JSON property name.

README.md
```diff
@@ -304,6 +301,35 @@ It will also exclude any files with the extension `.spec.js`.
 **Note:** Be wary of automatic OS glob expansion when specifying include/exclude globs with the CLI.
 To prevent this, wrap each glob in single quotes.
 
+### Including files within `node_modules`
+
+We always add `**/node_modules/**` to the exclude list, even if not specified in the config.
+You can override this by setting `--exclude-node-modules=false`.
+
+For example, the following config will instrument everything in `lib/` and `node_modules/@my-org/`, except for `node_modules/@my-org/something/unwanted.js`.
```
coreyfarrell: except for `node_modules/@my-org/something/unwanted.js` and files in any folder named `test`.  The idea of the `**/test/**` exclude is that it will exclude all `test` folders within `lib/` or `node_modules/@my-org/`.

README.md
```diff
@@ -304,6 +301,35 @@ It will also exclude any files with the extension `.spec.js`.
...
+All other files in `node_modules` are filtered out by the include rules.
+The exclude rules prevent nyc instrumenting test files and `node_modules/@my-org/something/unwanted.js`
```
coreyfarrell: I think just delete this line as it repeats what we say above?

README.md
AndrewFinlay: And for some reason I thought the big red squiggly line was about spelling...

README.md
AndrewFinlay: Yeah, I don't think I explained that very well, I think I was trying to highlight differences in behaviour and ignore the obvious stuff.  But that really isn't good enough for documentation.

README.md
```diff
@@ -286,6 +302,35 @@ It will also exclude any files with the extension `.spec.js`.
...
+    "exclude": [
+      "node_modules/@my-org/something/unwanted.js",
+      "**/test/**"
+    ],
+    "excludeNodeModules": false
+  }
+}
+```
+
+**Note:** You can also include `node_modules` files with a negated exclude rule `!**/node_modules/@my-org`.
```
coreyfarrell: Instead of `You can also include`, maybe `An alternative method to include node_modules files is with ...`?

coreyfarrell: Actually on further thought maybe don't even mention the negated exclude for node_modules?  I feel like now that we have `excludeNodeModules` boolean this is just the wrong way to accomplish the goal.

AndrewFinlay: True, and anyone already using can continue to without approval from the docs.  I'll get rid of it.

PR #1052, opened 2019-04-04
docs: move setup docs out of the readme

JaKXz:


README.md
```diff
@@ -421,76 +421,9 @@ rather than having to ignore every instance of that method:
 }
 ```
 
-## Integrating with coveralls
+## [:arrow_forward: Integrating with coveralls](./docs/setup-coveralls.md)
```
coreyfarrell: Should we add the `:arrow_forward: `?  This will break existing links to the automatic anchor https://github.com/istanbuljs/nyc#integrating-with-coveralls.

I definitely like the idea of making README.md smaller.  One question though, should this information be moved to http://istanbul.js.org/?

PR #1053, opened 2019-04-04
feat: Add support for --exclude-node-modules to subcommands.

coreyfarrell:
This adds support for the `--exclude-node-modules` option/setting to
instrument, check-coverage and report sub-commands.

Add testing to verify that `--exclude-node-modules=false` is honored for
all commands.


lib/commands/instrument.js
```diff
@@ -111,6 +117,7 @@ exports.handler = function (argv) {
     preserveComments: argv.preserveComments,
     include: argv.include,
     exclude: argv.exclude,
+    excludeNodeModules: argv.excludeNodeModules,
```
JaKXz: can we just spread the argv object here instead of having to add each option manually?

```js
const nyc = new NYC({ ...argv })
```

coreyfarrell: Actually the main command and all other sub-commands just pass argv directly to nyc.  I'm going to review the call path to make sure this doesn't unexpectedly introduce any options that were previously unused but now get used.

JaKXz: I only suggested the spread in case we needed to override certain options for whatever reason, but yeah that sounds good!

coreyfarrell: Looks like I'm going to be doing a follow-up to add a bunch of flags to the `nyc instrument --help` listing.  Nothing harmful but this will pass a few options to `nyc instrument` if they're set via configuration (cache, cache-dir are two examples).

I think I'm going to do a followup which will create a `lib/commands/shared-flags.js`, this will export functions that add groups of flags - for example one will add the test-exclude options, one will add the istanbul-lib-instrument options, etc.

coreyfarrell: Additional note: we can't use object spread yet, not until we drop node.js 8.

lib/commands/instrument.js
```diff
@@ -100,20 +106,7 @@ exports.handler = function (argv) {
     ? './lib/instrumenters/istanbul'
     : './lib/instrumenters/noop'
 
-  const nyc = new NYC({
```
bcoe: It's a little weird that this refactor happens in this pull, seems like an improvement to picking and choosing parameters though.

coreyfarrell: @JaKXz saw that I was manually adding yet another option to this constructor, asked if we could do this.

coreyfarrell:
> I know this has been explained to me before, but refresh me on why this couldn't be handled by passing in alternate exclude rules that don't include node_modules.

exclude rules are handled by `test-exclude` which adds `**/node_modules/**` to the exclude list if it's not there already, unless you add the `excludeNodeModules: false` option.

Issue #892, opened 2018-07-17
Dose nyc support yaml format configuration file?


poppinlp:
I found configuration file named `.nycrc` in your README which is in JSON format. So dose this support yaml configuration file?

coreyfarrell:
Only JSON is supported.

poppinlp:
OK. Thanks. Do you think you need a PR for that? :)

coreyfarrell:
I'd be open to this if it can be done without bloating NYC.  Specifically I do not want `npm install nyc` to actually include modules needed to process YAML (or other format types).  Instead I'd be looking for NYC to report an error if `.nycrc.yaml` is found but the module needed to load it is not installed.  So `npm install --save-dev nyc js-yaml` would be needed.

I think we'll need something similar to `interpret` (but not exactly the same).  `interpret` is designed to hook the node.js `require()` funciton, we cannot do that.  So NYC cannot use `require-yaml`, it would need to use a module that simply loads YAML without touching the node.js module loader (js-yaml I think?).

poppinlp:
OK. I agree with you.
I'll think about what you said. And yep, i think `js-yaml` is good.

JaKXz:
@poppinlp are you still interested in writing that PR?

PR #1054, opened 2019-04-04
Add support for yaml configuration file

c-bandy:
I added support for `.yml` and `.yaml` config files as per request in #892.

I wasn't sure how to implement tests for this, so suggestions are welcome.


JaKXz:
@c-bandy thank you for the PR! would you mind adding to the README where it mentions the different kinds of rc files you can have?

lib/config-util.js
```diff
@@ -18,18 +18,35 @@ function guessCWD (cwd) {
...
   const rcPath = findUp.sync(rcOptions, { cwd: guessCWD(cwd) })
   let config = {}
 
   if (rcPath) {
-    if (rcPath.toLowerCase().endsWith('.js')) {
+    const rcPathLower = rcPath.toLowerCase()
+    if (rcPathLower.endsWith('.js')) {
       config = require(rcPath)
+    } else if (/.+\.y(a?)ml$/.test(rcPathLower)) {
+      if (moduleExists('js-yaml')) {
```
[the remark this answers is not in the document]
isaacs: It might be worth check for `yaml` as well?  It's pretty nice.  The only difference is that you'd use `require('yaml').parse()` instead of `require('js-yaml').load()`.

package.json
```diff
@@ -107,6 +108,7 @@
     "strip-indent": "^2.0.0",
     "tap": "^12.6.1",
     "which": "^1.3.1",
+    "yaml": "^1.4.0",
```
coreyfarrell: No point installing `yaml` dev dependency, it'll never be tested (js-yaml is found first).

lib/config-util.js
```diff
@@ -18,18 +18,45 @@ function guessCWD (cwd) {
   return cwd
 }
 
+function moduleExists (name) {
+  try {
+    return typeof require.resolve(name) !== 'undefined'
+  } catch (e) { return false }
```
coreyfarrell: I missed this before, please add line-feeds around the `return false` and tell nyc to ignore the unreachable code:
```js
try {
  return typeof require.resolve(name) !== 'undefined';
} catch (e) {
  /* istanbul ignore next */
  return false;
}
```

lib/config-util.js
```diff
@@ -18,18 +18,45 @@ function guessCWD (cwd) {
...
       config = require(rcPath)
+    } else if (/.+\.y(a?)ml$/.test(rcPathLower)) {
+      const installedYamlParser = ['js-yaml', 'yaml'].find((name) => moduleExists(name))
+      switch (installedYamlParser) {
+        case 'js-yaml':
+          config = require('js-yaml').load(
+            fs.readFileSync(rcPath, 'utf8')
+          )
+          break
+        case 'yaml':
```
coreyfarrell: Need to tell istanbul to ignore the `yaml` case as well since testing will always find js-yaml.

coreyfarrell:
@c-bandy Also please mention these new configuration files in our README.md.  See https://github.com/istanbuljs/nyc#configuring-nyc

c-bandy:
@coreyfarrell Added! 👍

lib/config-util.js
```diff
@@ -18,18 +18,49 @@ function guessCWD (cwd) {
...
   const rcPath = findUp.sync(rcOptions, { cwd: guessCWD(cwd) })
   let config = {}
 
   if (rcPath) {
-    if (rcPath.toLowerCase().endsWith('.js')) {
+    const rcPathLower = rcPath.toLowerCase()
+    if (rcPathLower.endsWith('.js')) {
       config = require(rcPath)
+    } else if (/.+\.y(a?)ml$/.test(rcPathLower)) {
+      const installedYamlParser = ['js-yaml', 'yaml'].find((name) => moduleExists(name))
```
bcoe: I'd honestly rather converge on one `yaml` parser, seems like overkill to support multiple flavors.

what's the reason to not just support `js-yaml`, it seems to be the more popular.

bcoe:
@c-bandy really appreciate this contribution 😄 the tests certainly seem like they're on the right track.

bcoe:
@c-bandy I missed that both `yaml` parsers are optional (hadn't noticed js-yaml was a developer dependency); I'm more open to checking for both 👍still might be nice to encapsulate the logic that checks for yaml parsers into a `lib/yaml.js` parser ... would be neat to potentially consider adding a similar `toml.js` parser as a follow on.

lib/config-util.js
JaKXz: @isaacs mentioned supporting both in a previous comment

lib/config-util.js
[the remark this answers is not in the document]
JaKXz: cc @bcoe

c-bandy:
@bcoe Thanks for your feedback! I will look into making these changes today

coreyfarrell:
@c-bandy please hold off for a bit, we've been having a discussion on our slack about the best way to handle yaml and other formats.  I don't think we've come to a consensus yet.  At this point we are focused on getting nyc@14.0.0 released.  YAML config will be a non-breaking change so we can consider it for 14.1.0.

isaacs:
Copying here my suggestion from the discussion elsewhere: since so many testing libraries include `js-yaml` anyway, it's probably best to just declare that one as a dep and use it.

bcoe:
@c-bandy sorry to be a pain, I know it's annoying when a PR is blocked by folks not being able to reach consensus. If you'd like to join in with the conversation, we actually have a slack here:

Would love to get your opinion as we opine.

c-bandy:
I have incorporated changed based on our discussions

PR #1055, opened 2019-04-04
feat: add processinfo index, add externalId

isaacs:
THIS IS A WIP COMMIT FOR REVIEW ONLY, DO NOT MERGE

- [ ] docs
- [x] tests

---

feat: add processinfo index, add externalId

If a NYC_PROCESSINFO_EXTERNAL_ID environment variable is set, then it is
saved in the processinfo as `externalId`.

BREAKING CHANGE: This adds a file named 'index.json' to the
.nyc_output/processinfo directory, which has a different format from the
other files in this dir.

Furthermore, when this file is generated, some additional helpful
metadata is memoized to the processinfo json files, to minimize the cost
of repeated generation.  (This isn't necessarily a breaking change, but
it is an update to the de facto schema for those files.)

As soon as possible, index generation and process tree display should be
migrated out to a new 'istanbul-lib-processinfo' library.

This opens the door to add features in the v14 release family to improve
support for partial/resumed test runs and file watching.

- When a process is run with --clean=false and a previously seen
  externalId, clear away all the coverage files in the set for that
  externalId.
- When a file is changed, a test runner can use the index to determine
  which tests (by externalId) ought to be re-run.

### DIFF
--- a/index.js
+++ b/index.js
@@ -312,6 +312,7 @@ NYC.prototype._wrapExit = function () {
 }
 
 NYC.prototype.wrap = function (bin) {
+  process.env.NYC_PROCESS_ID = this.processInfo.uuid
   this._addRequireHooks()
   this._wrapExit()
   this._loadAdditionalModules()
@@ -341,7 +342,7 @@ NYC.prototype.writeCoverageFile = function () {
     coverage = this.sourceMaps.remapCoverage(coverage)
   }
 
-  var id = this.generateUniqueID()
+  var id = this.processInfo.uuid
   var coverageFilename = path.resolve(this.tempDirectory(), id + '.json')
 
   fs.writeFileSync(
@@ -355,6 +356,7 @@ NYC.prototype.writeCoverageFile = function () {
   }
 
   this.processInfo.coverageFilename = coverageFilename
+  this.processInfo.files = Object.keys(coverage)
 
   fs.writeFileSync(
     path.resolve(this.processInfoDirectory(), id + '.json'),
@@ -421,8 +423,7 @@ NYC.prototype.writeProcessIndex = function () {
   const infos = fs.readdirSync(dir).filter(f => f !== 'index.json').map(f => {
     try {
       const info = JSON.parse(fs.readFileSync(path.resolve(dir, f), 'utf-8'))
-      // on thiis first read pass, also map the pids to uuids
-      info.uuid = path.basename(f, '.json')
+      info.children = []
       pidToUid.set(info.uuid, info.pid)
       pidToUid.set(info.pid, info.uuid)
       infoByUid.set(info.uuid, info)
@@ -436,34 +437,17 @@ NYC.prototype.writeProcessIndex = function () {
   }).filter(Boolean)
 
   // create all the parent-child links and write back the updated info
-  const needsUpdate = new Set()
   infos.forEach(info => {
-    if (info.ppid && info.ppid !== '0' && !info.parent) {
-      info.parent = pidToUid.get(info.ppid)
-      needsUpdate.add(info)
-    }
     if (info.parent) {
       const parentInfo = infoByUid.get(info.parent)
       if (parentInfo.children.indexOf(info.uuid) === -1) {
         parentInfo.children.push(info.uuid)
-        needsUpdate.add(parentInfo)
       }
     }
   })
 
   // figure out which files were touched by each process.
   const files = infos.reduce((files, info) => {
-    if (!info.files) {
-      try {
-        info.files = Object.keys(JSON.parse(fs.readFileSync(
-          path.resolve(this.tempDirectory(), info.coverageFilename),
-          'utf-8'
-        )))
-      } catch (er) {
-        return files
-      }
-      needsUpdate.add(info)
-    }
     info.files.forEach(f => {
       files[f] = files[f] || []
       files[f].push(info.uuid)
@@ -478,15 +462,16 @@ NYC.prototype.writeProcessIndex = function () {
       index.processes[info.uuid].parent = info.parent
     }
     if (info.externalId) {
+      if (index.externalIds[info.externalId]) {
+        throw new Error(`External ID ${info.externalId} used by multiple processes`)
+      }
       index.processes[info.uuid].externalId = info.externalId
       index.externalIds[info.externalId] = {
         root: info.uuid,
         children: info.children
       }
     }
-    if (info.children && info.children.length) {
-      index.processes[info.uuid].children = Array.from(info.children)
-    }
+    index.processes[info.uuid].children = Array.from(info.children || [])
     return index
   }, { processes: {}, files: files, externalIds: {} })
 
@@ -500,12 +485,6 @@ NYC.prototype.writeProcessIndex = function () {
         children.push(...nextGen.filter(uuid => children.indexOf(uuid) === -1))
       }
     }
-  })
-
-  needsUpdate.forEach(info => {
-    fs.writeFileSync(
-      path.resolve(dir, info.uuid + '.json'), JSON.stringify(info)
-    )
   })
 
   fs.writeFileSync(path.resolve(dir, 'index.json'), JSON.stringify(index))
--- a/index.js
+++ b/index.js
@@ -458,9 +458,7 @@ NYC.prototype.writeProcessIndex = function () {
   // build the actual index!
   const index = infos.reduce((index, info) => {
     index.processes[info.uuid] = {}
-    if (info.parent) {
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -413,6 +413,105 @@ NYC.prototype.report = function () {
...
+    }
+    if (info.externalId) {
+      index.processes[info.uuid].externalId = info.externalId
+      index.externalIds[info.externalId] = {
+        root: info.uuid,
+        children: info.children
+      }
+    }
+    if (info.children && info.children.length) {
+      index.processes[info.uuid].children = Array.from(info.children)
```
coreyfarrell: I think this can be done unconditionally (I'd prefer `index.processes[info.uuid].children` to be an empty array rather than omitted.

index.js
```diff
@@ -413,6 +413,105 @@ NYC.prototype.report = function () {
...
+
+  // build the actual index!
+  const index = infos.reduce((index, info) => {
+    index.processes[info.uuid] = {}
+    if (info.parent) {
+      index.processes[info.uuid].parent = info.parent
+    }
+    if (info.externalId) {
+      index.processes[info.uuid].externalId = info.externalId
+      index.externalIds[info.externalId] = {
```
coreyfarrell: What if someone accidentally uses the same externalId twice?  Not suggesting that needs to be supported, IMO it's probably an error.  I wouldn't object to a simple error:
```js
throw new Error(`External ID ${info.externalId} used by multiple processes`)
```

index.js
```diff
@@ -413,6 +413,105 @@ NYC.prototype.report = function () {
...
+  // create all the parent-child links and write back the updated info
+  const needsUpdate = new Set()
+  infos.forEach(info => {
+    if (info.ppid && info.ppid !== '0' && !info.parent) {
+      info.parent = pidToUid.get(info.ppid)
+      needsUpdate.add(info)
+    }
+    if (info.parent) {
+      const parentInfo = infoByUid.get(info.parent)
+      if (parentInfo.children.indexOf(info.uuid) === -1) {
```
coreyfarrell: Do we actually want to have `info.children` or `info.parent` in the per-process files?  This info will be written to `index.json`, I think we could completely avoid rewriting the per-process files.

lib/process.js
```diff
@@ -13,6 +13,7 @@ function ProcessInfo (defaults) {
   this.root = null
   this.coverageFilename = null
   this.nodes = [] // list of children, filled by buildProcessTree()
+  this.children = [] // just uuids, not full nodes
```
coreyfarrell: Assuming we go in the direction of not rewriting the process files I'd rather not add this here so `children` never gets added to the `processinfo/[uuid].json` files.

index.js
isaacs: Isn't that sort of the point, though?  Fairly soon, I'd like to be able to re-run a test with a matching externalId, and have nyc know to clean up _just_ the coverage and processinfo associated with that test.  I suppose while that is still pending, it's going to produce a weird result, but also, we're not actually using the index for anything until then either.

coreyfarrell: Hmmm my concern is for situations where you have 2 different tests using the same externalId.  How can we catch that?

index.js
isaacs: The current process tree generation is a bit brittle because it relies exclusively on pids and ppids, which are re-used.  Storing the uuid _somewhere_ felt important, but yeah, it is in the index.json already, so probably not necessary per-process, especially if it's the only reason to rewrite the files.

coreyfarrell: This is a actually good point, especially if you're dealing with long running `watcher` process.  What if the `lib/process.js` constructor initialized `info.uuid`, then have `NYC.writeCoverageFile` just use that value instead of generating a new uuid?  Maybe even support something like `NYC_PROCESSINFO_PARENT_UUID` for `lib/process.js` to initialize `info.parent`?

Somewhat thinking out loud here, not sure how reasonable this is.

index.js
isaacs: I think, in a subsequent update, we "catch" that by seeing that an externalId is already already in the index at the start of the process, and clearing away its coverage and processinfo entries, so that when the index is regenerated, it's no longer present.

I see what you're saying, though, that you might have something like this:

```js
spawn(node, 'foo.js', { env: { NYC_PROCESSINFO_EXTERNAL_ID: 'foo' } })
spawn(node, 'bar.js', { env: { NYC_PROCESSINFO_EXTERNAL_ID: 'foo' } })
```

right next to each other.

In that case, yes, it's user-error, and now that I think about it, when we do automatically clear out the external id'd files on subsequent runs, it'll never happen _except_ when there's a race in the same process group, if we do it right, which would make a throw here appropriate.

The `--watch` pseudocode then looks like:

- run `EID=test/foo.js node test/foo.js`
- run `EID=test/bar.js node test/bar.js`
- Generate index (test/foo.js covers lib/foo.js, test/bar.js covers lib/bar.js)
- detect change in `lib/foo.js`, consult index.json for which procs to re-run
- run `EID=test/foo.js node test/foo.js`
  - nyc finds `test/foo.js` externalId in index.json, remove all coverage and processinfo data (NB: index.json now out of sync)
  - nyc runs `node test/foo.js`, generates coverage and processinfo files
- Generate index (still only one `EID=test/foo.js` present)

Care would have to be taken that if there's _another_ change to lib/foo.js while the test is running, we don't run the next run until the first is done.  (Or kill it or something.)  But a lot of tests will need that anyway, because of fixtures and the like, so it's not a _new_ concern for tap/ava/whoever.

Ok, yes, throw here.  Will do :)

index.js
isaacs: Aha!  I was wrong, the current process prefixes the pids with the root uuid, so it's safe.  But still, using the parent uuid directly would be much clearer, imo.

lib/process.js
```diff
@@ -12,13 +15,14 @@ function ProcessInfo (defaults) {
...
-  this.nodes = [] // list of children, filled by buildProcessTree()
-
-  this._coverageMap = null
 
   for (var key in defaults) {
     this[key] = defaults[key]
   }
+
+  if (!this.uuid) {
+    this.uuid = uuid()
```
coreyfarrell: Why not just set this at the start of the function?

index.js
```diff
@@ -448,19 +524,25 @@ NYC.prototype._checkCoverage = function (summary, thresholds, file) {
...
+        path.resolve(this.processInfoDirectory(), f),
         'utf-8'
-      )))
+      ))
     } catch (e) { // handle corrupt JSON output.
-      return {}
+      return null
     }
-  })
+    if (f !== 'index.json') {
```
coreyfarrell: I'm confused, don't we just want to exclude `index.json` from this function?

index.js
```diff
@@ -412,6 +414,80 @@ NYC.prototype.report = function () {
...
+      if (index.externalIds[info.externalId]) {
+        throw new Error(`External ID ${info.externalId} used by multiple processes`)
+      }
+      index.processes[info.uuid].externalId = info.externalId
+      index.externalIds[info.externalId] = {
+        root: info.uuid,
+        children: info.children
+      }
+    }
+    index.processes[info.uuid].children = Array.from(info.children || [])
```
coreyfarrell: Nit: The `|| []` is unreachable (`info.children` will always be an array).

lib/process.js
isaacs: When we're instantiating a ProcessInfo from a file, it'll have a uuid in the `defaults` object.  Doing it here prevents calling `uuid()` unnecessarily.

index.js
isaacs: I mean, sure, we could.  I just figured we could pass one big set of `{[filename]: $data}` to the tree generation bit.  But it could also just be 2 args just as easily, I guess.

lib/process.js
coreyfarrell: Ahh yes `uuid()` can be expensive.. Good call then.

isaacs:
I think keeping it out of the docs is sensible for now.  I've been taking notes in my early sketch of istanbul-lib-processinfo anyway, but I'll probably put that off until tap v13 is finished, since I'll be able to test out some of the concepts ad hoc and see what works.  There may be some additional tweaks necessary, but I don't think the data structure needs to change from what it is right now, so we can avoid any semver major changes in the near future.

Would you like to squash, or shall I do the honors?

coreyfarrell:
> Would you like to squash, or shall I do the honors?

Lets give it a couple days to see if @addaleax has any comments since she originally implemented the processinfo functionality.  If we don't get any objections I'll squash and merge.  Probably do a second pre-release of nyc 14 by the middle of the week.

Issue #678, opened 2017-09-20
Provide --complete-copy Option for Instrumentation as in Istanbul


juancarlosfarah:
I used to run my instrumentation step with `istanbul` using something like this:

```
cross-env NODE_ENV=test istanbul instrument -i ./src/main/webapp/src -o ./src/main/webapp/instrumented --complete-copy
```

It basically instrumented all of the `.js` files, but also copied everything else, so that I could then build that folder and package it with my backend code (in this case a Java backend).

I'm trying to find a way to do it with `nyc`. When I run:

 ```
cross-env NODE_ENV=test nyc instrument ./src/main/webapp/src ./src/main/webapp/instrumented
```

That takes care of the files needing instrumentation, but I cannot get the `--complete-copy` behaviour present in `istanbul`.

I am currently doing the build with Webpack and `babel-plugin-istanbul`, which takes care of the problem for now, but it would be great if there was an equivalent using `nyc instrument`.

bcoe:
@juancarlosfarah this sounds like a great addition to functionality.

For anyone who happens to come across this issue, this seems like a great first contribution.

PR #1056, opened 2019-04-04
feat: instrument `--complete-copy` implementation

AndrewFinlay:
So this was split out from the `nyc instrument --include --exclude` work.

When running `nyc instrument <input> <output>` this will copy everything across from input to output, excluding the `.git` and `<output>` folders.  

There are a few caveats though.  
 * This implementation will dereference symlinks as it copies them, and in some cases this could be a problem i.e. `node_modules/.bin`.  
 * This also will not copy across empty directories, but this wouldn't be too hard to remedy.

If I can't find a good fix for the symlinks problem, my gut feeling on this one is that this should probably be an opt in with a `--full` switch.


coreyfarrell:
> If I can't find a good fix for the symlinks problem, my gut feeling on this one is that this should probably be an opt in with a --full switch.

I think `--complete-copy` would be good / impossible to confuse flag, this has actually been requested at #678.

index.js
```diff
@@ -195,7 +197,15 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
     const stats = fs.lstatSync(input)
     if (stats.isDirectory()) {
       inputDir = input
-      this.walkAllFiles(input, visitor)
+
+      const filesToInstrument = this.exclude.globSync(input)
+
+      if (this.config.completeCopy && output) {
+        const globOptions = { dot: true, nodir: true, ignore: ['**/.git', '**/.git/**', path.join(output, '**')] }
+        glob.sync(`${path.resolve(input)}/**`, globOptions)
```
coreyfarrell: Please use `path.resolve(input, '**')`.

index.js
```diff
@@ -194,7 +196,15 @@ NYC.prototype.instrumentAllFiles = function (input, output, cb) {
     const stats = fs.lstatSync(input)
     if (stats.isDirectory()) {
       inputDir = input
-      this.walkAllFiles(input, visitor)
+
+      const filesToInstrument = this.exclude.globSync(input)
```
bcoe: `instrumentAllFiles` is getting to be a pretty large method. I'd be tempted to potentially move this logic into `lib/instrument` and split it into a couple helper methods.

PR #1057, opened 2019-04-05
fix: set processinfo pid/ppid to actual numbers

isaacs:
As of Node.js v6, process.ppid is available, and a number.  So, there's
no need to pass the parent PID through the environment, which casts it
to a string.


coreyfarrell:
Note for when this merges: We should tag the squashed commit as a `BREAKING CHANGE` since this changes the data type of the pid/ppid fields in processinfo files.

PR #1061, opened 2019-04-07
feat: always build the processinfo temp dir

isaacs:
Always build the processinfo data directory (with index) in
${tmpdir}/processinfo.  This more easily supports cases where the
processinfo index may be useful for later analysis, or generating the
process tree post hoc.

BREAKING CHANGE: this removes the --build-process-tree option, which
appeared in previous 14.0.0 prerelease versions.

### DIFF
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -65,9 +65,7 @@ if ([
   ), function (done) {
     var mainChildExitCode = process.exitCode
 
-    if (argv.showProcessTree || argv.buildProcessTree) {
-      nyc.writeProcessIndex()
-    }
+    nyc.writeProcessIndex()
 
     if (argv.checkCoverage) {
       nyc.checkCoverage({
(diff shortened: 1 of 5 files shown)


coreyfarrell:
All of my comments below the line assume nobody has objections to always building the processinfo files.  I'm OK with it but want to give others a chance to speak.  @bcoe @JaKXz @addaleax?

---

This will not be labeled as a breaking change.  I plan to delete "add `--build-process-tree` feature" from the new features section of the changelog to prevent confusion.  14.0.0-alpha.0 was a non-tagged release so when we eventually do the final 14.0.0 release the changelog will be generated as if no pre-release versions were created (all changes between 13.3.0 and 14.0.0 will be listed as new to 14.0.0).

The cleanest way to handle this would be to revert #1028 in a separate commit to master and merge those changes into this PR.  Then when `standard-version` generates the changelog it would see the revert and ignore #1028.

Further detail: we don't have strictness enabled in yargs, so `nyc --build-process-tree` will continue to work without error and the process tree will be build.

isaacs:
Very happy to be the one to help obliterate my previous contribution ;)

Issue #839, opened 2018-05-11
Support/feature: Streaming instrumentation


ORESoftware:
With istanbul, we could avoid using the fs, and just stream instrumented code, like so:

```js
app.use(function(req,res,next){
    const k = cp.spawn('istanbul instrument foo.js');
    k.stdout.pipe(res);
});
```

we also could do it this way:

```js
import istanbul = require('istanbul');
const instrumenter = new istanbul.Instrumenter();

app.use(function(req,res,next){
   const filename = path.resolve(req.foo);
   fs.readFile(filename, function(err, data){
     instrumenter.instrument(data, filename, function (err, generatedCode) {
        res.write(generatedCode);
        res.end();
      });
  });
});
```

are either of these things possible with NYC?

JaKXz:
Hi @ORESoftware sorry for the late reply, and thanks for filing this issue, it looks interesting. There is work being done on the `instrument` command in #893 to match the old istanbul functionality. I'm not sure if it actually prints to stdout like your first example... I could be wrong though.

You could also look at the istanbuljs monorepo (the way jest does) and use those libraries yourself to do what you're trying to do.

PR #1063, opened 2019-04-10
docs: instrument docs update

AndrewFinlay:
Add documentation for `--complete-copy`, and describe how to instrument source JIT on an express server.

This should go some way to addressing #839, which looks like it's already implemented but not documented.

### DIFF
--- a/docs/instrument.md
+++ b/docs/instrument.md
@@ -1,0 +1,49 @@
+
+## Producing instrumented source
+
+The `nyc instrument` command can produce instrumented source files.
+These files are suitable for client side deployment during end to end testing.
+You can either pre-instrument your source, or write instrumented source to a stream.
+
+Run `nyc instrument --help` to display a full list of available command options.
+ 
+### Pre-instrumented source 
+ 
+You can create pre-instrumented source code by running:
+
+```bash
+nyc instrument <input> [output]
+```
+
+`<input>` can be any file or directory within the project root directory.
+The `[output]` directory is optional and can be located anywhere, if not set the instrumented code will be sent to `stdout`.
+For example, `nyc instrument . ./output` will produce instrumented versions of any source files it finds in `.` and store them in `./output`.
+
+The `--delete` option will remove existing output.
+
+The `--complete-copy` option will copy all files from the `input` directory to the `output` directory.
+When using `--complete-copy` nyc will not copy the contents of a `.git` folder to the output directory.
+
+
+**Note:** `--complete-copy` will dereference any symlinks during the copy process, this may stop scripts running properly from the output directory.
+
+### Instrument source on request
+
+You can provide instrumented source as a stream with:
+ 
+```bash
+nyc instrument <input>
+```
+
+This form of the command will stream instrumented source directly to `stdout`, that can then be piped to another process.
+You can use this behaviour to create a server that can instrument files on request.
+The following example shows streaming instrumentation middleware capable of instrumenting files on request.
+
+```javascript
+app.use((req, res, next) => {
+  const myOptions = ""
+  const filename = myHelper.getFilename(req)
+  const nyc = cp.spawn(`nyc instrument ${myOptions} ${filename}`)
+  nyc.stdout.pipe(res)
+})
+```
--- a/docs/instrument.md
+++ b/docs/instrument.md
@@ -1,5 +1,4 @@
-
-## Producing instrumented source
+# Producing instrumented source
 
 The `nyc instrument` command can produce instrumented source files.
 These files are suitable for client side deployment during end to end testing.
@@ -7,7 +6,7 @@ You can either pre-instrument your source, or write instrumented source to a str
 
 Run `nyc instrument --help` to display a full list of available command options.
  
-### Pre-instrumented source 
+## Pre-instrumented source 
  
 You can create pre-instrumented source code by running:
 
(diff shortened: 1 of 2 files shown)


AndrewFinlay:
@JaKXz that can be done, where do you want the link to go?

JaKXz:
In the readme for now as a heading link, like the coveralls and codecov links

docs/instrument.md
```diff
@@ -0,0 +1,49 @@
+
+## Producing instrumented source
```
JaKXz: We can up all the headers in this file now that it's alone :) 

There's also an extra newline here

docs/instrument.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+**Note:** `--complete-copy` will dereference any symlinks during the copy process, this may stop scripts running properly from the output directory.
+
+### Instrument source on request
+
+You can provide instrumented source as a stream with:
+ 
+```bash
+nyc instrument <input>
```
JaKXz: I don't think this intro is necessary with the example given.

docs/instrument.md
```diff
@@ -0,0 +1,49 @@
...
+
+The `--delete` option will remove existing output.
+
+The `--complete-copy` option will copy all files from the `input` directory to the `output` directory.
+When using `--complete-copy` nyc will not copy the contents of a `.git` folder to the output directory.
+
+
+**Note:** `--complete-copy` will dereference any symlinks during the copy process, this may stop scripts running properly from the output directory.
+
+### Instrument source on request
```
JaKXz: Should this be "Providing instrumented source" or something like that?

docs/instrument.md
```diff
@@ -0,0 +1,49 @@
...
+### Pre-instrumented source 
+ 
+You can create pre-instrumented source code by running:
+
+```bash
+nyc instrument <input> [output]
+```
+
+`<input>` can be any file or directory within the project root directory.
+The `[output]` directory is optional and can be located anywhere, if not set the instrumented code will be sent to `stdout`.
```
JaKXz: maybe specify that these should be relative paths?

docs/instrument.md
```diff
@@ -0,0 +1,49 @@
...
+For example, `nyc instrument . ./output` will produce instrumented versions of any source files it finds in `.` and store them in `./output`.
+
+The `--delete` option will remove existing output.
```
JaKXz: from where?

docs/instrument.md
```diff
@@ -0,0 +1,49 @@
...
+
+The `--complete-copy` option will copy all files from the `input` directory to the `output` directory.
```
JaKXz: Will it override the instrumented output? I think more specificity would be good here.

AndrewFinlay: That seems obvious to me, I don't see why you'd run `nyc instrument` to simply copy files and not instrument them.  But it can change if it needs to.

docs/instrument.md
AndrewFinlay: Cool, will make this more specific

docs/instrument.md
AndrewFinlay: Yeah, I had trouble coming up with a good title for this.  Originally I was thinking `Just in time instrumentation` or `Streaming instrumentation`

docs/instrument.md
AndrewFinlay: After running a few tests, it works with both relative and absolute paths, on OSX at least.

docs/instrument.md
JaKXz: That's what I mean, I don't really understand from the wording... I assume the output will be `input.js` and `input-instrumented.js` but I don't know if that's actually right

docs/instrument.md
JaKXz: True... Curious how it would behave on windows

docs/instrument.md
```diff
@@ -0,0 +1,41 @@
...
+The `--delete` option will remove the existing output directory before instrumenting.
+
+The `--complete-copy` option will copy all remaining files from the `input` directory to the `output` directory.
+When using `--complete-copy` nyc will not copy the contents of a `.git` folder to the output directory.
+
+**Note:** `--complete-copy` will dereference any symlinks during the copy process, this may stop scripts running properly from the output directory.
+
+## Streaming instrumentation
+
+This form of the command will stream instrumented source directly to `stdout`, that can then be piped to another process.
```
JaKXz: This sentence should just start with `nyc instrument` will stream...

coreyfarrell: @AndrewFinlay do you have a moment to address this so we can merge?  If not I can make the change.

bcoe:
@JaKXz @AndrewFinlay, just bumping this up; what do we need to do to get this over the finish line?

JaKXz:
@bcoe done!

docs/instrument.md
AndrewFinlay: I've been working on @JaKXz 's queries about how instrument handles absolute and relative paths in Windows, but life, family and Easter come first.  I think this is the only remaining piece of work for this PR, and my gut feeling says that if Windows can't accept both relative and absolute paths that may be a bug anyway.  So I'll get my Windows test env up and if I find anything that needs to change I'll pop it in a new PR,

Oh and thanks to all of you @JaKXz, @coreyfarrell & @bcoe for your help in getting these instrument changes over the line.  It's been fun.

PR #1070, opened 2019-04-17
fix: Exit with code 1 when nyc doesn't know what to do.

coreyfarrell:
When nyc is not told to do anything this is an error.  We already
printed the help message to stderr, now we exit with code 1 to indicate
that nyc exited without doing anything.

Add a test to cover this edge case and verify that both `nyc` and `nyc
--help` produce output as expected to stdout and stderr, exit with the
proper code.

### DIFF
--- a/test/nyc-integration.js
+++ b/test/nyc-integration.js
@@ -12,7 +12,7 @@ const glob = require('glob')
 const isWindows = require('is-windows')()
 const rimraf = require('rimraf')
 const makeDir = require('make-dir')
-const spawn = require('child_process').spawn
+const { spawn, spawnSync } = require('child_process')
 const si = require('strip-indent')
 
 require('chai').should()
@@ -1077,6 +1077,23 @@ describe('the nyc cli', function () {
         done()
       })
     })
+  })
+
+  it('help shows to stderr when main command doesn\'t know what to do', () => {
+    const opts = {
+      cwd: fixturesCLI,
+      env,
+      encoding: 'utf8'
+    }
+
+    const help = spawnSync(process.execPath, [bin, '--help'], opts)
+    const unknown = spawnSync(process.execPath, [bin], opts)
+    help.status.should.equal(0)
+    unknown.status.should.equal(1)
+    help.stderr.should.equal('')
+    unknown.stdout.should.equal('')
+    help.stdout.should.not.equal('')
+    help.stdout.should.equal(unknown.stderr)
   })
 
   describe('args', function () {
--- a/bin/nyc.js
+++ b/bin/nyc.js
@@ -85,5 +85,6 @@ if ([
   })
 } else {
   // I don't have a clue what you're doing.
+  process.exitCode = 1
   yargs.showHelp()
 }


test/nyc-integration.js
```diff
@@ -1079,6 +1079,40 @@ describe('the nyc cli', function () {
...
+      }
+      proc.stdout.on('data', chunk => {
+        msgs[type].out += chunk
+      })
+      proc.stderr.on('data', chunk => {
+        msgs[type].err += chunk
+      })
+      proc.on('close', code => {
+        code.should.equal(expectCode)
+        waitingFor--
```
JaKXz: While I think this is clever, I feel like it could be done simpler. I've been wanting to introduce https://npm.im/await-spawn for a while, what do you think of it here?

coreyfarrell: I think that would be cool once we can use async/await (nyc@15).  As long as we're still supporting node.js 6 and not transpiling test/*.js I'd rather leave it as is.

JaKXz: It returns a promise so we don't have to use the new syntax just yet. Just:

```js
return spawn(...)
  .then(() => {
    //asserts
  })
```

coreyfarrell: I'm OK with using await-spawn for these new tests but I don't think we should do a mass change to `return spawn(...).then(...)`.  I'd much rather we just leave existing tests alone until we can use async/await sytax.

coreyfarrell: Actually it looks like `await-spawn` won't let us see both stdout and stderr.  I think our testing would be better served by something which mimics the API of `spawnSync`, just with promises.  So `spawn(...).then(({stdout, stderr, status}) => { ... })`.  This is something which is currently lacking IMO, we should be checking for empty stderr in most tests (including when the exit code is 0).

coreyfarrell: Any feelings about using [execa](https://github.com/sindresorhus/execa#readme) in tests?  It's actually already installed via yargs so although it's a bigger module it wouldn't even increase our node_modules size.

JaKXz: Sure! execa looks like a better fit, I agree :)

coreyfarrell: Very nice.

test/nyc-integration.js
```diff
@@ -14,6 +14,8 @@ const rimraf = require('rimraf')
 const makeDir = require('make-dir')
 const spawn = require('child_process').spawn
 const si = require('strip-indent')
+// BUGBUG: when upgrading execa to 2.x stripEof is renamed to stripFinalNewline.
```
JaKXz: should we not upgrade because of this? what's wrong?

coreyfarrell: 2.x isn't released yet and will require node.js 8.

test/nyc-integration.js
```diff
@@ -1079,6 +1081,25 @@ describe('the nyc cli', function () {
...
   })
 
+  it('help shows to stderr when main command doesn\'t know what to do', () => {
+    const opts = {
+      cwd: fixturesCLI,
+      env,
+      stripEof: false
+    }
+    return Promise.all([
+      execa(process.execPath, [bin, '--help'], opts),
```
bcoe: any reason we can't just use `spawnSync` here, I've been trying to use less dependencies when possible.

coreyfarrell: execa is already a dependency (it's installed by yargs).  I'd like to leave the door open for parallel testing which can never happen if we use sync methods everywhere.

test/nyc-integration.js
coreyfarrell: @bcoe has requested that we not use execa, for now I'm just going to use spawnSync.

bcoe: :+1: love to use Node's built-ins where possible, and it felt a bit weird that we were adding a caveat:

```
// BUGBUG: 
```

unrelated to our test; thanks for entertaining me being curmudgeonly.

test/nyc-integration.js
JaKXz: ^ agree with @coreyfarrell here

PR #1071, opened 2019-04-17
chore: Add test for `nyc --no-clean`.

coreyfarrell:
Verify that running `nyc command1 && nyc --no-clean command2` causes
results results to be merged.


JaKXz:
> using snapshots for these types of tests
@coreyfarrell and I were talking about this on slack... I will try to dedicate some time to this again. I have a few branches of WIP but I keep ending up expanding the scope because it requires quite a few changes to properly use tap's test api...

PR #1074, opened 2019-04-18
docs(codecov): favour npx over installing locally

JaKXz:


docs/setup-codecov.md
```diff
@@ -1,26 +1,26 @@
 # Integrating with codecov.io
 
-> **tl;dr**: 
-> `nyc --reporter=lcov npm test && npx codecov`
+**tl;dr**: 
+### `npx nyc --reporter=lcov npm test && npx codecov`
```
coreyfarrell: This may be the reason some people are running nyc inside nyc.  If `npm test` runs tests in nyc then it might be better for people to do `npm test && npx nyc report --reporter lcov && npx codecov`.  Not sure a great way to document this, just wanted to mention it and maybe see what you think.

JaKXz: I see what you mean. I'll update to reflect that. Thanks!

PR #1075, opened 2019-04-18
chore: Convert some tap tests to run parallel and use snapshots.

coreyfarrell:


package.json
```diff
@@ -96,9 +96,11 @@
     "any-path": "^1.3.0",
     "chai": "^4.2.0",
     "coveralls": "^3.0.3",
+    "cross-env": "^5.2.0",
     "is-windows": "^1.0.2",
     "lodash": "^4.17.11",
     "newline-regex": "^0.2.1",
+    "pify": "^4.0.1",
```
JaKXz: @bcoe we're ok with adding this dependency? as @coreyfarrell said `execa` is already brought in and would be great to use here.

coreyfarrell: This is temporary, can be replaced with `util.promisify` once we drop node.js 6.

package.json
```diff
@@ -4,12 +4,14 @@
   "description": "the Istanbul command line interface",
   "main": "index.js",
   "scripts": {
-    "pretest": "npm run clean && npm run instrument",
-    "test": "tap -t360 --no-cov -b ./test/*.js && npm run report",
-    "posttest": "standard",
+    "lint": "standard",
+    "pretest": "npm run lint && npm run clean && npm run instrument",
+    "test": "tap -t360 --no-cov -b test/*.js",
+    "resnap": "cross-env TAP_SNAPSHOT=1 tap -t360 --no-cov -b test/nyc-integration-snap.js",
```
isaacs: The more common convention is to call this script `"snap"`.  (In tap v13, if the script is named "snap", then it regenerates snapshots by default.)

Also, why just that one file?  Snapshots are useful whenever output is being tested, may as well just run the whole suite.

coreyfarrell: I was just running the one file because I was regenerating the snap file many, many times.  Right now nyc-integration-snap.js is the only source which uses snapshots, I didn't want to go too far without feedback.

One question, in test/nyc-integration-snap.js I have:
```js
if (process.env.TAP_SNAPSHOT !== '1') {
  t.jobs = os.cpus().length
}
```

Obviously setting `t.jobs` is set to enable running multiple tests at a time.  I disabled this when snapshots are being generated to force a stable order.  Is this something tap could improve on?  Either automatically locking `t.jobs == 1` when snapshots are being generated or otherwise do something to enforce a stable order of the snapshot file?  Possibly would use a lot of memory to buffer all snapshots until they can be written in order so maybe just ignore attempts to enable parallelism?

coreyfarrell:
@JaKXz @bcoe I've made a few tweaks, mainly moving some code from the test file to helpers.  Partly this is due to the test @isaacs posted in #1077 made me realize that we might want to share more code.  One internal behavior change instead of adding stdio chunks directly to a string I'm now using an array of buffers and Buffer.concat at the end.  In theory this shouldn't have an effect but I just like the code better (saw it in #1077).

package.json
isaacs: The memory overhead is pretty minimal.  But yeah, forcing `t.jobs=1` is good when snapshotting, or else you get a lot of churn in the snapshot file.  But better yet, it should probably just sort the snapshot file before writing it.

test/helpers/temp-dir-setup.js
```diff
@@ -0,0 +1,31 @@
...
+const rimraf = pify(_rimraf)
+const mkdtemp = pify(fs.mkdtemp)
+
+function tempDirSetup (t, testFile) {
+  const { dir, name } = path.parse(testFile)
+  const tempDirBase = path.resolve(dir, 'temp-dir-' + name)
+
+  makeDir.sync(tempDirBase)
+
+  // Do not use array function for beforeEach
```
JaKXz: typo: arrow fn

test/helpers/run-nyc.js
```diff
@@ -0,0 +1,84 @@
+const path = require('path')
+const { spawn } = require('child_process')
+
```
bcoe: I might call this `helpers.js`, because it's exporting a few other helpers other than just the `runNyc` method`.

test/nyc-integration-snap.js
```diff
@@ -0,0 +1,187 @@
+const path = require('path')
+const os = require('os')
+
+const t = require('tap')
```
bcoe: I find this files naming a little weird; I've been tending to call my snapshot files something like "foo.snap", and I thought at first this file was your snapshot not a test file. I'd be tempted to just keep the tests in `nyc-integration.js`, but perhaps in a new `describe` block, like `describe('snapshot tests')` then once all the tests are moved over, get rid of that describe.

test/helpers/run-nyc.js
JaKXz: since it's in the helpers folder, perhaps just `index.js`?

test/nyc-integration-snap.js
JaKXz: @bcoe it's harder (not impossible) to use the snapshot matching with the mocha globals mode in tap, so I think this was a very reasonable approach to lower the difficulty of the move. We can definitely make more moves after this is merged

bcoe: sounds good, let's just add a TODO to this effect then.

test/helpers/run-nyc.js
coreyfarrell: I've refactored the helpers dir to have an index.js which pulls in implementation from individual files.

test/nyc-integration-snap.js
coreyfarrell: I've renamed the test files so that we have snapshot tests in `nyc-integration.js` and the tests which still need to be migrated are in `nyc-integration-old.js`.  This way we don't have the `snap` in the test filename.

PR #1077, opened 2019-04-19
 feat: Add support for env.NYC_CONFIG_OVERRIDE 

isaacs:
This allows a process to spawn a child process with an environment
variable NYC_CONFIG_OVERRIDE.  This is a JSON string which overrides any
values in the NYC_CONFIG env.  It is not deleted in the child process,
so will be contagious unless unset, and it is the responsibility of the
caller to manage.

The first intended use case of this feature is to allow node-tap to map
tests to the specific portion of a system under test that they cover.
In this way, unit tests can be more focused, and only re-run when their
specific unit has changed.

There are, of course, many other uses that this could be put to, and it
did not seem appropriate to add a special hook _just_ for overriding
the include list.


coreyfarrell:
I'm concerned about our expanding environmental variable surface.  Not specifically against this but I'd like to hear what @JaKXz and @bcoe have to say / if they can think of any alternative methods of accomplishing the goal.

isaacs:
@coreyfarrell #1078 removes some environs that are no longer load-bearing, if you're concerned about that.

coreyfarrell:
@isaacs I think removing unused environs in the next major release will be good but my concern is more about adding environmental variables which can be set from outside NYC to manipulate how it runs.

PR #1080, opened 2019-04-21
fix: Purge source-map cache before reporting if cache is disabled.

### LINKED ISSUE #1049: Typescript coverage doesn't include some lines after 13.2.0 update
## Link to bug demonstration repository.

https://github.com/damyanpetev/typescript-coverage-repro
(install `nyc@latest` to see the difference)

## Expected Behavior
Coverage report on a TypeScript project to include all relevant lines and if/else paths.
## Observed Behavior
After updating v13.1.0  to latest (specifically above v13.2.0) coverage percents differ and seem to ignore some lines (specifically noticed that with if/else blocks). Consequently, the coverage also doesn't report the if/else paths that are not taken:
image

## Forensic Information

**Operating System:** Windows 10
**Environment Information:** tested on both node v8.15.1/v10.15.2 with npm 6.4.1.
(initial commit on demo repo has [package-lock.json](https://github.com/damyanpetev/typescript-coverage-repro/blob/e1694525e8826fa2f055562fe259b4f5de1a1d8f/package-lock.json) with packages installed)

coreyfarrell:
@bcoe this (combined with istanbuljs/istanbuljs#375) seems to fix #1049 (and likely others like it).


bcoe:
@istanbuljs/core awesome work 👍 weirdly, `npm link` isn't working well for me to test my changes in `istanbul-lib-source-map`. might need to release a `next` release, just for testing purposes.

lib/source-maps.js
```diff
@@ -4,14 +4,20 @@ const libSourceMaps = require('istanbul-lib-source-maps')
 const fs = require('fs')
 const path = require('path')
 
-const sourceMapCache = libSourceMaps.createSourceMapStore()
+let sourceMapCache = libSourceMaps.createSourceMapStore()
```
bcoe: with your refactor to stop making new instances of `Nyc`, perhaps you're right that this doesn't need to be a singleton?

lib/source-maps.js
```diff
@@ -4,14 +4,20 @@ const libSourceMaps = require('istanbul-lib-source-maps')
...
 function SourceMaps (opts) {
   this.cache = opts.cache
   this.cacheDirectory = opts.cacheDirectory
   this.loadedMaps = {}
   this._sourceMapCache = sourceMapCache
 }
 
+SourceMaps.prototype.purgeCache = function () {
+  sourceMapCache = libSourceMaps.createSourceMapStore()
```
bcoe: I don't understand why we'd need to throw out source maps like this,  might be worth digging into things a bit more, and figuring out how we end up with incompatible source-maps in the first place.

I'm not opposed to landing this as a patch  as a temporary measure, but it feels like we're continuing to complicate our source-map and caching logic, which is already a mess. I think you're right that we just need to get way more test cases in the codebase.

coreyfarrell: Same reason we need to throw out transform results (they set `cache: false`).

coreyfarrell:
> might need to release a next release, just for testing purposes

For istanbul-lib-source-maps I think definitely.  Once we do that I think we should ping Simen from jest to ensure it doesn't cause regressions for them.

index.js
```diff
@@ -236,6 +236,12 @@ class NYC {
     return this._transform(code, filename)
   }
 
+  maybePurgeSourceMapCache () {
+    if (!this.cache) {
```
JaKXz: Maybe I'm crazy but isn't this clause written backwards? My guess is that it should be `!!this.cache` but I don't know for sure, just wanted to ask to be sure.

coreyfarrell: The code is correct.  lib/source-map.js needs to cache source-maps in memory during execution of nyc.  But if caching is disabled then we need to purge the caches before running reports.

JaKXz: Ah. The naming got me.

PR #1082, opened 2019-04-22
fix: Use correct config property for parser plugins

coreyfarrell:


coreyfarrell:
I can't think of a reasonable way to have an automated test against this regression.  The importance of this option will become less important once istanbuljs/istanbuljs#379 is merged.

PR #1083, opened 2019-04-23
chore: Enable testing of node.js 12 in place of node.js 11.

coreyfarrell:


JaKXz:
I'd prefer to keep the `node` specification because that will update to node v12 (latest release) soon and we don't have to keep updating it when new versions come out.

coreyfarrell:
We'll have to update it before 13 is released no matter what, at that point `node` will become 13 and we would suddenly stop testing 12.  Unless I'm misunderstanding?

JaKXz:
Fair enough, then we should keep `node` and just add `12`.

coreyfarrell:
I'm not following.  If we keep `node` and add `12` it'll double-test node.js 12 right now, unless you're suggesting that we should leave it alone now and remember to add 12 after 13 is released?

JaKXz:
The latter - that'll be around the time we deprecate node 6 anyway

coreyfarrell:
This does not make sense to me but I'm not up for arguing it.

PR #1084, opened 2019-04-24
chore: node.js 12 compatibility for object snapshot test.

coreyfarrell:
node.js 12 changes the format of `util.inspect`, this causes snapshot
mismatch.  We only use object snapshot testing in one place so call
`util.inspect` manually with `compact: false` option to force consistent
output before / after node.js 12.

---

CC @isaacs not sure if tap@12 can/should do anything about this, kind of a tough situation as `compact: false` would be a breaking change for tap.  Either way figure you might want to know about this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "lint": "standard",
     "pretest": "npm run lint && npm run clean && npm run instrument",
     "test": "tap -t360 --no-cov -b test/*.js",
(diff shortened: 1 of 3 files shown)


coreyfarrell:
Apparently `compact: false` won't work anyways - it's not supported by node.js <=8.

PR #1088, opened 2019-04-25
feat: Support retrieving config from config.toNYCRC()

### LINKED ISSUE #1087: feat: Support `nycRC()` in configuration object
I'm toying with programmatically created configurations at [@cfware/nyc](https://github.com/cfware/cfware-nyc).  This module exports a class with methods implemented as both static factory functions and chainable non-static manipulation functions.  The idea is that you could create an nyc.config.js as follows:
```js
'use strict';

module.exports = require('@cfware/nyc')
  .fullCoverage()
  .include('lib/**')
  .exclude('lib/do-not-cover.js')
  .settings;
```

This example has two main benefits over normal declarative config:
* `.fullCoverage()` sets multiple properties (similar to `tap --100`)
* `.exclude()` does not clear defaults so `lib/do-not-cover.js` is in addition to `testExclude.defaultExcludes`.

I'd love to avoid having to explicitly use `.settings` in my `nyc.config.js` files.  This could be supported if the nyc `lib/config-util.js` did something like:
```js
if (Symbol.for('nycrc') in config) {
  config = config[Symbol.for('nycrc')]
}
```
In all supported versions of node `Symbol.for('nycrc') === Symbol.for('nycrc')` so this would allow nyc to pull it's configuration from config generator objects such as mine.  I've already verified that adding this code does not break any existing tests.

@bcoe @JaKXz @isaacs any thoughts / objections?

coreyfarrell:
Fixes #1087

### DIFF
--- a/docs/config-generator.md
+++ b/docs/config-generator.md
@@ -1,0 +1,66 @@
+# Writing a generic configuration generator class
+
+It is possible to write a class which programmatically generates your nyc configuration.
+In this case a `toNYCRC()` method should be provided which returns the actual configuration.
+
+A simple example which allows you to append the default exclude list instead of replacing
+it:
+
+```js
+'use strict';
+
+const {defaultExclude} = require('test-exclude');
+
+class ConfigGenerator {
+    constructor() {
+        this.settings = {
+            all: true,
+            exclude: [...defaultExclude]
+        };
+    }
+
+    toNYCRC() {
+        return this.settings;
+    }
+
+    fullCoverage(perFile = true) {
+        Object.assign(this.settings, {
+            checkCoverage: true,
+            checkCoverage: true,
+            perFile,
+            lines: 100,
+            statements: 100,
+            functions: 100,
+            branches: 100
+        });
+
+        return this;
+    }
+
+
+    exclude(...globs) {
+        this.settings.exclude.push(...globs);
+
+        return this;
+    }
+}
+
+module.exports = () => new ConfigGenerator();
+```
+
+# Using your configuration generator
+
+The following example assumes the above script was published as `@my-org/nyc-config`:
+
+```js
+'use strict';
+
+module.exports = require('@my-org/nyc-config')()
+    .fullCoverage(true)
+    .exclude('build/**');
+```
+
+When the object created by `@my-org/nyc-config` is given to nyc the `toNYCRC()` method
+is recognized and used to find the actual configuration.  Without `toNYCRC()` each package
+which used `@my-org/nyc-config` to generate config would have to add `.settings` to the
+end of the call chain.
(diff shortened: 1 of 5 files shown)


test/fixtures/cli/nyc-config-js/nycrc-config.js
```diff
@@ -1,9 +1,13 @@
 module.exports = {
-  'check-coverage': true,
-  'per-file': true,
-  lines: 100,
-  statements: 100,
-  functions: 100,
-  branches: 100,
-  exclude: []
+  toNYCRC() {
```
coreyfarrell: @JaKXz any suggestion on if/how this should be documented?

JaKXz: I would write something in `docs/` with a link to it at the bottom of the readme.

coreyfarrell: I added the link below the section about publishing reusable configs, this seemed like a better place than at the very end.  Let me know what you think.

docs/config-generator.md
```diff
@@ -0,0 +1,66 @@
...
+```js
+'use strict';
+
+module.exports = require('@my-org/nyc-config')()
+    .fullCoverage(true)
+    .exclude('build/**');
+```
+
+When the object created by `@my-org/nyc-config` is given to nyc the `toNYCRC()` method
+is recognized and used to find the actual configuration.  Without `toNYCRC()` each package
```
JaKXz: nit: The standard that we've been using recently and I think has been working well is using one line per sentence. It makes writing, reviews, and modifications easier.

docs/config-generator.md
```diff
@@ -0,0 +1,66 @@
...
+        return this;
+    }
+}
+
+module.exports = () => new ConfigGenerator();
+```
+
+# Using your configuration generator
+
+The following example assumes the above script was published as `@my-org/nyc-config`:
```
JaKXz: Need to specify that this is in `nyc.config.js` file.

docs/config-generator.md
```diff
@@ -0,0 +1,66 @@
...
+
+A simple example which allows you to append the default exclude list instead of replacing
+it:
+
+```js
+'use strict';
+
+const {defaultExclude} = require('test-exclude');
+
+class ConfigGenerator {
```
JaKXz: maybe we should publish this as a preset package? `@istanbuljs/base-config-generator` which people could then extend? is that overkill?

docs/config-generator.md
coreyfarrell: My editor does not word wrap, so more than 150 characters is bad for me.  Can we impose that as a hard limit (line-feed in the markdown before column 150 even if the sentence is longer)?

PR #1093, opened 2019-04-28
Readme Reshuffle

JaKXz:
The intent here is to make the readme much better for newcomers and getting started faster. The major change is promoting the config packages and moving the more advanced features and options towards the bottom. Let me know what you think!

Fixes ...many issues. Hopefully.
Resolves ...many questions. Hopefully.

Easier diff to read: https://github.com/istanbuljs/nyc/tree/82a6dd5

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ nyc uses the project root directory when:
  * creating globs for include and exclude rules during file selection
  * loading custom require hooks from the `require` array
 
-nyc may create artefact directories within the project root, such as:
+nyc may create artefact directories within the project root, with these defaults:
   * the report directory, `<project-root>/coverage`
   * the cache directory, `<project-root>/node_modules/.cache/nyc`
   * the temp directory, `<project-root>/.nyc_output`
@@ -338,26 +338,40 @@ rather than having to ignore every instance of that method:
 }
 ```
 
-## Support for custom require hooks (babel, typescript, etc.)
+## Combining reports from multiple runs
+If for whatever reason you have different test runners in your project or a different series of test runs for different kinds of tests, nyc will automatically combine the coverage report for you if configured correctly with the `--no-clean` flag and the `report` command.
+Originally inspired by @janiukjf in #1001, here's an example, where the `test:*` scripts (not shown) invoke only your test runner(s) and not nyc:
 
-nyc supports custom require hooks like [`@babel/register`]. nyc can load
-the hooks for you, [using the `--require` flag](#require-additional-modules).
+```json
+{
+  "scripts": {
+    "cover": "npm run cover:unit && npm run cover:integration && npm run cover:report",
+    "cover:unit": "nyc --silent npm run test:unit",
+    "cover:integration": "nyc --silent --no-clean npm run test:integration",
+    "cover:report": "nyc report --reporter=lcov --reporter=text"
+  }
+}
+```
 
-Source maps are used to map coverage information back to the appropriate lines
-of the pre-transpiled code. You'll have to configure your custom require hook
-to inline the source-map in the transpiled code. For Babel that means setting
-the `sourceMaps` option to `inline`.
+### What about `nyc merge`?
+
+The `nyc merge` command is for producing one _raw coverage output file_ that combines the results from many test runs.
+So if you had the above setup and needed to produce a single `coverage.json` for some external tool, you could do:
+
+```json
+{
+  "scripts": {
+    "cover:merge": "npm run cover:unit && npm run cover:integration && nyc merge .nyc_output coverage.json"
+  }
+}
+```
 
 ## Source-Map support for pre-instrumented codebases
 
-If you opt to pre-instrument your source-code (rather than using a just-in-time
-transpiler like [`@babel/register`]) nyc supports both inline source-maps and
-`.map` files.
+If you opt to pre-instrument your source-code (rather than using a just-in-time transpiler like [`@babel/register`]) nyc supports both inline source-maps and `.map` files.
 
-_Important: If you are using nyc with a project that pre-instruments its code,
-run nyc with the configuration option `--exclude-after-remap` set to `false`.
-Otherwise nyc's reports will exclude any files that source-maps remap to folders
-covered under exclude rules._
+_Important: If you are using nyc with a project that pre-instruments its code, run nyc with the configuration option `--exclude-after-remap` set to `false`.
+Otherwise nyc's reports will exclude any files that source-maps remap to folders covered under exclude rules._
 
 ## [Integrating with coveralls](./docs/setup-coveralls.md)
 
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ This table is a quick TLDR for the rest of this readme and there are more advanc
 | `exclude` | See [selecting files for coverage for more info](#selecting-files-for-coverage) | `Array<String>` | [list](https://github.com/istanbuljs/istanbuljs/blob/master/packages/test-exclude/index.js#L176-L184) |
 | `reporter` | [coverage reporters to use](https://istanbul.js.org/docs/advanced/alternative-reporters/) | `Array<String>` | `['text']` |
 | `report-dir` | where to put the coverage report files | `String` | `./coverage` |
+| `skip-full` | don't show files with 100% statement, branch, and function coverage | `Boolean` | `false` |
 | `temp-dir` | directory to output raw coverage information to | `String` | `./.nyc_output` |
 
 Configuration can also be provided by `nyc.config.js` if programmed logic is required:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -13,207 +13,111 @@ Istanbul's state of the art command line interface, with support for:
...
+This table is a quick TLDR for the rest of this readme and there are more advanced docs available.
 
-To view your coverage report:
+| Option name | Description | Type | Default |
+| ----------- | ----------- | ---- | ------- |
+| `all` | whether or not to instrument all files (not just the ones touched by your test suite) | `Boolean` | `false` |
+| `check-coverage` | check whether coverage is within thresholds, fail if not | `Boolean` | `false` |
+| `extension` | a list of extensions that nyc should attempt to handle in addition to `.js` | `Array<String>` | `['.js']` |
+| `include` | [globs of files to be included from instrumentation](#selecting-files-for-coverage) | `Array<String>` | `['**']`|
+| `exclude` | See [selecting files for coverage for more info](#selecting-files-for-coverage) | `Array<String>` | [list](https://github.com/istanbuljs/istanbuljs/blob/master/packages/test-exclude/index.js#L176-L184) |
```
coreyfarrell: I'd say move ` for more info` out of the link.

For the default link the list is OK for now but I think we will need to move the defaults list in test-exclude to a separate file.  The problem here is that any change to test-exclude code is likely to make this link point to the wrong line range.

README.md
```diff
@@ -13,207 +13,111 @@ Istanbul's state of the art command line interface, with support for:
...
+You can set these in any of the files listed above, or from the command line.
+This table is a quick TLDR for the rest of this readme and there are more advanced docs available.
 
-To view your coverage report:
+| Option name | Description | Type | Default |
+| ----------- | ----------- | ---- | ------- |
+| `all` | whether or not to instrument all files (not just the ones touched by your test suite) | `Boolean` | `false` |
+| `check-coverage` | check whether coverage is within thresholds, fail if not | `Boolean` | `false` |
+| `extension` | a list of extensions that nyc should attempt to handle in addition to `.js` | `Array<String>` | `['.js']` |
+| `include` | [globs of files to be included from instrumentation](#selecting-files-for-coverage) | `Array<String>` | `['**']`|
```
coreyfarrell: Since this is the same link `exclude` references maybe we should have the same text?  `See [selecting files for coverage] for more info`.

README.md
```diff
@@ -13,207 +13,111 @@ Istanbul's state of the art command line interface, with support for:
...
-nyc uses the project root directory when:
- * looking for source files to instrument
- * creating globs for include and exclude rules during file selection
- * loading custom require hooks from the `require` array
-
-nyc may create artefact directories within the project root, such as:
-  * the report directory, `<project-root>/coverage`
-  * the cache directory, `<project-root>/node_modules/.cache/nyc`
-  * the temp directory, `<project-root>/.nyc_output`
+To publish and reuse your own `nyc` configuration, simply create an npm module that exports an `index.json` with your `nyc` config.
```
coreyfarrell: Not sure we want to hit this level of detail here, but the npm module does not have to use `index.json`.  Modules can be written in CJS instead of JSON, https://github.com/istanbuljs/istanbuljs/blob/master/packages/nyc-config-hook-run-in-this-context/index.js for example.

README.md
```diff
@@ -454,10 +313,40 @@ rather than having to ignore every instance of that method:
 }
 ```
 
+## Accurate stack traces using source-maps
```
coreyfarrell: I realize you didn't write this section / it's just been moved but from what I can tell it does not work at all.  The problem is that by default source-map-support does not support inline source-maps (like those produced by nyc itself).  Interesting note `@babel/register` runs `source-map-support` in a way which finds the source-map's produced by babel, so AFAIK right now the only way to get accurate stack traces is to have nyc use `@babel/register` + `babel-plugin-istanbul`.

See https://github.com/evanw/node-source-map-support/issues/239

JaKXz: Should I just delete it for now? I was on the fence about it since I wasn't sure, and I think it warrants a more advanced doc later on when that issue is resolved

README.md
JaKXz: ah good catch. thanks, will fix

README.md
coreyfarrell: I'm leaning towards yes.

bcoe: I agree .. the more work I do on source-maps, the more I realize this is something we should try to patch in Node.js itself. This would at least, hopefully, mean folks using Node 12.x would eventually have better stack traces.

README.md
```diff
@@ -13,207 +13,111 @@ Istanbul's state of the art command line interface, with support for:
...
 {
   "scripts": {
-    "test": "nyc mocha"
+    "test": "npx nyc mocha"
   }
 }
 ```
 
-Alternatively, you can install nyc globally and use it to execute `npm test`:
+Or use your package manager to add a dev dependency: `npm i -D nyc` or `yarn add -D nyc`.
```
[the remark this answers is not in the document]
JaKXz: I think I prefer the npx first approach because it gets rid of the problem of people having outdated versions of nyc, and as a bonus for those who prefer to have less in their project node_modules but still shares nyc across their machine. It's basically what `npm i -g` always wanted to be imo

I'll make sure to mention not having nyc already in your test script... and to make adjustments if you have to run other executables. Though with gulp specifically I'm not sure why that wouldn't be in your dependencies, you usually have to write a gulpfile based on the node api. Still a valid point, I'll try to figure out the wording today

README.md
JaKXz: I think that will just have to change when that change happens in test-exclude

README.md
```diff
@@ -13,207 +13,111 @@ Istanbul's state of the art command line interface, with support for:
...
+nyc allows you to inherit other configurations using the key `extends`.
+As an example, an alternative way to configure nyc for `babel-plugin-istanbul` would be to use the [@istanbuljs/nyc-config-babel preset](https://www.npmjs.com/package/@istanbuljs/nyc-config-babel):
 
-```bash
-nyc report --reporter=<custom-reporter-name>
+```json
+{
+  "nyc": {
+    "extends": "@istanbuljs/nyc-config-babel"
+    //add your custom overrides here
```
bcoe: If we make this snippet `.js` rather than `.json`, it will look less wonky.

README.md
```diff
@@ -235,14 +235,17 @@ nyc may create artefact directories within the project root, such as:
 
 ## Require additional modules
 
-The `--require` flag can be provided to `nyc` to indicate that additional
-modules should be required in the subprocess collecting coverage:
+The `--require` flag can be provided to `nyc` to indicate that additional modules should be required in the subprocess collecting coverage:
```
coreyfarrell: I'm thinking the example we give should be `nyc --require esm mocha`.  I suggest this as `esm` is not involved in instrumentation, further enforce that babel / typescript projects should use `@istanbuljs/nyc-config-*`.

Not sure it should be mentioned here but the `--require` flag also operates on the main nyc process for use by `--all`.  In situations with `nyc --all true --instrument false` and `babel-plugin-istanbul` setup the `--all` option only works if `--require @babel/register` is passed to nyc.  Passing it to mocha would cause the tests to be instrumented but unloaded sources would not be seen.

README.md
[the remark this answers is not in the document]
JaKXz: @bcoe I think (/feel strongly) the npx first is still more effective, and we can still have people install nyc@next with npx or `npm i`

I was debating whether I should specify `npx nyc@latest` in the default instructions, and I think I will so that future troubleshooting instructions can be to try `npx nyc@next` or whatever tag we need to specify. 

Finding out which version is not difficult, so we can add that to the issue template.

README.md
JaKXz: Will do thanks

JaKXz:
@bcoe @coreyfarrell thanks for your feedback - I've made some changes and additions. Please try the new link and let me know what you think: https://github.com/istanbuljs/nyc/tree/3e0624e

README.md
JaKXz: Added this section!

README.md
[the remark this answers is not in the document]
JaKXz: @bcoe after chatting with @coreyfarrell I have been convinced. I have fixed this section so I'm resolving this.

README.md
```diff
@@ -11,32 +11,34 @@ _Having problems? want to contribute? join our [community slack](http://devtools
...
+    "test": "mocha",
+    "coverage": "nyc npm run test"
   }
 }
 ```
 
-Or use your package manager to add it as a dev dependency: `npm i -D nyc` or `yarn add -D nyc`.
-You can also use nyc to call npm scripts (assuming they don't already have nyc executed in them), like so:
+You can use also `npx` instead of installing nyc as a dependency, but you might get updates you are not ready for; to get around this, pin to a specific major version by specifying, e.g. `nyc@14`.
+In general, simply add `nyc` before your test runner is invoked:
```
coreyfarrell: simply add `nyc`..?  Should this say `npx nyc@latest`?  Maybe just remove this line?

Other than this line it everything looks good.

coreyfarrell:
Thanks for this great work @JaKXz!

PR #1100, opened 2019-05-06
fix(cli): Report error if unwanted positional arguments are received

### LINKED ISSUE #401: `nyc report html` should maybe tell users that it’s not what they want?
Okay, probably it’s just me having trained myself to use `istanbul report` over a long time, but I tend to always write `nyc report html` and am confused by the result that there is no actual HTML output.
I’ve begun to catch up always correcting to `-r html`, but maybe other people who transition to `nyc` would find it helpful if `nyc report` outputs a warning when it receives at least one positional argument?

What do you think?

coreyfarrell:
This applies to check-coverage, instrument, merge and report.  Passing
additional arguments will now cause the help script to be displayed and
an error exit code.

Unknown flags `nyc report --unknown=1` are still not reported.
Reporting unknown flags would require additional work as
`yargs.strict()` causes unknown items from configuration to be reported,
including flags that are defined for the global command but not a
sub-command.

Fixes #401

---

Is this a breaking change?  I lean towards no but can see the argument.  Likely this will first be released to 15.0.0 anyways but I'm not sure if it should be noted under breaking changes in the changelog.

### DIFF
--- a/test/nyc-integration.js
+++ b/test/nyc-integration.js
@@ -45,11 +45,11 @@ t.test('--check-coverage fails when the expected coverage is below a threshold',
 }))
 
 // https://github.com/istanbuljs/nyc/issues/384
-t.test('--check-coverage fails when check-coverage command is used rather than flag', t => {
+t.test('check-coverage command is equivalent to the flag', t => {
   return testSuccess(t, {
     args: [process.execPath, './half-covered.js']
   }).then(() => testFailure(t, {
-    args: ['check-coverage', '--lines', '51', process.execPath, './half-covered.js']
+    args: ['check-coverage', '--lines', '51']
   }))
 })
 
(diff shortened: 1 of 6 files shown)


test/nyc-integration.js
```diff
@@ -49,7 +49,7 @@ t.test('--check-coverage fails when check-coverage command is used rather than f
   return testSuccess(t, {
     args: [process.execPath, './half-covered.js']
   }).then(() => testFailure(t, {
-    args: ['check-coverage', '--lines', '51', process.execPath, './half-covered.js']
+    args: ['check-coverage', '--lines', '51']
```
JaKXz: Could we rename this test to something like `check-coverage command is equivalent to the flag`

PR #1107, opened 2019-05-09
docs: Point to updated location of default exclude list

coreyfarrell:


README.md
```diff
@@ -100,7 +100,7 @@ This table is a quick TLDR for the rest of this readme and there are more advanc
 | `check-coverage` | Check whether coverage is within thresholds, fail if not | `Boolean` | `false` |
 | `extension` | List of extensions that nyc should attempt to handle in addition to `.js` | `Array<String>` | `['.js']` |
 | `include` | See [selecting files for coverage] for more info | `Array<String>` | `['**']`|
-| `exclude` | See [selecting files for coverage] for more info | `Array<String>` | [list](https://github.com/istanbuljs/istanbuljs/blob/master/packages/test-exclude/index.js#L176-L184) |
+| `exclude` | See [selecting files for coverage] for more info | `Array<String>` | [list](https://github.com/istanbuljs/istanbuljs/blob/master/packages/test-exclude/default-exclude.js) |
```
JaKXz: This list is also mentioned (and already outdated) in the selecting files for coverage section. We should fix that, ideally with a github embed in the readme in that section rather than the static list...

coreyfarrell: I can link the content but I'm not sure how to embed content in a README?  I think that might be a special feature just for issue/PR comments?

PR #1133, opened 2019-06-19
chore: Use snapshots for spawn output testing

coreyfarrell:


coreyfarrell:
@bcoe @JaKXz this is necessary before I can update to the istanbuljs alpha deps.  The issue is that the `text` report has changed and we have a ton of tests which test the output of that report.  This way I'll be able to just regenerate the snapshots after upgrading and verify with `git diff` that formatting changed but data remained the same.

PR #1134, opened 2019-06-20
feat!: Drop node.js 6, upgrade dependencies

coreyfarrell:


lukeapage:
I don't know if its related to this change or istanbul, but I made instructions to test the master of the instabuljs mono-repo with it dog-fooding its own source, for trying out report changes:

https://gist.github.com/lukeapage/c7844b3bd7fab7eed260bfef387848b9

After fixing the none report bug though, it still doesn't work.

It looks like the merge command on this pr as run here:
https://github.com/istanbuljs/istanbuljs/blob/master/monorepo-merge-reports.js#L19

doesn't merge, but generates the reports as it might do with normal report generation.

coreyfarrell:
The issue is not node.js 8, it's Windows (yes, how shocking).  I suspect the problem is due to https://github.com/istanbuljs/istanbuljs/pull/422, not sure if this is exposing a failure in test-exclude or if the nyc tests are just wrong for Windows.  It's unlikely I'll be able to work on this before next week.

coreyfarrell:
Finally got around to dealing with the test.  Basic issue is that testExclude.shouldInstrument no longer accepts `/` based paths on Windows, the root path needs to be drive based (thus `C://` or similar).  This is a non-issue in practice, the problem is that when the test ran `nyc.exclude.shouldInstrument('/cwd/foo.js', 'foo.js')`, this is something that would never happen in real life.

PR #1138, opened 2019-07-04
chore: Remove test artefacts that can ensure failure on subsequent tests

AndrewFinlay:
If the tests fail with a bad 'test/fixtures/cli/foo-cache', then subsequent tests will also fail unless the cache is cleaned.
A failed test run also leaves a lot of files in 'test/temp-dir-nyc-integration' that are cleaned on success, but not before the start of the next run in case of failure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "test": "tap -t360 --no-cov -b test/*.js",
     "snap": "cross-env TAP_SNAPSHOT=1 npm test",
     "posttest": "npm run report",
-    "clean": "rimraf ./.nyc_output ./node_modules/.cache ./.self_coverage ./test/fixtures/.nyc_output ./test/fixtures/node_modules/.cache ./self-coverage",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "test": "tap -t360 --no-cov -b test/*.js",
     "snap": "cross-env TAP_SNAPSHOT=1 npm test",
     "posttest": "npm run report",
-    "clean": "rimraf ./.nyc_output ./node_modules/.cache ./.self_coverage ./test/fixtures/.nyc_output ./test/fixtures/node_modules/.cache ./self-coverage",
+    "clean": "rimraf ./.nyc_output ./node_modules/.cache ./.self_coverage ./test/fixtures/.nyc_output ./test/fixtures/node_modules/.cache ./test/fixtures/cli/foo-cache ./test/temp-dir-nyc-integration ./self-coverage",
```
coreyfarrell: Instead of `./test/temp-dir-nyc-integration` lets use `./test/temp-dir-*`.  The same temp-dir pattern is used by test/config-override.js, will potentially be used by other test files in the future.

PR #1144, opened 2019-07-12
feat: Use source base name to prefix cache files

coreyfarrell:
This utilizes a new feature of caching-transform@4 to prefix cache instrumented sources and source-maps with a more useful name.  For example `lib/worker.js` might get cached as `node_modules/.cache/nyc/worker-d84d3c7a86d2657c33e0c6f4fe3013270d7887ae76e3bf14bde6286cffc79b47.js`.  Also added a test to verify creation of the cached source map when appropriate.

### DIFF
--- a/index.js
+++ b/index.js
@@ -93,9 +93,10 @@ class NYC {
   }
 
   _createTransform (ext) {
-    var opts = {
+    const opts = {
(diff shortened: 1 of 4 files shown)


bcoe:
this seems reasonable to me, potentially helps with debugging?

Issue #1135, opened 2019-06-26
Support in-place instrumentation (again)


DanTup:
It used to be possible to instrument in-place, but this is rejected since https://github.com/istanbuljs/nyc/pull/1007.

It's not clear from the PR why, but it broke things for me, and trying to use a different folder is complicated (because there are many paths into the JS folder).

I understand this can be risky, but in my case it's just the transpiled output of typescript. It'd be nice if this was supported, even if you had to pass an extra flag to opt-in to it.

coreyfarrell:
We require a link to a repository showing an example project where this not working.

DanTup:
All you need to do is run `npm instrument` with the same input/output on any project. The code checks for this and exits with an error (as of the PR linked above):

https://github.com/istanbuljs/nyc/blob/4bc9c146d18c988dda2a7c20bce8d861c370245e/lib/commands/instrument.js#L96-L99

I haven't committed the upgrade that breaks this because I need my coverage to work in the meantime. This is a feature request to support this again.

coreyfarrell:
@AndrewFinlay do you have any thoughts?  I know this needs to be blocked if `--delete` is enabled, outside of that I'm just not sure.  It seems dangerous to rewrite files in-place, though it appears that `@babel/cli` allows it (a bug or a feature?).

@DanTup If you provided a link to a repository which uses in-place instrumentation I might be able to suggest a better way.  Maybe `nyc instrument --complete-copy=true src dest` could work for you.

DanTup:
> It seems dangerous to rewrite files in-place

Many things are dangerous, but making them impossible is really restrictive. This feature used to work, and I'm running it on the output of TypeScript so there's really no problem. You could make a `--force` or `--some-really-scary-flags-that-acknowledge-the-danger` flag if you're concerned people won't understand - but taking it away just means people will hang around on old versions if there's no easy way to switch to out-of-place instrumentation.

I did try for some time to make the new version work, but there many places reference this folder (such as all the VS Code test entry points) so I gave up and just rolled back (I'd already spent several hours trying to get this working again after adding webpack - after spending many hours getting it working the first time!).

My setup is super-janky (for a whole host of reasons, involving VS Code, webpack, mismatched source map paths between packed/unpacked and some weird behaviour from nyc depending on which folder it's executed from within):

https://github.com/Dart-Code/Dart-Code/blob/7c64f5e3de13afa1c64f4b0a56877e74d7264c3a/package.json#L1454-L1468

[...]

Issue #1096, opened 2019-05-02
testing: Use isolated cwd for `--all` tests


coreyfarrell:
Tests which use `--all` should be run from a `cwd` which contains a very limited number of files.  Currently any change to files within `test/fixtures/cli` causes snapshot output produced by `nyc --all` to change.

PR #1149, opened 2019-07-21
feat: Allow `nyc instrument` to instrument code in place

AndrewFinlay:
This PR addresses #1135

This change adds the `--in-place` switch to nyc instrument
If `--in-place` is specified, the output directory will be set to equal the input directory for the instrument command.
This has the effect of replacing any file in the input directory that should be instrumented, with its instrumented counterpart.
The command will throw an error if the --delete option is specified.
The only way to instrument in place is with the --in-place switch, setting the input and output directories to be the same will not work
If `--in-place` is set the instrument command ignores any output directory specified with the command
If `--in-place` is set the instrument command will disable the `--complete-copy` switch as it is unnecessary.

I've made a few small code improvements to the instrument command spec.
I've also added tests to the old integration tests folder, I figured I could put tests here for the time being before they get moved to the new tap format (by me or someone else).

### DIFF
--- a/docs/instrument.md
+++ b/docs/instrument.md
@@ -20,6 +20,8 @@ For example, `nyc instrument . ./output` will produce instrumented versions of a
 
 The `--delete` option will remove the existing output directory before instrumenting.
 
+The '--in-place' option will allow you to run the instrument command
+
 The `--complete-copy` option will copy all remaining files from the `input` directory to the `output` directory.
 When using `--complete-copy` nyc will not copy the contents of a `.git` folder to the output directory.
 
@@ -27,8 +29,8 @@ When using `--complete-copy` nyc will not copy the contents of a `.git` folder t
 
 ## Streaming instrumentation
 
-`nyc instrument` will stream instrumented source directly to `stdout`, that can then be piped to another process.
-You can use this behaviour to create a server that can instrument files on request.
+`nyc instrument <input>` will stream instrumented source directly to `stdout` and that output can then be piped to another process.
+You can use this behaviour to create a server that dynamically instruments files on request.
 The following example shows streaming instrumentation middleware capable of instrumenting files on request.
 
 ```javascript
--- a/docs/instrument.md
+++ b/docs/instrument.md
@@ -20,7 +20,7 @@ For example, `nyc instrument . ./output` will produce instrumented versions of a
 
 The `--delete` option will remove the existing output directory before instrumenting.
 
-The '--in-place' option will allow you to run the instrument command
+The `--in-place` option will allow you to run the instrument command.
 
 The `--complete-copy` option will copy all remaining files from the `input` directory to the `output` directory.
 When using `--complete-copy` nyc will not copy the contents of a `.git` folder to the output directory.
--- a/lib/commands/instrument.js
+++ b/lib/commands/instrument.js
@@ -31,32 +31,37 @@ exports.builder = function (yargs) {
     .option('source-map', {
       default: true,
       type: 'boolean',
-      description: 'should nyc detect and handle source maps?'
+      describe: 'should nyc detect and handle source maps?'
     })
     .option('produce-source-map', {
       default: false,
       type: 'boolean',
-      description: "should nyc's instrumenter produce source maps?"
+      describe: "should nyc's instrumenter produce source maps?"
     })
     .option('compact', {
       default: true,
       type: 'boolean',
-      description: 'should the output be compacted?'
+      describe: 'should the output be compacted?'
     })
     .option('preserve-comments', {
       default: true,
       type: 'boolean',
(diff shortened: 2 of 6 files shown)


AndrewFinlay:
Just a quick note that the I'm aware that this breaks the test snapshot and you probably don't want additions to the old integration test file.  Since I'm a little low on time at the moment I figured that it'd be good to get this out there and it'd be quicker to make tests for the old integration set that I'm familiar with.  Anyway let me know what needs to be done from here, thanks

docs/instrument.md
```diff
@@ -20,15 +20,17 @@ For example, `nyc instrument . ./output` will produce instrumented versions of a
 
 The `--delete` option will remove the existing output directory before instrumenting.
 
+The '--in-place' option will allow you to run the instrument command
```
coreyfarrell: End with a `.`?

lib/commands/instrument.js
```diff
@@ -92,40 +97,45 @@ exports.builder = function (yargs) {
...
+
   // If instrument is set to false enable a noop instrumenter.
   argv.instrumenter = (argv.instrument)
     ? './lib/instrumenters/istanbul'
     : './lib/instrumenters/noop'
 
-  const nyc = new NYC(argv)
+  if (argv.inPlace) {
+    argv.output = argv.input
+    argv.completeCopy = false
```
coreyfarrell: Should `--in-place --complete-copy` be an error like `--delete --in-place`?

test/nyc-integration-old.js
```diff
@@ -329,7 +329,53 @@ describe('the nyc cli', function () {
 
         const proc = spawn(process.execPath, args, {
           cwd: fixturesCLI,
-          env: env
+          env
+        })
+
+        let stderr = ''
+        proc.stderr.on('data', function (chunk) {
```
coreyfarrell: Can we drop the `--delete` option from this test so we test just `nyc instrument ./ ./`?  You have a test for blocking in-place delete below.

test/nyc-integration-old.js
```diff
@@ -329,7 +329,53 @@ describe('the nyc cli', function () {
...
+
+        proc.on('close', function (code) {
+          code.should.equal(1)
+          stderr.should.include('cannot instrument files in place')
+          done()
+        })
+      })
+
+      it('can write files in place with --in-place switch', function (done) {
+        const args = [bin, 'instrument', '--in-place', '--include', '*/file1.js', './instrument-inplace']
```
coreyfarrell: What happens if we run the test twice?  I feel like we need to make a copy of the `instrument-inplace` directory then run the test on the copy.

coreyfarrell:
Additional note the snapshots getting broken by this test is actually referenced in #1096.  I just haven't had time to tweak those tests which use `--all` to use an isolated directory.

docs/instrument.md
AndrewFinlay: I also had single quotes instead of back ticks... eek.

lib/commands/instrument.js
AndrewFinlay: I'm thinking it should be a no-op and shouldn't require a warning.  If you could enable the `--complete-copy` switch for `instrument --in-place` you would expect the same result whether or not it was active.

test/nyc-integration-old.js
AndrewFinlay: Initially I was thinking of a solution along the lines of making a copy of a test directory then instrumenting in place, then I thought it probably wouldn't matter for the CI infrastructure.  It will of course matter when running in your local env, so I'll make this a bit friendlier.

lib/commands/instrument.js
```diff
@@ -92,40 +97,45 @@ exports.builder = function (yargs) {
...
     if (relPath !== '' && !relPath.startsWith('..')) {
       rimraf.sync(argv.output)
     } else {
-      console.error(`nyc instrument failed: attempt to delete '${process.cwd()}' or containing directory.`)
-      process.exit(1)
+      return errorExit(`attempt to delete '${process.cwd()}' or containing directory.`)
     }
   }
 
+  console.log(`Running nyc from ${process.cwd()}`)
```
coreyfarrell: This will break streaming instrumentation, must be removed.  Example:
```sh
nyc instrument something.js > instrumented/something.js
```

AndrewFinlay: That's a bit embarrassing, that's some code I drop in while testing nyc from another project to confirm that my npm link has worked.  I think I should spend a little more time reviewing my code and not commit in such a hurry.

Thanks for the heads up Corey.

AndrewFinlay: Updating snapshots...

coreyfarrell: It happens, this is why we do reviews :smile:

PR #1169, opened 2019-09-06
feat: Add `--use-spawn-wrap=true` option

coreyfarrell:
Completely bypass spawn-wrap unless overridden with this new option.


bin/nyc.js
```diff
@@ -36,21 +34,60 @@ if ([
...
+     * Node.js 11 and older do not support specifying `--require "/path/with spaces"`.
+     *  - For node.js 12+ we use the proper quoted syntax.
+     *  - For older versions we just directly use the path to lib/wrap.js as long as
+     *    it does not contain any spaces.
+     *  - For older versions when the path to lib/wrap.js contains a space we must
+     *    combine NODE_PATH and NODE_OPTIONS.
+     *
+     * Ref https://github.com/nodejs/node/pull/24065
+     */
+    /* istanbul ignore next */
```
coreyfarrell: The branches following this line are version dependent.  I have tested all three branches locally using node.js 8, 10 and 12.  I've tested with nyc in a directory with and without spaces and double-quote.

isaacs:
This is cool.  When did the support for `env.NODE_OPTIONS` land?

coreyfarrell:
> This is cool. When did the support for `env.NODE_OPTIONS` land?

node.js 8.0.0 added the option: https://nodejs.org/dist/latest-v8.x/docs/api/cli.html#cli_node_options_options

node.js < 12 does not support quoting the path given to `--require` so it is not possible to directly do:
```js
process.env.NODE_OPTIONS = `--require ${pathWithSpaces}`
```

bin/nyc.js
```diff
@@ -36,21 +34,60 @@ if ([
...
+        nodePath = process.env.NODE_PATH + sep + nodePath
+      }
 
-  if (argv['babel-cache'] === false) {
-    // babel's cache interferes with some configurations, so is
-    // disabled by default. opt in by setting babel-cache=true.
-    env.BABEL_DISABLE_CACHE = process.env.BABEL_DISABLE_CACHE = '1'
-  }
+      process.env.NODE_PATH = nodePath
+      nodeRequire = ` --require wrap.js`
```
coreyfarrell: Using NODE_PATH in this way is probably sketchy.  I'm concerned this could hijack unrelated `require` actions by exposing the `node_modules` in nyc to code that shouldn't be using it.  Likely only an issue if `NODE_PATH` is already being used, but still a concern.

isaacs:
I was thinking about the case where someone clears the environment variable before calling spawn, like `spawn(process.execPath, ['foo.js'], { env: { NODE_OPTIONS: 'different' } })`.

We could update nyc's `wrap.js` to munge ChildProcess.spawn kind of like how spawn-wrap does, but without writing a shim file, such that it _always_ puts the environment variable back in.

Then the only hazard would be something like `spawn('bash', ['escape.sh'])` where `escape.sh` does `NODE_OPTIONS= node foo.js`, but that's enough of an explicit step that we should probably respect it anyway.

coreyfarrell:
> I was thinking about the case where someone clears the environment variable before calling spawn, like `spawn(process.execPath, ['foo.js'], { env: { NODE_OPTIONS: 'different' } })`.
> 
> We could update nyc's `wrap.js` to munge ChildProcess.spawn kind of like how spawn-wrap does, but without writing a shim file, such that it _always_ puts the environment variable back in.

I thought of this, I've created https://github.com/cfware/node-preload as a potential alternative to spawn-wrap.  The functionality has enough platform/version edge cases that it probably deserves it's own testing.  I'm actually looking to get it to a state where it might be possible to get the functionality into the node.js core.  My module is not published to npm yet, I want to get a little feedback first.

> Then the only hazard would be something like `spawn('bash', ['escape.sh'])` where `escape.sh` does `NODE_OPTIONS= node foo.js`, but that's enough of an explicit step that we should probably respect it anyway.

Agreed.  Also we can't control everything.

bin/nyc.js
JaKXz: Does the ENOMEM error still occur on Windows with node 10+?

bin/nyc.js
```diff
@@ -36,21 +34,60 @@ if ([
...
 
-  var env = {
-    // Support running nyc as a user without HOME (e.g. linux 'nobody'),
-    // https://github.com/istanbuljs/nyc/issues/951
-    SPAWN_WRAP_SHIM_ROOT: process.env.SPAWN_WRAP_SHIM_ROOT || process.env.XDG_CACHE_HOME || require('os').homedir(),
-    NYC_CONFIG: JSON.stringify(argv),
-    NYC_CWD: process.cwd()
-  }
+  if (argv.setNodeOptions) {
+    const wrapper = require.resolve('../lib/wrap.js')
```
coreyfarrell: The idea with `node-preload` is that we would replace this conditional block with the following:
```js
const nodePreload = require('node-preload')
nodePreload.preloadInsert(require.resolve('../lib/wrap.js'))
nodePreload.propagateEnv('NYC_CWD', process.cwd())
nodePreload.propagateEnv('NYC_CONFIG', JSON.stringify(argv))
if (argv['babel-cache'] === false) {
  nodePreload.propagateEnv('BABEL_DISABLE_CACHE', '1');
  process.env.BABEL_DISABLE_CACHE = 1;
}
```

Obviously I welcome comments/suggestions on improving the API of this.

bin/nyc.js
coreyfarrell: It does still occur, specifically on 10.16.0 and above.  Technically we could add `10.15.0` into testing.

See test log from yesterday when testing node-preload - https://travis-ci.org/cfware/node-preload/builds/582661358

PR #1212, opened 2019-10-23
Add support for relative path option

bgalletbgallet:
Add the possibility to use --relative=true option in command line
This is useful if you run your test on a linux environment (docker, vm, server) and dev on a windows one.
If you want your coverage to work with your IDE, you need relative path instead of absolute ones.

Linked with pull request https://github.com/istanbuljs/istanbuljs/pull/492

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,7 +57,6 @@ class NYC {
     this._showProcessTree = config.showProcessTree || false
     this._eagerInstantiation = config.eager || false
     this.cwd = config.cwd || process.cwd()
-    this.projectRoot = config.projectRoot || this.cwd
     this.reporter = [].concat(config.reporter || 'text')
 
     this.cacheDirectory = (config.cacheDir && path.resolve(config.cacheDir)) || findCacheDir({ name: 'nyc', cwd: this.cwd })
@@ -441,7 +440,7 @@ class NYC {
       reports.create(_reporter, {
         skipEmpty: this.config.skipEmpty,
         skipFull: this.config.skipFull,
-        projectRoot: this.projectRoot,
+        projectRoot: this.cwd,
         maxCols: process.stdout.columns || 100
       }).execute(context)
     })
--- a/index.js
+++ b/index.js
@@ -428,6 +428,7 @@ class NYC {
       reports.create(_reporter, {
         skipEmpty: this.config.skipEmpty,
         skipFull: this.config.skipFull,
+        relative: this.config.relative === 'true',
         maxCols: process.stdout.columns || 100
       }).execute(context)
     })
--- a/index.js
+++ b/index.js
@@ -57,6 +57,7 @@ class NYC {
     this._showProcessTree = config.showProcessTree || false
     this._eagerInstantiation = config.eager || false
     this.cwd = config.cwd || process.cwd()
+    this.projectRoot = config.projectRoot || this.cwd
     this.reporter = [].concat(config.reporter || 'text')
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -428,6 +428,7 @@ class NYC {
       reports.create(_reporter, {
         skipEmpty: this.config.skipEmpty,
         skipFull: this.config.skipFull,
+        relative: this.config.relative === 'true',
```
coreyfarrell:
```suggestion
        cwd: this.cwd,
```

bgalletbgallet: Hi @coreyfarrell 
Shouldn't I use this.config.cwd to access the cwd option ?

coreyfarrell: No, please use `this.cwd`.  In some edge cases it's possible that `this.config.cwd` might be unset but this.cwd has the default `process.cwd()`.

https://github.com/istanbuljs/nyc/blob/master/index.js#L60

bgalletbgallet: Perfect I didn't see it was already set. I do it tonight
Thank you for your anwser

coreyfarrell: I've altered the suggested change per my latest comment on the istanbuljs PR.

bgalletbgallet:
@coreyfarrell Everything should be ok, the issue with --cwd option is still present but this PR is not concerned anymore.
By default the lcov report is relative. A customer projectRoot option can be used to change the 'from' argument of the path.relative() function

coreyfarrell:
@bcoe This adds a `projectRoot` option to nyc which could be set differently than `cwd`.  Do you think this is useful?  I'm concerned this could cause confusion for users.

Ref #1234 - the config loading is complicated and potentially needs some tweaks before we do a final release of nyc 15.

bcoe:
@coreyfarrell this would be extremely useful to me, I'm currently floating a patch for the Node.js nightly builds, with almost this functionality (so that merging works):

https://codecov.io/gh/nodejs/node/src/64ee64d3afc837350ee39058cef9179e92f3554c/lib/fs.js

coreyfarrell:
@bgalletbgallet could you explain to me why we need a separate option for `projectRoot` compared to `cwd`?  I'm very concerned this will cause confusion to end users, and interested in knowing about a scenario where the value of `cwd` will would be unacceptable as the base of reporting.  Keep in mind that by default `cwd` will be the directory containing the package.json which applies to `process.cwd()`.  I'm afraid to add complexity without understanding the justification.

A minimal reproduction which installs `git://github.com/bgalletbgallet/nyc#master` for the nyc dependency showing that it is necessary to set `projectRoot` to get functional results would be great.

bgalletbgallet:
@coreyfarrell I don't know, I was about to delete it and @bcoe told that it was usefull to him. I don't see any case where you need a different projectRoot than the default cwd. If it's ok for @bcoe, we will delete it

index.js
```diff
@@ -440,6 +441,7 @@ class NYC {
       reports.create(_reporter, {
         skipEmpty: this.config.skipEmpty,
         skipFull: this.config.skipFull,
+        projectRoot: this.projectRoot,
```
coreyfarrell:
```suggestion
        projectRoot: this.cwd,
```

index.js
```diff
@@ -57,6 +57,7 @@ class NYC {
     this._showProcessTree = config.showProcessTree || false
     this._eagerInstantiation = config.eager || false
     this.cwd = config.cwd || process.cwd()
+    this.projectRoot = config.projectRoot || this.cwd
```
coreyfarrell: Let's please delete this line.

PR #1217, opened 2019-11-01
Add enterprise language

phated:
This adds the enterprise language for Tidelift with the nyc tracking code.

I think we should use this in the base of the `istanbuljs` repo too.

@coreyfarrell @bcoe Any objections to me adding a "For Enterprise" page to the website?


coreyfarrell:
Thanks @phated.  I have no objection to adding something about enterprise support to the webpage (I assume you are talking about istanbul.js.org).  Not sure how/where that should go?

PR #1237, opened 2019-12-04
docs: link to next tracking issue in changelog

### LINKED ISSUE #1236: Where is the changelog for v15 beta releases?
There are no git release tags, no entries in CHANGELOG.md, no version commits on master, and no special v15 branch. I honestly don't even understand where the v15 beta versions in npm are coming from. Is it moving to a new repo? Or is someone manually publishing version to the npm registry without tagging version in git?

How do I know whether to upgrade in my repos?

Thanks!

serhalp:
Information about the next version in development, which is published to the NPM registry without git version tags and without CHANGELOG entries, is tracked in an open Github issue. This is not very discoverable on its own. This makes it more discoverable by linking to it from the master CHANGELOG.

Closes #1236.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## Next
+
+The change log for the `next` version (15) can be found [here](https://github.com/istanbuljs/nyc/issues/1104).
+
 ## [14.1.1](https://github.com/istanbuljs/nyc/compare/v14.1.0...v14.1.1) (2019-05-09)
 
 


coreyfarrell:
Our changelog is auto-generated so I cannot merge this change.  The plan is that nyc is not far from final release.  In the future I will make a point to always pin the 'New Pending Release' issue so it is more discoverable, I think that will be adequate.  Also note that side-publishing is only done for semver-major beta releases.  In the case of for example 15.0.1 although we will briefly publish to npm under the `next` tag but that release will get changelogs committed, tagged and immediately pushed to github.  It's only in the case of beta's where I want to ensure all breaking changes are listed under 15.0.0 instead of forcing users to review one or more betas.

PR #1247, opened 2019-12-24
(chore) Remove the unused js-yaml dependency.

XhmikosR:
It seems this is now moved to @istanbuljs/load-nyc-config


coreyfarrell:
Thanks for catching this!

PR #1250, opened 2019-12-27
(chore): Update standard to v14.3.1

XhmikosR:
I just ran `npm run lint -- --fix`.


XhmikosR:
Maybe tap-snapshots should be excluded from linting?

coreyfarrell:
`tap-snapshots` definitely needs to be excluded.  To be honest I'm considering switching to use eslint directly and potentially some other rule changes.  For now let's hold on this PR.

XhmikosR:
I will exclude tap-snapshots and I still think this should be merged after that. I will push a few more tweaks to my #1253 PR.

PR #1251, opened 2019-12-27
(chore): Update decamelize to v3.2.0

XhmikosR:


coreyfarrell:
Thanks for this PR but the decamelize dependency is being held back on purpose due to current version depending on `xregexp` which is 9.6MB.  This is also the case for yargs, see https://github.com/yargs/yargs/pull/1377.

XhmikosR:
Oh, cool I wasn't aware of that.

PR #1252, opened 2019-12-27
(chore) Remove the unused uuid dependency

XhmikosR:
This seems to be included in istanbul-lib-processinfo, but please confirm.


package.json
```diff
@@ -72,15 +72,14 @@
     "istanbul-lib-source-maps": "^4.0.0",
     "istanbul-reports": "^3.0.0",
     "make-dir": "^3.0.0",
+    "node-preload": "^0.2.0",
```
XhmikosR: The sorting is done automatically by npm

PR #1253, opened 2019-12-27
(chore) Minor ES6 tweaks.

XhmikosR:
These aren't caught by standard but shouldn't hurt nevertheless. Let me know if you want something changed, I did this automatically, not by hand :)


XhmikosR:
This is ready for review. I'd still like to have #1250 merged before this one

XhmikosR:
Most of these were found with [xo](https://github.com/XhmikosR/nyc/tree/xo) but this is pretty opinionated so I'll leave this to the maintainers to decide later :)

XhmikosR:
@coreyfarrell should I drop this?

PR #1255, opened 2019-12-27
(chore) Switch to AppVeyor for Windows testing

XhmikosR:
@coreyfarrell the only changes you should make in AppVeyor settings are these:

* Skip branches without appveyor.yml
* Save build cache in Pull Requests

Notes:

- [x] we could always install the latest version of Node.js; if you want this let me know
- [x] I haven't added Node.js 13, should I add it?
- [x] I test one x86 build and everything else x64. I don't think it matters, so maybe I should just use x64 for all versions?
- [x] I have added caching (caching the global npm folder) which will be invalidated if any of these files changes: appveyor.yml, package.json, package-lock.json

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 # nyc
 
-[![Build Status](https://travis-ci.org/istanbuljs/nyc.svg?branch=master)](https://travis-ci.org/istanbuljs/nyc)
-[![Coverage Status](https://coveralls.io/repos/istanbuljs/nyc/badge.svg?branch=)](https://coveralls.io/r/istanbuljs/nyc?branch=master)
+[![Build Status](https://img.shields.io/travis/istanbuljs/nyc/master.svg)](https://travis-ci.org/istanbuljs/nyc)
+[![Windows Build Status](https://img.shields.io/appveyor/ci/istanbuljs/nyc/master.svg?label=Windows%20build)](https://ci.appveyor.com/project/istanbuljs/nyc/branch/master)
+[![Coverage Status](https://img.shields.io/coveralls/github/istanbuljs/nyc/master.svg)](https://coveralls.io/r/istanbuljs/nyc?branch=master)
 [![NPM version](https://img.shields.io/npm/v/nyc.svg)](https://www.npmjs.com/package/nyc)
 [![Conventional Commits](https://img.shields.io/badge/Conventional%20Commits-1.0.0-yellow.svg)](https://conventionalcommits.org)
 [![community slack](https://devtoolscommunity.herokuapp.com/badge.svg)](https://devtoolscommunity.herokuapp.com)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,31 @@
+version: "{build}"
+
+shallow_clone: true
+
+environment:
+  matrix:
+    - NODEJS_VERSION: "13"
+      PLATFORM: x64
+    - NODEJS_VERSION: "12"
+      PLATFORM: x64
+    - NODEJS_VERSION: "10"
+      PLATFORM: x64
+    - NODEJS_VERSION: "8"
+      PLATFORM: x64
+
+install:
+  - ps: Update-NodeJsInstallation (Get-NodeJsLatestBuild $env:NODEJS_VERSION) $env:PLATFORM
+  - npm ci
+
+test_script:
+  - node --version
+  - npm --version
+  - npm test
+
+build: off
+
+matrix:
+  fast_finish: true
+
+cache:
+  - '%APPDATA%\npm-cache\ -> appveyor.yml,package.json,package-lock.json'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,11 @@
 language: node_js
 os:
-  - windows
   - linux
   - osx
 node_js:
   - "node"
   - 10
   - 8
-matrix:
-## An ENOMEM error occurs with 10+ under Travis-CI for Windows.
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -1,7 +1,8 @@
 # nyc
 
-[![Build Status](https://travis-ci.org/istanbuljs/nyc.svg?branch=master)](https://travis-ci.org/istanbuljs/nyc)
-[![Coverage Status](https://coveralls.io/repos/istanbuljs/nyc/badge.svg?branch=)](https://coveralls.io/r/istanbuljs/nyc?branch=master)
+[![Linux Build Status](https://img.shields.io/travis/istanbuljs/nyc/master.svg?label=Linux%20build)](https://travis-ci.org/istanbuljs/nyc)
+[![Windows Build Status](https://img.shields.io/appveyor/ci/istanbuljs/nyc/master.svg?label=Windows%20build)](https://ci.appveyor.com/project/istanbuljs/nyc/branch/master)
+[![Coverage Status](https://img.shields.io/coveralls/github/istanbuljs/nyc/master.svg)](https://coveralls.io/r/istanbuljs/nyc?branch=master)
```
coreyfarrell: Why change the badge URL?

appveyor.yml
```diff
@@ -0,0 +1,28 @@
+version: "{build}"
+
+shallow_clone: true
+
+environment:
+  matrix:
+    - NODEJS_VERSION: "8"
+      PLATFORM: x86
```
coreyfarrell: I think we're fine without x86 specific testing, let's just drop this.  It can always be added later if we find that it's needed.

appveyor.yml
```diff
@@ -0,0 +1,28 @@
+version: "{build}"
+
+shallow_clone: true
+
+environment:
+  matrix:
```
coreyfarrell: Nit: can this be sorted latest version first like we do in .travis.yml?

README.md
XhmikosR: Just so that every badge is using shields.io. I can revert this, I just find it more consistent.

appveyor.yml
XhmikosR: Yeah they have some aliases. I will add it. But how about always installing the latest version? For example, right now Node.js 12.4.0 is out but AppVeyor is still using 12.13.1

README.md
coreyfarrell: Hmmm actually leave this, it will improve HTTP connection sharing.

XhmikosR: Yup, that's my main reason for the switch too.

XhmikosR:
Alright, I pushed my changes, if it passes I will squash the patches.

So, assuming you made the two changes in settings I mentioned, and you are OK with how this works, you can merge after that.

Forgot to mention I have specified `shallow_clone: true` which is basically cloning with `depth: 1`

appveyor.yml
coreyfarrell: What I meant is let's have appveyor.yml list "Current", then 12, then 10, then 8 in the matrix (newest first instead of oldest first).

As for which versions to run the goal is to always run the latest of each supported semver-major.  For nyc 15 this is node.js 8, 10, 12 and for now 13.

XhmikosR: Alright, I will switch these. BTW Travis is not testing Node.js 12 anymore since `node` means 13

coreyfarrell:
> Alright, I pushed my changes, if it passes I will squash the patches.

No need, I use `Squash and merge` on the github interface even if you already squashed.  We use commit logs for changelog generation and this gives us the best results (the PR is added to the first line of the commit).

> So, assuming you made the two changes in settings I mentioned, and you are OK with how this works, you can merge after that.

Settings are changed per your suggestion.

> Forgot to mention I have specified `shallow_clone: true` which is basically cloning with `depth: 1`

This is fine.

XhmikosR:
Some further notes:

>No need, I use Squash and merge on the github interface even if you already squashed. We use commit logs for changelog generation and this gives us the best results (the PR is added to the first line of the commit).

I'm still old school :P

1. Always installing the latest Node.js version results in some slowness, ~1min 30s
2. The free AppVeyor plan is limited in concurrency unfortunately

I will push the last changes and if everything is green you can merge.

XhmikosR:
OK so it seems the `Current` alias does not work with always using the latest version, or at least I couldn't make it work. I just used `13` for now.

XhmikosR:
Should I add `fast_finish: true`?

coreyfarrell:
> Should I add `fast_finish: true`?

Yes please

XhmikosR:
One last thing, what is the AppVeyor project URL? it seems https://ci.appveyor.com/project/istanbuljs/nyc isn't right

XhmikosR:
NVM my mind was stuck for a while :P It's https://ci.appveyor.com/project/coreyfarrell/nyc

coreyfarrell:
That's weird, I don't understand why it's listed under by username instead of istanbuljs?  Did I do something wrong?

XhmikosR:
I remember I had that issue too in the past. Not sure how to solve it. But I updated the badge/link to point to your project for now.

coreyfarrell:
Sorry please switch the badge back, I just found the setting so the appveyor URL is now https://ci.appveyor.com/project/istanbuljs/nyc

XhmikosR:
Alright, this is ready. As much as I'd like to always test against the latest Node.js versions, this is significantly slower. On the other hand this will allow you to test say v14 even before AppVeyor adds it to their images.

PR #1256, opened 2019-12-27
(chore) Minor Markdown tweaks

XhmikosR:


XhmikosR:
I just ran this through [markdownlint](https://dlaa.me/markdownlint/). Your call about adding prettier, although it should definitely be more consistent.

Which reminds me, I missed the docs. I will update the PR.

bcoe:
@XhmikosR markdownlint would totally be fine too 👍 I'd just be tempted to try to automate this process, as well as checking in the fixes; make `npm run lint` fail if `markdownlint` fails.

coreyfarrell:
I am not a fan of prettier for JS code.  The demands it makes about line length are very annoying and IMO actually make the code more difficult to read.

I don't know what prettier would do the markdown, I don't have a strong opinion about using prettier for things that are not JS.

bcoe:
I don't have a strong opinion about the linter we use, but if we want to enforce a style on folks, I like it to be automated ... otherwise contributors end up getting feedback on their edits to README.md based on conventions floating around in someone's head.

XhmikosR:
TBH I don't like prettier either. Not sure how it performs on Markdown files.

An alternative would be to use remark-lint or markdownlint. Again, I'm not sure how good their autofix works, which is the strong point of prettier.

coreyfarrell:
I'm in favor of adding markdownlint as a dev dependency and running it from `npm run lint`.

XhmikosR:
I could make a PR with markdownlint-cli, just keep in mind its auto-fixing it's not like prettier where everything is fixed for the user (IIRC).

XhmikosR:
Just so we are safe, are the doc Markdown files used somewhere else apart from GitHub?

coreyfarrell:
README.md of the latest release is displayed at https://npmjs.com/package/nyc.  Other than that I think it's github only.

Oh before I forget CHANGELOG.md would need to be excluded by any automated lint, it is automatically generated by `standard-version`.

XhmikosR:
OK, so we should be fine. I think this PR can be merged and I'll try to make a PR for markdownlint later.

coreyfarrell:
I talked with @bcoe and we feel that using prettier would be OK as long as it were configured to ignore JS and TS code.  I'm actually leaning towards using `xo` for the actual linter but if you don't mind please allow me to push that change as I'll likely want to disable a couple rules and it's easier for me to make those decisions if I work through the switch myself.

XhmikosR:
Your project, your call :)

I actually did use xo in https://github.com/istanbuljs/nyc/pull/1253#issuecomment-569305690 but with a config pretty opinionated so that I ignored the style changes

PR #1258, opened 2019-12-28
chore: Switch from standard to xo

coreyfarrell:
I set `space: 2` in the xo config as the default is tabs but I think @JaKXz has mentioned that spaces are more readable on mobile devices.  If this is incorrect my weak preference is to use tabs per xo default.

Most changes in this PR were automatic.

CC @XhmikosR

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,6 @@
   ],
   "xo": {
     "ignore": [
-      "self-coverage",
-      "tap-snapshots",
       "test/fixtures"
     ],
     "space": 2
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "the Istanbul command line interface",
   "main": "index.js",
   "scripts": {
-    "lint": "standard",
+    "lint": "xo",
     "pretest": "npm run lint && npm run clean && npm run instrument",
     "test": "tap",
     "snap": "npm test -- --snapshot",
@@ -22,12 +22,13 @@
     "bin/*.js",
     "lib/**/*.js"
   ],
-  "standard": {
+  "xo": {
     "ignore": [
-      "/tap-snapshots/",
-      "**/fixtures/**",
-      "**/test/build/*"
-    ]
+      "self-coverage",
+      "tap-snapshots",
+      "test/fixtures"
+    ],
+    "space": 2
   },
   "keywords": [
     "coverage",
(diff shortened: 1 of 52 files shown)


XhmikosR:
Ouch so many changes :/

Do you really want to change the style? I don't mind personally, I just feel this kind of changes should be done once.

I think we are better off with merging first #1253 so that the non-stylistic changes are not "hidden"

package.json
```diff
@@ -22,12 +22,13 @@
...
     "lib/**/*.js"
   ],
-  "standard": {
+  "xo": {
     "ignore": [
-      "/tap-snapshots/",
-      "**/fixtures/**",
-      "**/test/build/*"
-    ]
+      "self-coverage",
```
XhmikosR: You don't need to add this because it's in .gitignore and xo respects it for the excludes

XhmikosR: Basically this should be enough, but please confirm:

```
    "ignores": [
      "**/fixtures/**"
    ],
```

coreyfarrell: Actually `tap-snapshots` isn't needed either (the latest release of xo ignores it by default).

XhmikosR:
Personally, I'd just say make sure you want to change semi or the indentation before going through with this. Because the current patch touches almost every line in the codebase.

It's OK to customize xo's rules. For example I always disable `capitalized-comments`.

That being said, the final decision is yours to make. Styling choices are pretty objective and what matters is consistency only.

coreyfarrell:
> Personally, I'd just say make sure you want to change semi or the indentation before going through with this. Because the current patch touches almost every line in the codebase.

WRT to semi's I'm absolutely positive I want them added and enforced, that's not an incidental change it's one of the major reasons I wish to switch away from `standard`.

I'm not going to rush merging this, going to give a chance for other maintainers to chime in on preferred indentation.  I agree if we're going to change the indentation it should be done now but I'm also fine with leaving the indentation at 2 spaces.

> It's OK to customize xo's rules. For example I always disable `capitalized-comments`.

I was considering disabling `capitalized-comments` (it's sometimes annoying) but it actually pushed me in the direction of better comments in a few places so I left it enabled.

XhmikosR:
>WRT to semi's I'm absolutely positive I want them added and enforced, that's not an incidental change it's one of the major reasons I wish to switch away from standard.

Then I guess all is good. As long as this kind of changes don't happen more than once, that is.

>I was considering disabling capitalized-comments (it's sometimes annoying) but it actually pushed me in the direction of better comments in a few places so I left it enabled.

Don't get me wrong, I see the reasoning behind this, I just don't want to have the linter against me when I want to do something more important. Similarly with the `spaced-comment` rule.

Either way, I do too prefer xo myself, I just customize it because it's very opinionated, but that can be good and bad :P

Anyway, most of the changes in #1253 were found using xo, which should make merging easy enough as long as you rebase and do `xo --fix` after the merge, but it's your call.

Issue #1271, opened 2020-01-23
What's the valid structure of .nycrc.json? 


mwoz123:
Readme says it should start with 
`{ "nyc" : { `
While with `"nyc":{` it doesn't work (at least for windows 10, and yesterday downloaded nyc) in .nycrc.json

Luckily I found this : 
https://github.com/istanbuljs/babel-plugin-istanbul/issues/120

Which let me try to remove the nyc parent

coreyfarrell:
The README shows examples of package.json where the configuration is placed inside an `nyc` object.  `.nycrc.json` and other configuration files just directly contain the configuration.

@AndrewFinlay do you have any thoughts on how we might clarify this?

mwoz123:
Maybe then would be good at least to mention that outside package.json 
`"nyc" : {` will cause the whole configuration file will be ignored, and should be removed ?

AndrewFinlay:
Personally I don't think writing the config examples from a nyc stanza perspective is too bad but I can see how it could be misleading.  At the very least the config examples should be consistent.

I can think of two solutions to this, the first is to drop the `"nyc": {` from the config examples in the documentation and state explicitly that example should be a part of config. The other solution is to look for a `nyc` block as the top level inside a config file and then use that instead of the top level config but I don't know how practical this is.

coreyfarrell:
My strong preference for this issue is a documentation fix.  Dropping `"nyc": {` from the config examples seems fine to me, probably explicitly state that configuration in package.json must be wrapped in the "nyc" key?  I'm not sure coherent documentation is not my strength.

PR #1276, opened 2020-02-02
docs: Remove `nyc` containing object in json config examples

AndrewFinlay:
Updated some of the text to make support the changes to the config examples.
Updated the doc for the `ignore-class-method` example to try and reflect that it accepts an array of strings, rather than just a single string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,11 +71,9 @@ You can then add the specific configuration options you want that aren't in that
 
 ```json
 {
-  "nyc": {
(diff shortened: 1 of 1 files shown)


AndrewFinlay:
This should cover the problems highlighted in #1271 

I've updated the doc for `ignore-class-method` according to the schema, please let me know if it isn't correct.

coreyfarrell:
@AndrewFinlay thanks for this patch, sorry for the long delay!

Issue #1263, opened 2020-01-02
JavaScript heap out of memory in v15.0.0.


josephzidell:
In v15.0.0, JS runs out of memory on a relatively small project while running nyc in a docker container. In v14.1.1 it all worked fine.

## Link to bug demonstration repository
This is a sensitive project and with private source control

## Expected Behavior
Coverage shows

## Observed Behavior
Javascript has a heap overflow

```
test-frontend |
test-frontend |   [] 100% (completed)
test-frontend |
test-frontend |  WEBPACK  Compiled successfully in 32137ms
test-frontend |
test-frontend |  MOCHA  Testing...
test-frontend |
test-frontend |
test-frontend | <--- Last few GCs --->
test-frontend |
test-frontend | [50:0x3172330]   138018 ms: Mark-sweep 1473.6 (1500.4) -> 1469.9 (1500.4) MB, 1653.7 / 0.1 ms  (average mu = 0.133, current mu = 0.003) allocation failure scavenge might not succeed
test-frontend | [50:0x3172330]   140149 ms: Mark-sweep 1474.3 (1500.9) -> 1470.5 (1500.9) MB, 1958.3 / 0.0 ms  (average mu = 0.107, current mu = 0.081) allocation failure scavenge might not succeed
test-frontend |
test-frontend |
test-frontend | <--- JS stacktrace --->
test-frontend |
test-frontend | ==== JS stack trace =========================================
test-frontend |
test-frontend |     0: ExitFrame [pc: 0x1374fd9]
test-frontend | Security context: 0x2a0f1cf008a1 <JSObject>
test-frontend |     1: SourceMapConsumer_allGeneratedPositionsFor [0x39a4144ed0d9] [/usr/src/app/node_modules/source-map/lib/source-map-consumer.js:~178] [pc=0x35a2ce60f67b](this=0x03f2718c0111 <BasicSourceMapConsumer map =
0x149ab65e7761>,0x2352fe700259 <Object map = 0x2d594ef7fbd1>)
test-frontend |     2: /* anonymous */(aka /* anonymous */) [0x24b0c0b9a129] [/usr/src/app/node_modules/@babel/...
test-frontend |
test-frontend | FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
test-frontend |
test-frontend | Writing Node.js report to file: report.20200102.194151.50.0.001.json
test-frontend | Node.js report completed
[...]

PR #1293, opened 2020-03-23
Data merge concurrency

SamuraiJack:
This PR fixes https://github.com/istanbuljs/nyc/issues/1263

Instead of fetching all reports into memory and then merging, we do that one-by-one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ test/build/
 self-coverage/
 *.swp
 needs-transpile.js
+/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ test/build/
 self-coverage/
 *.swp
 needs-transpile.js
-/.idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ test/build/
 self-coverage/
 *.swp
 needs-transpile.js
+/.idea
```
coreyfarrell:
```suggestion
```

I'd prefer not to add IDE specific excludes to the repository.  You can put this into `.git/info/exclude` on your own clone.

index.js
```diff
@@ -408,11 +408,16 @@ class NYC {
 
   async getCoverageMapFromAllCoverageFiles (baseDirectory) {
     const map = libCoverage.createCoverageMap({})
+    const files = await this.coverageFiles(baseDirectory)
 
-    const data = await this.coverageData(baseDirectory)
```
coreyfarrell: The changes to this function seem fine, one request could you add a comment before the definition of `coverageData` saying `// TODO: Remove from nyc v16`?

PR #1312, opened 2020-05-17
feat: support passing `--reporter-options` to nyc's own reporters

brettz9:
This aims to allow nyc reporters to receive config (without need for setting environmental variables).

My own interest is in building a reporter which can output, for example, the top 10 visited lines of a file or project. I'd like for users to be able to pass in config which can control how much is listed, what information, etc.

If it is ok, maybe I could adapt this so that the reporter could even get the entire `config` object (e.g., if I adapted the badge-making tool I created, [coveradge](https://www.npmjs.com/package/coveradge), to work as a reporter, I could immediately detect the user's thresholds, e.g., `statements`, without having to re-query them).

### DIFF
--- a/index.js
+++ b/index.js
@@ -446,6 +446,7 @@ class NYC {
       reports.create(_reporter, {
         skipEmpty: this.config.skipEmpty,
         skipFull: this.config.skipFull,
+        reporterOptions: this.config.reporterOptions,
         projectRoot: this.cwd,
         maxCols: process.stdout.columns || 100
       }).execute(context)


coreyfarrell:
I appreciate your contribution but I lean towards not wanting to incorporate this functionality in this way.  This would cause the same options to go into every report and cause confusion since some options would not be supported.  Additionally some options (`file` for example) cannot be set the same for each reporter.

I'm thinking that complex report configuration could be supported through `.nycrc` / other configuration files but unsupported by CLI report setup.  For example:
```js
{
	"reporter": {
      "lcov": {
        "file": "custom-filename.lcov"
      }
    }
}
```

This would require some significant testing to ensure things work properly (IE with the above config `nyc --reporter none` would need to set `"reporter": ["none"]`.  Work for this would span multiple repositories:
* [@istanbuljs/load-nyc-config](https://github.com/istanbuljs/load-nyc-config) would need logic to deal with switching between array and object
* [@istanbuljs/schema](https://github.com/istanbuljs/schema) would need to specify that `reporter` can be an object
* [nyc](https://github.com/istanbuljs/nyc) may need to have logic to handle `--reporter` argument overwriting what is provided by configuration.  Additionally nyc would need a patch to handle passing the new `reporter` config structure to actual reporters.

These changes would need testing.  Sorry to raise the bar so high on this functionality but I think this is what will be needed to provide proper support in a way that won't cause confusion.  Advanced warning I'm not sure if this can be done semver-minor, it could end up needing to be deferred to nyc 16.

brettz9:
Btw, I've gotten a very 0.1.0-ish version out for my visit count reporter: https://github.com/brettz9/nyc-line-visit-count-reporter .

It is not very polished now and it could no doubt improve in robustness relative to the built-in reporters (part of my reason for filing #1318), but if you would like to add such a reporter, it is pretty handy to have this line count info (even if, as discussed at https://github.com/istanbuljs/istanbuljs/issues/565 , this wouldn't always reflect bottlenecks in performance if the high line count was nevertheless performant code relative to some less visited but problematic/high latency code).

Btw, I'd be grateful if you have any immediate ideas as to why, though running my reporter with `execFile` from within a Mocha test works, using `nyc` on that test does not consistently get the reporter to run (even the native reporters don't work in this way). I was relying on `cwd` which could be complicating things, or maybe it is due to problems running nyc on another nyc process. Just thought I'd ask if you had any immediate ideas. But probably not the most common use case to worry about. Thanks!

PR #1313, opened 2020-05-17
chore: add `bugs` (used, e.g., by npmjs)

brettz9:


coreyfarrell:
Thanks for contributing.  Although `npm bugs nyc` seems to navigate to the proper github issues URL by default it is nice to set this explicitly.

brettz9:
Cool, thanks. https://www.npmjs.com/package/nyc doesn't make an issues link when there is no `bugs`, but I gotcha re: `npm bugs`. Thanks!

PR #1339, opened 2020-08-07
Add introductory section to README

mrsarm:
Awasome project! thanks.

That said, I came to this project because IntelliJ suggested me to use it in order to execute mocha tests with code coverage, so like any developer, before accept the IntelliJ suggestion I went to the official Npm section to see what is this package about, and like here in the Github README, you can see a lot of links and information but not what `ncy` actually does, so I just copy&pasted the introductory section from the official site.


JaKXz (approved):
Makes sense to me! Thanks!

Issue #1367, opened 2020-11-19
There is a Vulnerability in nyc 15.1.0


martinoppitz:
### Description

There is a vulnerability in y18n: https://snyk.io/test/npm/nyc/15.1.0

### Solution

Try to update / upgrade yargs.

KTOmega:
+1, I have a project that's impacted by this vulnerability rooted in our dependency on `nyc` (and `mocha`).

An issue I see is that the latest `yargs` doesn't support Node v8, whereas `nyc` seems to, based on the `package.json` & the tests in the linked PR. Does the `nyc` team have a roadmap for deprecating support for Node v8?

coreyfarrell:
nyc 15 will never drop support for node.js 8, doing so is semver major (node.js 8 being EOL is not relevant to this).  nyc 16 will require node.js 10 but it is not going to be rushed out (no current timeframe for nyc 16 to be released).  The only way this issue can possibly be hit is if the system environment variable `LC_ALL` is set to `__proto__` before nyc runs.  This would only happen if your system is compromised to begin with.

https://github.com/yargs/y18n/issues/112 is the upstream bug, if a fix is backported to y18n 4.x then yargs 15 and thus nyc 15 will be able to pull the fix per semver range.

Leaving this ticket open so others can see the status.

bcoe:
@coreyfarrell I have released `y18n@4.0.1` which has the patch.

coreyfarrell:
@bcoe thanks for taking care of this!

Closing this ticket as an in-range update is now available.

firefoxNX:
will there be a nyc release to update y18n to version 4.0.1?

coreyfarrell:
> will there be a nyc release to update y18n to version 4.0.1?

No, nyc only gets a new release for dependency updates when an out of range update is required.

firefoxNX:
Sorry I am new to nodejs. What do you mean by "out of range update"?

coreyfarrell:
Out of range versions would not be allowed by the dependency definitions.  In this case `y18n@^4.0.0` is pulled in by the version of yargs we require.  The `^4.0.0` means that any version 4.x.x will be accepted (it's in range).  If your project is still installing 4.0.0 then you need to run updates locally.

PR #1368, opened 2020-11-19
fix: upgrade yargs to v16 to fix a high vulnerability

martinoppitz:
Fix: https://github.com/istanbuljs/nyc/issues/1367


coreyfarrell:
This is not possible, nyc 15 supports node.js 8 so it cannot upgrade to yargs 16.  I'll comment further on the issue.

PR #1370, opened 2020-12-03
feat: Add support to change files tree in reporter . --summarizer=flat|nested|pkg option

blackswanny:
During debug i found there is no way to change how files are represented in report (flat, tree, etc) despite this feature is available in `istanbul-lib-report` package that generates files nodes tree for report. This allows to pass override of this feature from nyc down to `istanbul-lib-report`

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,6 +366,20 @@ If you opt to pre-instrument your source-code (rather than using a just-in-time 
 _Important: If you are using nyc with a project that pre-instruments its code, run nyc with the configuration option `--exclude-after-remap` set to `false`.
 Otherwise nyc's reports will exclude any files that source-maps remap to folders covered under exclude rules._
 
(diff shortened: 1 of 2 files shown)


michaelhays:
Is the failed check preventing this from being merged? This option would be really useful!

blackswanny:
@coreyfarrell it fails on some particular NodeJS vers , but not all. And idk, why my change would have caused it. It shouldn't be related

Gu7z:
Is this project still actively maintained? I believe implementing this feature would be incredibly beneficial and enhance the utility of the project significantly. Looking forward to any updates on this.

@SimenB @bcoe

michaelhays:
@Gu7z No, there hasn't been a commit here in nearly 3 years. I'd recommend using https://github.com/bcoe/c8

PR #1401, opened 2021-04-14
Preservation of actual extensions specified

BDGrace:
Upper case characters are being transformed into lower case characters currently.  

This would keep extensions exactly as they are specified.

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,7 +57,6 @@ class NYC {
 
     this.extensions = [].concat(config.extension || [])
       .concat('.js')
-      .map(ext => ext.toLowerCase())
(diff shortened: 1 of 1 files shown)


l198881 (approved):

index.js
```diff
@@ -57,7 +57,6 @@ class NYC {
 
     this.extensions = [].concat(config.extension || [])
       .concat('.js')
-      .map(ext => ext.toLowerCase())
```
l198881, line 60: .map(ext => ext.toUpperCase())

l198881 (approved):

PR #1413, opened 2021-06-14
Bump deprecate package uuid@3.x

cschanaj:


AncientSwordRage:
It looks like the above branches would run if we dropped support for node 10, which seems reasonable in 2022?

PR #1441, opened 2021-12-11
build: replace deprecated uuid dependency

johnrees:
UUID 3 is deprecated

`Please upgrade to version 7 or higher. Older versions may use Math.random() in certain circumstances, which is known to be problematic. See https://v8.dev/blog/math-random for details.`

Upgrading it will be very helpful to push things along for libraries depending on this.

Thanks!


pshenmic:
Bump this

Duplicated: https://github.com/istanbuljs/nyc/pull/1413

PR #1484, opened 2022-08-19
make input relative to --cwd path

loidy:


sw-tracker:
Any updates on accepting this change? I am also experiencing this problem...

PR #1550, opened 2024-02-27
[Breaking Change] Node Upgrade to >=20.0.0

BboyAkers:
# What is in this PR?
- Bumping version of the required node version to LTS `>=v20.0.0`

# Why?
Current node version is not maintained.

Additional Info:
Currently getting this error below while running `npm run test`
Screenshot 2024-02-26 at 8 13 17 PM

I did an `npm list jackspeak` to see further where this package is since it's not in the top-level dependency tree. This was the result. For context, I did upgrade tap to the latest version, `18.7.0`.
Screenshot 2024-02-26 at 8 17 30 PM

Would love some eyes on what you might think could be the issue. I'll continue hacking away at debugging this error! @bcoe  🙂

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nyc",
-  "version": "15.1.0",
+  "version": "16.1.0",
   "description": "the Istanbul command line interface",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


bcoe:
I think addressed by https://github.com/istanbuljs/nyc/pull/1552

PR #1585, opened 2024-11-04
chore(deps): updated multiple deps

work933k:
Hereby i propose to update multiple of the dependencies.
Testing is done via the unit-tests.

Thanks in advance, hope it helps.


work933k:
Need to fix Windows issue

Issue #1563, opened 2024-05-28
Use of deprecated glob and rimraf dependencies


jwasnoggin:
When installing `nyc` there are a number of npm warnings about deprecated packages:
```
$ npm i -D nyc 
npm WARN deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm WARN deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm WARN deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm WARN deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm WARN deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
```

It looks like `glob` is only used in tests, so maybe this should be moved to `devDependencies`?

Usages of `rimraf` will need to be modified slightly but this [doesn't look too onerous](https://github.com/isaacs/rimraf?tab=readme-ov-file#v3-to-v4).

## Environment Information

```
  System:
    OS: Linux 6.8 Fedora Linux 38 (Workstation Edition)
    CPU: (8) x64 Intel(R) Core(TM) i5-8350U CPU @ 1.70GHz
    Memory: 3.95 GB / 15.46 GB
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
  npmPackages:
    nyc: ^15.1.0 => 15.1.0 
    ts-node: ^10.9.2 => 10.9.2 
    typescript: ^5.4.5 => 5.4.5 
```

DinoChiesa:
It's complicated.  rimraf is used directly by nyc, but also in various modules that are dependencies of nyc.  Including [istanbul-lib-processinfo](https://github.com/istanbuljs/istanbul-lib-processinfo), which itself has a devDependency on nyc.  
```
$ npm list rimraf
nyc@17.0.0 /Users/dev/istanbuljs/nyc
├─┬ istanbul-lib-processinfo@2.0.3
│ └── rimraf@3.0.2 deduped
├── rimraf@3.0.2
├─┬ spawn-wrap@2.0.0
│ └── rimraf@3.0.2 deduped
├─┬ standard@14.3.4
│ └─┬ eslint@6.8.0
[...]

PR #1591, opened 2025-03-26
chore(deps): Bump glob to v9, rimraf to v4, test-exclude to v7

anomiex:
Resolves deprecation warnings about glob, inflight, and rimraf. See #1563.


lynxSven:
@bcoe Is something still needed for this PR? Maybe you can help us?

PR #1622, opened 2026-07-30
fix: do not let --all overwrite coverage of files that were loaded

### LINKED ISSUE #1372: nyc --all option reporting wrong coverage results if all files are instrumented
## Expected Behavior

nyc should show correct coverage results of all instrumented files with --all option.

For **example**, if the application's main.js loads a.js and b.js but your unit tests only load a.js your total coverage could show as **some percentage** covered. If any of those files remain uncovered they will be factored into the report with a default of 0% coverage.

## Observed Behavior

This was the result i am getting with **all** option enabled, shows no coverage results for any files even it failed for the files touched by my test suite shown below:  

coverage_withall

If i disabled the --all option, it shows the coverage report for the files touched by test suite as shown below:

coverage_withoutall

### Troubleshooting steps
- [x] still occurring when I put `cache: false` in my nyc config
```
{
  cache: false,
  all: true,
  "exclude": ["./coverage/**", "./dist/**", "**/*.spec.ts"]
}
```

## Environment Information
"istanbul-instrumenter-loader": "^3.0.1",
"nyc": "^15.1.0",

i am using cypress test runner and istanbul-instrumenter-loader to instrument the source code with this configuration in webpack:
```
 {
    test: /\.(js|ts)$/,
    loader: 'istanbul-instrumenter-loader',
    options: { esModules: true },
    enforce: 'post',
    include: require('path').join(__dirname, '..', 'src'),
    exclude: [
    /\.(e2e|spec)\.ts$/,
    /node_modules/,
    /(ngfactory|ngstyle)\.js/,
    ]
}
```

Eljees:
Fixes #1372

`--all` can wipe out the coverage of files that *were* exercised, which is why the report comes back at
0% for everything rather than just for the untouched files.

`addAllFiles()` writes its placeholder straight into the global coverage object, unconditionally:

```js
const coverage = coverageFinder()
const lastCoverage = this.instrumenter().lastFileCoverage()
if (lastCoverage) {
  coverage[lastCoverage.path] = {
    ...lastCoverage,
    // Only use this data if we don't have it without `all: true`
    all: true
  }
}
```

The comment already states the intended rule — *only use this data if we don't have it without
`all: true`* — but there is no condition in the code implementing it. When a file already has real
counters in `coverage[path]`, the freshly instrumented (all-zero) copy replaces them.

The fix is the missing check. Everything else stays as it was: files that were never loaded still get
their zeroed placeholder.

### Before / after

New case in `test/add-all-files.js` — real counters are put in the global coverage object first, then
`addAllFiles()` runs, and the test asserts they survived:

```
before:  not ok - real hit count survives --all
         not ok - entry is not marked as an --all placeholder
         # { total: 15, pass: 13, fail: 2 }

after:   ok - real hit count survives --all
         ok - entry is not marked as an --all placeholder
         # { total: 15, pass: 15 }
```

Run with `tap test/add-all-files.js` after `npm run instrument`, on node 24.

I kept the wording of the original comment and only moved it above the `if`, since it now describes a
condition that actually exists.

AI-assisted (LLM used for drafting); the change and both runs above are mine.


Eljees:
Gentle ping - this has been open since 30 July.

No CI has run on it yet. The check suite for `00d907e` is sitting at `action_required` with zero runs, so the workflow is waiting on a maintainer to approve a run for a first-time contributor. Pending approvals expire after about 30 days, so this one runs out towards the end of August and the branch would have to be re-pushed.

The change and the before/after are in the description. Happy to rebase, or to close it if it is not wanted.

Eljees:
Following up - the pending run approval for 00d907e is close to its ~30-day expiry (opened 30 July), after which the branch would need a re-push. Happy to re-push if that's easier, or if this isn't wanted I'm fine closing it.

Eljees:
Re-pushed (`569f120`, empty commit, no code changes) to refresh the pending Actions approval window before it expired. Still happy to close instead if that's preferred.

